PR #31, opened 2014-11-27
test all solvers on the system, allow user to change default solver

karanveerm:
1. `get_default_solver` and `set_default_solver` can be used to retrieve the current default solver.
2.  `can_solve_mip` and `can_solve_sdp` can be used to see if a solver supports those operations. So far Mosek, Gurobi and GLPK support MIPs and only SCS supports SDPs/ exponential cones.
3. We can use `include test/run_tests_all_solvers.jl` to run every test a solver supports.
4. We can use `include test/run_tests.jl` to run all the tests supported by the current default solver.
5. Including a test file directly does not check to see if the solver supports those cones.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,8 @@ before_install:
 script:
   - julia -e 'Pkg.init(); run(`ln -s $(pwd()) $(Pkg.dir("Convex"))`); Pkg.pin("Convex"); Pkg.resolve()'
   - julia -e 'using Convex; @assert isdefined(:Convex); @assert typeof(Convex) === Module'
-  - if [ $JULIAVERSION = "julianightlies" ]; then julia --code-coverage test/run_tests.jl; fi
-  - if [ $JULIAVERSION = "juliareleases" ]; then julia test/run_tests.jl; fi
+  - if [ $JULIAVERSION = "julianightlies" ]; then julia --code-coverage test/run_tests_all_solvers.jl; fi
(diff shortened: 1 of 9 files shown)


joehuchette:
Any chance you could tag a new version? I have a user trying to change the solver from ECOS to SCS, and best I can tell the only way to do that is with ``set_default_solver``, which is not in v0.0.1 in METADATA.

madeleineudell:
I'm happy to increment to v0.0.2 now. (I'd like to wait for SCS to be
released before we level up to v0.1.0). Any objections?

On Tue, Jan 6, 2015 at 8:25 AM, Joey Huchette <notifications@github.com>
wrote:

> Any chance you could tag a new version? I have a user trying to change the
> solver from ECOS to SCS, and best I can tell the only way to do that is
> with set_default_solver, which is not in v0.0.1 in METADATA.
>
> —
> <https://github.com/cvxgrp/Convex.jl/pull/31#issuecomment-68888676>.
>



-- 
Madeleine Udell
Stanford University
www.stanford.edu/~udell

karanveerm:
@joehuchette You can change the solver by passing in the solver along with the problem when calling `solve!`. We'll tag a new version anyway since v0.1.0 is very outdated.

```
```

@madeleineudell , yes, let's wait until SCS is released and dual values are working before v0.1.0. 
The current master won't work for Mosek SDPs, so I'm going to merge in the `sdp_symmetry` branch and release it in a few minutes after that.

Issue #40, opened 2014-12-30
Implement missing atoms


karanveerm:
Among major atoms I can think of:

(Edited as per comments below)
- [ ] log det
- [x] lambda max
- [x] lambda min
- [ ] huber

Are there any important ones that we don't have and I'm missing / atom requests?

davidlizeng:
Yup I can take on the other 2 tonight

madeleineudell:
cool. a few implementation details, just for fun: 

1) huber(x) = 

minimize s^2 + |n|
subject to s+n = x

(ie, you don't need to define it piecewise)

2) logdet(A) = 

maximize sum(log(D[i,i]))
subject to
  D is diagonal
  Z is upper triangular
  diag(D) == diag(Z)
  [D Z; Z' A] is PSD

proof: (adapted from cvxpy documentation, with more details)
Every matrix has an LDL factorization A = LSL^T, with L lower triangular with unit entries on the diagonal, and S diagonal. We'll show that at the solution to the above problem, A = (Z^T D^{-1}) D (D^{-1} Z) = Z^T D^{-1} Z, ie L = Z^T D^{-1} and S = D is an LDL factorization of A. 

Of course, the knarliest constraint is the PSD one: but we know (by Schur complements) it's equivalent to requiring A to be PSD, and A - Z^T D^{-1} Z to be PSD.

First we'll show that the Z and D given by the LDL factorization satisfy the constraints. It's easy to see that they satisfy the first three constraints, and as for the last, we will use the Schur complement characterization. A - Z^T D^{-1} Z = 0, which is PSD, so the last constraint is also satisfied. At this feasible point, the objective value is just the log of the product of the eigenvalues of A (using some nifty properties of the LDL factorization), ie, logdet(A).

Now, for any other feasible point D and Z, we still know A - Z^T D^{-1} Z is PSD, and so by linearity and multiplicativity of det, and the properties we required of D and Z, this implies

det(A) - det(Z^T D^{-1} Z) >= 0
det(A) - det(Z^T) det(D^{-1}) det(Z) >= 0
det(A) - det(D) >= 0

so the D we got from the LDL factorization is the one with the largest determinant. But we're maximizing the (log) determinant of D, so at the solution logdet(D) = logdet(A).

davidlizeng:
Thanks for the details, these were quite helpful in implementing the atoms!

Closed by #41

PR #41, opened 2015-01-02
huber loss and log det atoms

davidlizeng:
new atoms for ee364a as per #40 
- huber loss
- log det

### DIFF
--- a/src/Convex.jl
+++ b/src/Convex.jl
@@ -54,6 +54,10 @@ include("atoms/exp_cone/log.jl")
 include("atoms/exp_cone/logsumexp.jl")
 include("atoms/exp_cone/entropy.jl")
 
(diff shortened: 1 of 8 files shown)


madeleineudell:
any reason why we need a new test_other? seems like the huber test could go in test or test2 (btw, is there a difference?), and logdet could go in test_sdp.

davidlizeng:
the huber test can go into test or test2 (not sure what the difference is either), but logdet needs both exp and sdp cones to be supported, so it kinda needs to be its own thing

madeleineudell:
sounds reasonable. but let's rename test_other to test_allcones or test_exp_and_sdp or something more evocative. other than that this PR seems good to go.

Issue #42, opened 2015-01-02
clean up tests


madeleineudell:
what is the difference between test and test2? if there is none, they should be merged; if there is a difference, the naming scheme should reflect it. perhaps test_lp?

karanveerm:
They need to be broken into test_lp and test_socp. I'm also planning to move the tests over to use FactCheck, so I'll probably do these together.

karanveerm:
Already closed by #47

PR #47, opened 2015-01-12
clean up tests to use FactCheck

davidlizeng:
Reorganized test.jl and test2.jl into affine, lp, and socp test files as per #42 . 
All tests also now use the FactCheck package.


karanveerm:
Few issues:

- This will always pass travis because of the exit status. Read: https://github.com/JuliaLang/FactCheck.jl#exit-status and fix.
- It's preferable to have test_lp separate, so we can test other LP solvers on it as well.

karanveerm:
@davidlizeng I think this is ready to merge into master. Coverage should go up another 20% after we test with SCS on travis, after which most of the uncovered lines are error checks, etc. of lower priority.

REQUIRE
```diff
@@ -1,3 +1,4 @@
 julia 0.3
 MathProgBase 0.3.8
 ECOS 0.4.1
+FactCheck 0.2.5
```
mlubin: This is usually listed in ``test/REQUIRE`` since it's only needed to run tests.

karanveerm: What would be the nicest way for FactCheck to still be present for travis?
JuMP seems to be doing this: https://github.com/JuliaOpt/JuMP.jl/blob/master/.travis.yml#L11

mlubin: Oops, I guess we never tidied that up. @IainNZ 
If you just use ``Pkg.test("Convex", coverage=true)`` then first the testing dependencies will be installed and then ``test/runtests.jl`` will be run. That's the "right" way to do it.

karanveerm: Done, thanks! I'm going to keep this PR open for another few days to see if anyone has other feedback, and merge it then.

Issue #29, opened 2014-11-19
Use dual values from ECOS.jl


IainNZ:
Yo,
I'm not sure if you want it, but ECOS.jl now gives you the duals through the mathprogbase interface (I've tagged this as ECOS.jl v0.3.0).
Testing may reveal errors because I wouldn't describe our current MPB conic interface tests comprehensive, but I'm pretty happy its correct for LPs and at least simple SOCPs.

PR #50, opened 2015-01-18
Duals

karanveerm:
Changes:

- duals seem to be correct. Tested with ECOS, Mosek and one of my SCS branches. Gurobi and GLPK don't seem to have getconicdual implemented.
- added a few TODOs in the code, mostly for @davidlizeng to add comments :smile: 
- Got rid of the now unnecessary extra n^2/2 constraints we added for SDPs. In retrospect, I should have probably put that in a separate PR.

Fixes https://github.com/JuliaOpt/Convex.jl/issues/29

### DIFF
--- a/src/conic_form.jl
+++ b/src/conic_form.jl
@@ -2,6 +2,7 @@ export ConicObj, ConicConstr, UniqueConicForms
 export +, -, *, promote_size, get_row
 export cache_conic_form!, has_conic_form, get_conic_form
 
+# TODO: Comment every single line
 ConicObj = Dict{Uint64, Value}
 
 function -(c::ConicObj)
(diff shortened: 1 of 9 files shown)


mlubin:
To get duals for Gurobi and GLPK we'll need a wrapper like https://github.com/JuliaOpt/MathProgBase.jl/blob/master/src/SolverInterface/quad_to_conic.jl but in the other direction. We also don't know how to get conic duals out of Gurobi at all, but that's another issue.

davidlizeng:
I think for now we can merge this into master to get ready for releasing 0.0.3, and then worry about duals for Gurobi/GLPK if someone requests them/at a later time.

PR #58, opened 2015-01-27
pnorm

hsnamkoong:
rational_norm.jl, power_to_socp.jl added. Handles the p-norms for p > 1 where p is rational. Reduces the
p-norm constraint to at most 2d ceil(log2(n + m)) + O(d) second order cone constraints. Here d is the dimension of the problem and p is converted to a rational with value p = n / m.


madeleineudell:
Two minor issues that I see:

1. The Travis build fails. It looks like this is because some errors accumulate when many SOCP cones are used. It's fine to just use a higher tolerance, eg 10*TOL.
image

2. Coverage has decreased significantly, indicating that not all the cases in power_to_socp.jl are exercised. We should be able to fix that by adding more tests with a wider range of rational powers.

mlubin:
Cool stuff, here are the coverage reports:
https://coveralls.io/files/444569489
https://coveralls.io/files/444569493

src/atoms/second_order_cone/rational_norm.jl
```diff
@@ -0,0 +1,118 @@
...
+export rational_norm
+
+### k-norm for rational k
+
+type RationalNormAtom <: AbstractExpr
+  head::Symbol
+  id_hash::Uint64
+  children::(AbstractExpr,)
+  size::(Int, Int)
+  k::Rational
```
mlubin: ``Rational`` is an abstract type, so any code that interacts with this field will be very slow. You may want ``Rational{Int64}`` or ``Rational{BigInt}`` instead.

jduchi:
The test failing because of 53 constraints added is odd; none of my tests (locally run) have that issue, and I chose powers that should generate relatively few constraints. [Happy to have some tests added that don't do that.] Hong, what happens when you run locally?

mlubin:
Here's a question that probably doesn't have a good answer, but I'll ask it anyway. For these atoms that introduce many conic constraints, how could users enforce that the constraint in its original form is satisfied to a given tolerance?

jduchi:
As these are p-norms, as long as p is rational, these conic constraints are exact representations of the original cone constraint (i.e. ||x||_p \le t is equivalent to the set of SOCP representations we introduce, at least in infinite precision.) In finite precision, I don't have enough experience with these representations to say anything with confidence.

mlubin:
Right, but when the problem is passed to a solver, it will ensure that each SOC constraint is satisfied up to a user-provided tolerance. Given that the norms are represented by multiple SOC constraints, the errors can easily accumulate beyond the constraint-wise tolerance. If a user really wants to ensure a tolerance for the norm representation, they would have to understand the transformation that's being performed and adjust the constraint-wise tolerance accordingly Bonferroni-style (if I can abuse the terminology).

madeleineudell:
the SOC constraint is \|x\|_2 <= p. suppose each pnorm constraint is
represented as \|x_i\|_2 <= p_i, i=1:k. if the SOC constraints are
saturated at the solution, then at the solution  \|x_i\|_2 == p_i \pm tol
for every i. (\pm might be solver dependent but non-random; not sure about
this.) if the errors are additive (which they may well be, because each SOC
constraint should be saturated, and the p's sum up into the objective
value), then our total error is k*tol.

If instead we multiply the inputs to the SOC constraints by k, then we get
\|k x_i\|_2 == k p_i \pm tol, ie, \|x_i\|_2 == p_i \pm tol/k, so the error
sums to tol, not k*tol.

So we've observed that Miles' "Bonferroni style" constraint-wise tolerance
modification can be implemented without user interference.

On Tue, Jan 27, 2015 at 3:59 PM, Miles Lubin <notifications@github.com>
wrote:

> Right, but when the problem is passed to a solver, it will ensure that
> each SOC constraint is satisfied up to a user-provided tolerance. Given
> that the norms are represented by multiple SOC constraints, the errors can
> easily accumulate beyond the constraint-wise tolerance. If a user really
> wants to ensure a tolerance for the norm representation, they would have to
> understand the transformation that's being performed and adjust the
> constraint-wise tolerance accordingly Bonferroni-style (if I can abuse the
> terminology).
>
> —
> <https://github.com/JuliaOpt/Convex.jl/pull/58#issuecomment-71755208>.
>



-- 
Madeleine Udell
Stanford University
www.stanford.edu/~udell

karanveerm:
Actually, the tests passed on julia release but failed on julia nightlies. I do not think there's an accuracy issue then for the current tests. I can try to look into this tomorrow.

@madeleineudell: Coverage did not decrease much because of the atom. Our coverage was already 71 for julia-releases, due to a [bug in coveralls](https://github.com/IainNZ/Coverage.jl/issues/43). Once the former is taken care of, the coverage will be mostly fine.

jduchi:
OK, honestly not following the issues with testing.

Re: precision of the solvers. My guess is that, as we are reducing a power inequality to a series of SOC constraints, we could get an exponential blowup in inaccuracy. So we might be off by a factor of (1 + \epsilon)^k, where k is the number of introduced constraints. But in the reductions, we introduce at most 2 * log_2(n) constraints for the inequality x^n \le 1; in this case, that's basically a linear error of 1 + k \epsilon as long as \epsilon is small.

Would it make sense to multiply each of the constraints for the p-norms by a factor of k, the number of introduced constraints? Happy to try to make that change.

madeleineudell:
1. @karanveerm, let us know when you've checked out the issue with the build on the nightlies. I don't understand why there'd be an interaction between the latest Julia version and the numerical accuracy of this reduction, which makes me somewhat nervous. Does the new Julia version calculate numerators and denominators of rationals differently?

2. since the numerical accuracy is not problematic on the stable version of Julia, it seems fine to defer the question of rescaling the constraints until we've thought about it some more. I'd like to run it by a few solver developers in order to make sure that our understanding of the tolerance guarantee is correct. this issue affects many other transformations to conic form in Convex.jl, though the depth of the problem is potentially greatest here since none of the other transformations recurse.

3. @jduchi, if in the meantime you want to try multiplying each of the constraints for the p-norms by a factor of k, just to see if the reduction in error is what we expect, that would be fantastic.

so proposed game plan is: 1) understand why it's failing on nightlies, then 2) merge (assuming it's just a weird Julia 0.4 bug), then 3) hammer out the tolerance/scaling issues.

jduchi:
I just made some modifications (pushed to @hsnamkoong) that fixed a bug in an edge case in the recursions; have a new look. I also modified the Rational to be Rational{Int64} as per @mlubin's suggestion.

I played with scaling/tolerance and didn't see a big difference in performance or the tolerances to which problems were solved (this may have been the edge case bug?).

jduchi:
Oh, and I did a few more comparisons of solvers. It looks like the ECOS solver does a better job solving these problems than does the SCS solver--the accuracy for the ECOS solver in the test https://github.com/hsnamkoong/Convex.jl/blob/master/test/test_socp.jl#L165 is about 1e-9, while for the SCS it is about 5e-3. Maybe best to either specify a solver or simply raise the tolerance? @madeleineudell or @karanveerm, thoughts?

madeleineudell:
The default tolerance is much tighter for ECOS (a second order solver) than
for SCS (a first order solver), so those results make sense; the accuracy
of SCS in the tests you mention is just 5 times the default tolerance of
1e-3. That also means users should expect less precise results using SCS
(unless they manually change the precision). It seems fine to me to leave
the default tolerance as it is, particularly so long as we have a warning
when more than 10 SOCPs are used to represent the p-norm cone.

@karanveerm, any progress on understanding what's up with the nightly build?

karanveerm:
I just ran the test that was passing on stable but failing on nightly again: https://travis-ci.org/JuliaOpt/Convex.jl/builds/48548945
Now both are failing. I believe we were previously passing on stable by pure luck since the solution being returned by SCS is fairly close to the correct values. SCS does not always return exactly the same value.

However, I was very confused why SCS won't be accurate enough, because it's default tolerance is `1e-3`, same as `TOL`. Turns out that it's reaching max no. of iterations and stopping. If I instead pass in `SCSSolver(max_iters=10000)`, everything works. I also tested this PR with Mosek and Gurobi and everything seems fine.

Overall, this PR is great. In order to make the Travis CI tests pass, we can either define a new tolerance, `RAT_NORM_TOL=1e-2` and use that, or use more iterations with SCS for tests in general, by passing in `push!(solvers, ECOSSolver(verbose=0, max_iters=10000))` in `runtests.jl`. Thoughts on this?

jduchi:
It would be good to have the ECOSSolver tested with high enough tolerance
(because for small problems .005 seems too loose for a second-order
solver), but it would also be a shame to have the SCSSolver have to run
with so many iterations for all the tests when it's only these (for
rational norm) that have trouble. Is there a way to specify that SCSSolver
should run longer only on these tests?

karanveerm:
Other tests won't be affected. SCSSolver won't run for 10k iterations for any other test, since they all converge in ~1000 iterations.

The way to specify SCSSolver to run longer on only these tests would be hacky:
```julia
s = get_default_solver()
if typeof(s).name.name == :SCSSolver
  push!(s.options, (:max_iters, 10000))
end
solve!(p, s)
```

madeleineudell:
How about this: let's increase maxiters on the tests, and change the
warning issued when more than 10 constraints are added to suggest
increasing maxiters and decreasing the tolerance of the solver for a more
accurate solution. At that point I think this PR will be ready to merge.

We can then investigate whether there are other options for rescaling
constraints to achieve the specified tolerance. That investigation will be
relevant for other problems in this package as well: in general the
tolerance we currently solve to goes as the specified tolerance times the
depth of the parse tree.

On Friday, January 30, 2015, Karanveer Mohan <notifications@github.com>
wrote:

> Other tests won't be affected. SCSSolver won't run for 10k iterations for
> any other test, since they all converge in ~1000 iterations.
>
> The way to specify SCSSolver to run longer on only these tests would be
> hacky:
>
> s = get_default_solver()if typeof(s).name.name == :SCSSolver
>   push!(s.options, (:max_iters, 10000))endsolve!(p, s)
>
> —
> <https://github.com/JuliaOpt/Convex.jl/pull/58#issuecomment-72275670>.
>


-- 
Madeleine Udell
Stanford University
www.stanford.edu/~udell

jduchi:
OK, I think things look good. Added the hack to that single failing test;
now ECOS and SCS work for all the tests. And I modified the warning as
@madeleineudell suggested.

karanveerm:
This looks great to me, thanks so much!

madeleineudell:
Hurrah! Thanks to everyone for the excellent work!

On Friday, January 30, 2015, Karanveer Mohan <notifications@github.com>
wrote:

> This looks great to me, thanks so much!
>
> —
> <https://github.com/JuliaOpt/Convex.jl/pull/58#issuecomment-72300173>.
>


-- 
Madeleine Udell
Stanford University
www.stanford.edu/~udell

PR #60, opened 2015-01-30
Update quick_tutorial.rst

tummykung:
Solve the right problem.


davidlizeng:
Thanks for the catch!

PR #65, opened 2015-02-07
Add portfolio optimization example

madeleineudell:


karanveerm:
The notebook can be found here: http://nbviewer.ipython.org/github/madeleineudell/Convex.jl/blob/master/examples/portfolio_optimization.ipynb

@madeleineudell For the plot, I believe you're trying to say that in most assets, we don't invest anything at all, and the rest shows the histogram of the allocation. Could you add a comment after the plot making it more clear and interpretable? Perhaps also add another line saying `find(x .> 1e-4)` to return the assets in which we invested? This is great otherwise!

IainNZ:
Can I suggest

```julia
plot(x=1:n,y=x.value,Geom.bar,Guide.xlabel("Asset Index"),Guide.ylabel("Fraction of Portfolio"))
```

Instead?

madeleineudell:
Excellent suggestions. You can view the revised notebook here
<http://nbviewer.ipython.org/github/madeleineudell/Convex.jl/blob/master/examples/portfolio_optimization.ipynb>
.

On Sat, Feb 7, 2015 at 2:39 PM, Iain Dunning <notifications@github.com>
wrote:

> Can I suggest
>
> plot(x=1:n,y=x.value,Geom.bar,Guide.xlabel("Asset Index"),Guide.ylabel("Fraction of Portfolio"))
>
> Instead?
>
> —
> <https://github.com/JuliaOpt/Convex.jl/pull/65#issuecomment-73386823>.
>



-- 
Madeleine Udell
Stanford University
www.stanford.edu/~udell

karanveerm:
:+1:

PR #72, opened 2015-03-24
Warmstart

madeleineudell:
This branch allows warmstarting for both SCS (when combined with my [PR](https://github.com/JuliaOpt/SCS.jl/pull/22) on SCS) and MILPs.

Currently the one limitation is that you can't pass in new options to the solver when warmstarting; I wasn't able to find a way to do this without occasional segfaults. But this branch is safe.

If it looks good I'll write the associated documentation.


madeleineudell:
Warmstarting SCS now works! Also added capability to fix/free variables to make it easier to implement alternating minimization for nonconvex problems.

bermanmaxim:
Thanks for the work on this branch. I am interesting in using a solution of a problem1 to warm-start a different problem2 (the same objective, with different constraints). I believe I could create `problem2.model` manually and call `setwarmstart!` using problem1.solution.primal if the var_to_ranges mapping stays the same between the two problems...

`populate_solution!` would be nice to use in that case in order to directly assign values to `Variable.value` for warm-starting, but it requires computing `var_to_ranges`, which is not readily exposed to the used...

By the way, the `Dict{Uint64, (Int, Int)}` on https://github.com/JuliaOpt/Convex.jl/blob/warmstart/src/solution.jl#L110 should be changed to `Dict{Uint64, @compat Tuple{Int, Int}}` for compatibility with julia 0.4 branch.

madeleineudell:
Thanks for reminding me of this issue! I've just completed a major refactor
of `solve!` and extended the warmstart functionality so that populating the
warmstart primal solution from variable values should be possible and
happen by default when you use `warmstart=true`.

However, note that many of the entries in the primal vector (those
corresponding to expressions, and not to variables) will still *not* be
populated. Doing that would require creating a `exp_to_ranges` map by
analogy to `var_to_ranges` so that each range in the primal corresponds to
either a variable or an expression, and so can be populated based on
variable values (by calling `evaluate` on expressions to get their derived
values).

Furthermore, the dual solution (if you're using a primal dual solver) will
not be able to be warmstarted, which may cause a slowdown.

In case you want to help, here's the functionality on the
nice-to-have-but-not-yet-implemented list:
* construct `exp_to_ranges` map by analogy to `var_to_ranges`
* warmstart primal from expressions
* warmstart dual from whatever dual values have been populated
* tests of warmstart functionality (eg, write a test corresponding to the
warm start you're hoping to do!)

PRs welcome...

On Thu, Jul 30, 2015 at 4:19 AM, Maxim Berman <notifications@github.com>
wrote:

> Thanks for the work on this branch. I am interesting in using a solution
> of a problem1 to warm-start a different problem2 (the same objective, with
> different constraints). I believe I could create problem2.model manually
> and call setwarmstart! using problem1.solution.primal if the
> var_to_ranges mapping stays the same between the two problems...
>
> populate_solution! would be nice to use in that case in order to directly
> assign values to Variable.value for warm-starting, but it requires
> computing var_to_ranges, which is not readily exposed to the used...
>
> By the way, the Dict{Uint64, (Int, Int)} on
> https://github.com/JuliaOpt/Convex.jl/blob/warmstart/src/solution.jl#L110
> should be changed to Dict{Uint64, @compat Tuple{Int, Int}} for
> compatibility with julia 0.4 branch.
>
> —
> <https://github.com/JuliaOpt/Convex.jl/pull/72#issuecomment-126277008>.
>



-- 
Madeleine Udell
www.stanford.edu/~udell

PR #74, opened 2015-04-09
Norm

### LINKED ISSUE #73: operator_norm vs norm
Hi 

I am doing a problem (2.3 from [EE364B exercises](http://stanford.edu/class/ee364b/364b_exercises.pdf)) where I am trying to approximate B as a combination of A_i's. The approximation is in the operator norm. i.e. max singular value. 

Here is the code.

```julia
using Convex

srand(0)
n, p, q = 5, 3, 6
A = {randn(p, q) for _ in 1:n}
B = randn(p, q)

function diff(x)
	d = B
	for i in 1:n
		d -= x[i] * A[i]
	end
	return d
end

################### CVX 
xcvx = Variable(n)
problem = minimize(operator_norm(diff(xcvx)))
solve!(problem)

println("Status : ", problem.status)
println("Value : ", problem.optval)
u, d, v = svd(diff(xcvx.value))
println("X: ", xcvx.value, norm(diff(xcvx.value)), d[1])
fst = problem.optval
```

I am having to say operator_norm in stead of norm. If I say norm, cvx is minimizing something else. I think it should be consistent with underlying Julia where norm of a matrix is the operator norm by default.

madeleineudell:
I updated the norm atom to be consistent with the behavior in base julia. In particular, norm(m, p) is the *induced* (operator) norm if m is a matrix, not the vector norm: [quoth the docs](http://julia.readthedocs.org/en/latest/stdlib/linalg/#Base.norm)

Closes #73.

### DIFF
--- a/src/Convex.jl
+++ b/src/Convex.jl
@@ -67,4 +67,5 @@ include("atoms/logdet.jl")
 
 ### utilities
 include("utilities/show.jl")
+include("utilities/deprecated.jl")
 end
--- a/src/atoms/norm.jl
+++ b/src/atoms/norm.jl
@@ -8,8 +8,7 @@ norm_fro(x::AbstractExpr) = norm_2(vec(x))
 # behavior of norm should be consistent with julia:
 # * vector norms for vectors
 # * operator norms for matrices
-# * norm(x, :fro) == vecnorm(x, 2)
-function norm(x::AbstractExpr, p=2)
+function norm(x::AbstractExpr, p::Number=2)
   if length(size(x)) <= 1 || minimum(size(x))==1
     # x is a vector
     if p == 1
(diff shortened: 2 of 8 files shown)


mlubin:
What does CVX do here?
Tons of test failures on travis.

madeleineudell:
Sorry about that! Should be fixed now. 

The only remaining test failures are from inaccuracies in our reshaping atoms (`reshape` and `vcat`), which I haven't touched. @davidlizeng, any idea why these are happening?

madeleineudell:
All the tests pass on the release, but fail on the nightly build.

jennyhong:
Tested locally on v0.3.8 and v0.4 after merging with master, and it worked fine. Merged with master now!

PR #77, opened 2015-05-11
compat fixes for 0.4

mlubin:
Can now run tests to completion on 0.4, but there are still numerical errors.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,2 +1,3 @@
 julia 0.3
 MathProgBase 0.3.8
+Compat
--- a/src/Convex.jl
+++ b/src/Convex.jl
@@ -1,4 +1,5 @@
 module Convex
+using Compat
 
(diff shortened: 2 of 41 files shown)


mlubin:
I have tests passing on 0.4 with the patch from JuliaLang/julia#11236

karanveerm:
This looks good, I was trying to fix this too locally but kept running into segfaults with Julia 0.4 :( 

Let's wait for the Julia PR to be accepted before we merge, although I think we should move away from hashing because of collisions completely (going to open a separate issue for that).

mlubin:
Merged now, though we may want to wait a couple of days until the fix gets pushed through the binaries and we get Travis passing.

mlubin:
All green now

mlubin:
Could we tag a new version to get Convex passing on the package evaluator?

jennyhong:
Will do!

PR #81, opened 2015-05-23
update quad_form to handle (slightly) complex eigenvalues

dressel:
I was running into problems with quad_form. My A matrix passed the is_symmetric test (tolerance of 1e-6), but the eigenvalues had very small complex parts. The pairwise compare operator in Julia does not like this, so it would choke. I just added a call to real around the eigenvalues. This happens after the check for symmetric matrices, so it should only happen when numerical issues cause small complex parts.

### DIFF
--- a/src/atoms/second_order_cone/quad_form.jl
+++ b/src/atoms/second_order_cone/quad_form.jl
@@ -16,7 +16,7 @@ function quad_form(x::AbstractExpr, A::Value)
   if !is_symmetric(A)
     error("Quadratic form only defined for symmetric matrices")
   end
-  V = real(eigvals(full(A)))
+  V = eigvals(Symmetric(full(A)))
 
   if all(V .>= 0)
     factor = 1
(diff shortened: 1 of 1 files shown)


IainNZ:
Is it possible to avoid the numerical issues in the first place by doing something like `eigvals(Symmetric(full(A)))`? Not sure if that users a more stable method

madeleineudell:
@dressel, could you check whether @IainNZ 's suggestion works for the matrices that have been giving you trouble? it seems like a more clear solution, so long as it does what we expect.

tseaward:
Hi, I have a similar(?) issue where the first eigenvalue from `eigvals(full(A))` is slightly negative (-3.70696e-7). Would a tolerance on this make sense?

mlubin:
What does CVX do in cases like this?

dressel:
@IainNZ @madeleineudell I apologize for the delay, it has been a hectic week. Iain's suggestion works with the matrices that have given me trouble, and I've updated the pull request.

madeleineudell:
@dressel: Looks great. Thanks!

@tseaward: I'm afraid that takes us down a slippery slope towards pretending indefinite matrices are positive semidefinite. And I guess that's something that a user could correct by modifying the matrix ahead of time... Does that work in your case?

tseaward:
@madeleineudell it does, thanks.

PR #89, opened 2015-06-26
removed underscores from all (external) atom names)

### LINKED ISSUE #87: simplify atom names
I don't like typing underscores. Furthermore, Julia's convention is not to use them except when they are necessary for names to be comprehended. It would also be easier to remember atom names if they all adhere to the same (no underscore) convention. Hence, I suggest the following changes:
- `lambda_max` -> `lambdamax`
- `lambda_min` -> `lambdamin`
- `sum_largest` -> `sumlargest`
- `geo_mean` -> `geomean`

etc. In fact, I don't see a single atom that truly _needs_ its underscore. Any objections?

madeleineudell:
This PR removes underscores from all atom names. Closes #87.

We need to recompile the docs before merging this.


madeleineudell:
@karanveerm @davidlizeng seems like the tables of operations are disappeared in the compiled docs after this commit. any idea why?

madeleineudell:
also need to deprecate the previously valid methods, rather than simply removing them.

PR #92, opened 2015-07-06
WIP: Presolve

madeleineudell:
This is a work in progress to implement a (fast) presolve routine to reduce the size of the problems Convex.jl sends to its solvers. 

This branch adds a keyword argument to `solve!` allowing the user to turn on the presolve (off by default). The presolve routine, in `presolve.jl`, eliminates equality constraints involving two variables or fewer, which can be done without reducing the sparsity of the system. However, it does change the sparsity _pattern_, which makes the current implementation extremely inefficient.

Before this is ready to be merged to master, 
1. We'll need to make the performance _much_ faster. We can test performance using `test_presolve_perf.jl`. 
2. We might also add other kinds of presolve reductions. See, for example, [this paper](http://www.turma-aguia.com/davi/doc/Andersen.pdf) by Anderson and Anderson.

### DIFF
--- a/src/presolve.jl
+++ b/src/presolve.jl
@@ -7,10 +7,6 @@
 # * don't give up on eliminating integer / boolean variables
 # * eliminate variables and constraints not connected to the objective
 
-# * integers
-# * frob norm
-# * inv_pos optval
-
 function presolve(c, A, b, constrcones, varcones, vartypes, verbose=false)
 	verbose && @show c, full(A), b, constrcones
 	m, n = size(A)
--- a/src/presolve.jl
+++ b/src/presolve.jl
@@ -1,0 +1,113 @@
+# To do
+
+# * consider how to populate dual variables corresponding to eliminated constraints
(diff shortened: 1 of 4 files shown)


mlubin:
I think this calls for a very clever abstraction layer to make the sparse matrix transformations fast.

src/presolve.jl
```diff
@@ -0,0 +1,159 @@
...
+						iv2 = rownnz[2]
+					end
+					bc = tb[iconstr]
+					Acv1 = tA[iconstr, iv1]
+					Acv2 = tA[iconstr, iv2]
+					Av2 = tA[:,iv2]
+					cv1 = tc[iv1]
+					cv2 = tc[iv2]
+					# Acv1*v1 + Acv2*v2 = bc => v2 = bc/Acv2 - Acv1/Acv2*v1
+					P[iv2, end] = bc / Acv2
```
mlubin: Probably need to be careful about dividing by very small values here

PR #97, opened 2015-08-08
Fix Convex on Julia 0.4

IainNZ:
~~This doesn't get Convex passing on 0.4, but, it gets it a bit further.~~

I didn't even know the original behavior worked for that `any` in `dot_sort`,
```
# Julia 0.3 behavior:
#  all((1,3)) --> 0
#  all((1,1)) --> 1
#  all((4,4)) --> 4
```
but this change seems to at least match the old behavior.

### DIFF
--- a/src/atoms/affine/index.jl
+++ b/src/atoms/affine/index.jl
@@ -86,3 +86,13 @@ end
 function getindex(x::AbstractExpr, I::AbstractVector{Bool})
     return [ x[i] for i in to_index(I) ]
 end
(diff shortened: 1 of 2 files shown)


IainNZ:
Apparently not quite fixed, I must need to update my local 0.4. Still better than current situation though I guess.

madeleineudell:
@IainNZ , I don't think the previous weird behavior was intentional... just changed it to vectorize any rectangular array.

PR #111, opened 2015-09-30
Fix deprecation warnings due to use of "int()" instead of "round(Int,)"

marcusps:
Nothing else to it.


src/constraints/sdp_constraints.jl
```diff
@@ -46,8 +46,8 @@ function conic_form!(c::SDPConstraint, unique_conic_forms::UniqueConicForms)
     # and the corresponding entries in the lower triangular part, so
     # symmetry => c.child[upperpart] 
     diagandlowerpart = find(tril(ones(n,n)))
-    lowerpart = Array(Int, int(n*(n-1)/2))
```
mlubin: Maybe just replace with ``div(n*(n-1),2)``?

marcusps: Will do.

PR #116, opened 2015-10-22
avoid splatting in problem construction

mlubin:
This sort of splatting got really slow (i.e., takes many minutes) in Julia 0.4 if you have thousands of constraints. Using ``collect`` seems to resolve the performance issue.

CC @emreyamangil


mlubin:
Travis failure is unrelated

PR #119, opened 2015-11-26
drop support for julia 0.3

mlubin:
Implementing https://github.com/JuliaOpt/MathProgBase.jl/pull/91 will require dropping 0.3 anyway, so this PR is the first piece of that. Also generally makes the code easier to maintain by removing the ``@compat`` weirdness.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ os:
     - linux
     - osx
 julia:
-    - 0.3
     - 0.4
     - nightly
 notifications:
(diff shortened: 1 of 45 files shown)


IainNZ:
`45 files changed` wow :+1:

tkelman:
Be sure to use `Pkg.tag("Convex", :minor)" when tagging the first version with this merged.

mlubin:
@madeleineudell, I'll be merging this along with the MPB changes early next week if there are no objections.

madeleineudell:
Looks good to me!

Issue #83, opened 2015-06-04
Memory leak


bodono:
Both ECOS and SCS exhibit memory leaking and crashing when running the script here: https://github.com/JuliaOpt/SCS.jl/issues/24. Creating a problem by handing and calling SCS.jl in a for-loop does not exhibit the same memory leakage (although it might be in a code-path I'm not testing), so my best guess is that it's in Convex.jl somewhere (I know julia is garbage collected, but memory leaks can still occur).

madeleineudell:
@bodono, thanks for finding this. But I'm not entirely sure how to go about looking for the problem in a garbage collected language...

mlubin:
[Module-level variables](https://github.com/JuliaOpt/Convex.jl/blob/067ab04d3a28ea3a856d0a33ba9a5c52f45f4e60/src/variable.jl#L46) could do it.

pluskid:
I need to run several thousand small linear programs in an iterative algorithm. The memory quickly exploded (with various solvers including Clp, ECOS, GLPK, etc.). Currently I have run for a few hundred iterations, save the results into file, exit Julia and then start a new session, load the saved results and continue running.

mlubin:
@pluskid, any reason you haven't tried JuMP?

pluskid:
@mlubin Thanks! No particular reason. I just arrived at juliaopt.org, found JuMP.jl and Convex.jl. I picked the second one by applying the first impression that Convex.jl is more dedicated to convex problems and the stereotype that dedicated packages typically runs faster. 

I will try JuMP later. :)

daviddelaat:
The global variables are indeed the problem. I submitted a pull request (https://github.com/JuliaOpt/Convex.jl/pull/123) with a temporary workaround. With this merged you have to call `Convex.clearmemory()` every now and then.

madeleineudell:
David, thanks for the PR! And for confirming that the global variables are at fault.

Any ideas on the best permanent solution? The global cache of variables is used to avoid extra work when variables and expressions are reused, even on different problems. The problem is that even when all other pointers to the variable or expression have evaporated, the pointer in this global cache remains, preventing garbage collection. 

Is there any way to check whether the *only* pointer to the julia object (variable or expression) is coming from this dictionary, at which point we can delete it from the dictionary as well to pave the way for garbage collection?

mlubin:
[...]

PR #123, opened 2016-02-10
Temporary fix for memory leaks

daviddelaat:


mlubin:
Not ideal but better than crashing

src/Convex.jl
```diff
@@ -74,4 +74,13 @@ include("atoms/logdet.jl")
 include("utilities/show.jl")
 include("utilities/iteration.jl")
 include("utilities/deprecated.jl")
+
+#Temporary workaround for memory leak
```
mlubin: Could you add a reference to https://github.com/JuliaOpt/Convex.jl/issues/83 here?

daviddelaat:
Yes, this is not ideal (and I think a fix is not too hard), but for now it at least enables us to solve many problems in one session. I added the ref to the issue.

PR #128, opened 2016-03-08
Added solution Unconstrained Markowitz Efficient Frontier example

Ayush-iitkgp:
Added more example for the unconstrained efficient frontier for the objective function 
**minimize**       lambda * risk - (1-lambda) * return         
where         0< lambda <=1

By varying the value of lambda from 0 to 1 in the steps of 1/2000, we obtain the unconstrained efficient frontier also known as Markowitz Efficient Fromtier.

The next obvious step is to add the example for the **cardinality constrained efficient frontier** where we limit the number of assets in which the portfolio manager can invest out of the given assets. This problem is Mixed Integer Quadratic Programming Problem.


madeleineudell:
Nice!

PR #141, opened 2016-06-01
Correcting diag atom

Ayush-iitkgp:
x = Variable(4,5)
Presently
y_1 = diag(x,4)
AbstractExpr with
head: diag
size: (0, 1)
sign: Convex.NoSign()
vexity: Convex.AffineVexity()

does not give error but it should be giving ideally as we can't have the kth diagonal where k is greater than or equal to both the number of rows and columns.

This PR fixes the present bug.
y_2 = diag(x,4)
**ERROR: Bounds error in calling diag**

### DIFF
--- a/src/atoms/affine/diag.jl
+++ b/src/atoms/affine/diag.jl
@@ -20,7 +20,7 @@ type DiagAtom <: AbstractExpr
   function DiagAtom(x::AbstractExpr, k::Int=0)
     (num_rows, num_cols) = x.size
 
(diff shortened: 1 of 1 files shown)


Ayush-iitkgp:
@madeleineudell is this check fail something to worry about ?

madeleineudell:
Nope, the check fail is no problem: if you click through, you'll see it's failing on the nightly Julia build, but succeeding on 0.4, so probably it's a problem with the nightly build :)

PR #144, opened 2016-06-15
Dot

madeleineudell:


mlubin:
This is usually called ``vecdot`` when the arguments aren't vectors, but convex.jl is free to overload ``dot`` as well.

madeleineudell:
I will happily bow to convention. How does this sound:

* change the name of `dot` to `vecdot`
* have `dot` point to `vecdot` for vector arguments
* `dot` for matrix arguments is an error

?

mlubin:
Sounds good!

madeleineudell:
Works on 4.0

PR #147, opened 2016-07-01
[WIP] Added partial trace

thomasschiet:


madeleineudell:
This looks great! Thanks for the contribution. 

Any chance you could put together a test and a line of documentation on it? The test should be added right below the [trace test](https://github.com/JuliaOpt/Convex.jl/blob/master/test/test_affine.jl#L196). I'll add the [documentation](http://convexjl.readthedocs.io/en/latest/operations.html) if you'd give a quick description of the dimensions of x, sys, and dim are, and what they do.

madeleineudell:
@thomasschiet, repinging: could you tell me mathematically what this atom does?

thomasschiet:
Sorry I missed the previous message! So the partial trace is a map `Tr_W: L(V ⊗ W) -> L(V)` defined as `Tr_W(R ⊗ S) = R Tr(S)`.

This can be computed by summing over  `(Id ⊗ e_j) R ⊗ S (Id ⊗ e_j')` for `j=1...n`, where `n` is the dimension of `S`, `Id` is the identity matrix on `R` and `e_j` are the standard basis vectors relevant for S. Of course, this could be generalised by considering the partial trace over an arbitrary system in the mapping:
`Tr_{W_j}: L(W_1 ⊗ ... ⊗ W_n) -> L(W_1 ⊗ ... ⊗ W_{j-1} ⊗ W_{j+1}  ⊗ W_n).`
That's what is done in this atom.
- `sys` refers to the index  of the system to trace out and should be an integer.
- `dims` refers to the sizes of the `W_j`'s and should be a vector of integers. 
- `x` is the matrix of which we want to take the partial trace. Its dimension should be the product of the entries in `dim`. 

I will also have a look at the test! Also, I see that I also made a mistake in the vexity as I think it should be affine. I am also not entirely sure about the sign and monotonicity.

madeleineudell:
The sign, vexity, and curvature are all fine. This is ready to merge when tests and documentation are added.

peterwittek:
I am having some trouble figuring out what this actually does. I made a toy example of a product state in C²⊗C², but using the partial trace, I get infeasibility: 
```julia
ρ = Variable(4, 4)
constraints = [partialtrace(ρ, 1, [2; 2]) == [0.09942819 0.29923607; 0.29923607 0.90057181]
               partialtrace(ρ, 2, [2; 2]) == [1 0; 0 0]
               trace(ρ) == 1
               ρ in :SDP]
p = satisfy(constraints)
solve!(p)
```

Ayush-iitkgp:
@peterwittek could you please frame a sample test case for partial trace? I am taking up this PR and trying to get it merged. Your help would be appreciated :)

peterwittek:
I am somewhat uncertain about the syntax for the test, but something like this would do:
```julia
using Convex
using FactCheck

TOL = 1e-2

facts("Partial trace tests") do
  context("Qubit tensor product") do
    ρ = Variable(4, 4)
    constraints = [partialtrace(ρ, 1, [2; 2]) == [0.09942819 0.29923607; 0.29923607 0.90057181]
                   partialtrace(ρ, 2, [2; 2]) == [1 0; 0 0]
                   trace(ρ) == 1
                   ρ in :SDP]
    p = satisfy(constraints)
    solve!(p)
    @fact ρ.value --> roughly([0.09942819 0. 0.29923607 0.; 0. 0. 0. 0.; 0.29923607 0. 0.90057181  0.; 0. 0. 0. 0.], TOL)
  end
 
end
```

src/atoms/affine/partialtrace.jl
```diff
@@ -0,0 +1,85 @@
...
+	function PartialTraceAtom(x::AbstractExpr, sys::Int, dims::Vector)
+		if x.size[1] ≠ x.size[2]
+			error("Only square matrices are supported")
+		end
+
+		if ! (1 ≤ sys ≤ length(dims))
+			error("Invalid system, should between 1 and ", length(dims), " got ", sys)
+		end
+
+		if x.size[1] ≠ prod(dims)
```
peterwittek: Since x is the tensor product of the two matrices, this condition checks whether the dims of the subsystems is correct.

src/atoms/affine/partialtrace.jl
```diff
@@ -0,0 +1,85 @@
...
+function conic_form!(x::PartialTraceAtom, unique_conic_forms::UniqueConicForms)
+	if !has_conic_form(unique_conic_forms, x)
+		sys = x.sys
+		dims = x.dims
+
+		# compute the partial trace of x by summing over
+		# (I ⊗ <j| ⊗ I) x (I ⊗ |j> ⊗ I) for all j's
+		# in the system we want to trace out
+		# This function returns every term in the sum
+		function term(ρ, j::Int)
```
Ayush-iitkgp: What's **j** here?

peterwittek:
This almost works:
```julia
A = Semidefinite(2)
B = [1 0; 0 0]
facts("Partial trace tests") do
  context("Qubit tensor product") do
    ρ = kron(A, B)
    constraints = [partialtrace(ρ, 1, [2; 2]) == [0.09942819 0.29923607; 0.29923607 0.90057181]
                   ρ in :SDP]
    p = satisfy(constraints)
    solve!(p)
    @fact ρ.value --> roughly([0.09942819 0. 0.29923607 0.; 0. 0. 0. 0.; 0.29923607 0. 0.90057181  0.; 0. 0. 0. 0.], TOL)
  end
end
```

Ayush-iitkgp:
@peterwittek ρ is no more a variable, it holds the output of the kron product therefore we it does not have value field (only the variables have **value** field) `evaluate(ρ)` works instead (this is the reason why we need to define evaluate method for each operatin).

evaluate(ρ) outputs:
4×4 Array{Float64,2}:
 0.0994306  0.299238  0.0  0.0
 0.299239   0.90057   0.0  0.0
 0.0        0.0       0.0  0.0
 0.0        0.0       0.0  0.0

which is different from:
4×4 Array{Float64,2}:
 0.0994282  0.0  0.299236  0.0
 0.0        0.0  0.0       0.0
 0.299236   0.0  0.900572  0.0
 0.0        0.0  0.0       0.0

Is there some typo in your test or the implementation of partial trace needs to be improved?

peterwittek:
Basically, this is the state we want:
```julia
A = [0.09942819 0.29923607; 0.29923607 0.90057181]
B = [1 0; 0 0]
kron(A, B)
```
which gives you
```julia
4×4 Array{Float64,2}:
 0.0994282  0.0  0.299236  0.0
 0.0        0.0  0.0       0.0
 0.299236   0.0  0.900572  0.0
 0.0        0.0  0.0       0.0
```
In the meantime, I tested the function `term`, it works just fine:
```julia
function term(ρ, sys::Int, j::Int, dims::Vector)
    bra = speye(1)
    ket = speye(1)
    i_sys = 1
    for dim in dims
        if i_sys == sys
            # create a vector that is only 1 at its jth component
            v = spzeros(dim, 1);
            v[j] = 1;
            bra = kron(bra, v')
            ket = kron(ket, v)
        else
            bra = kron(bra, speye(dim))
            ket = kron(ket, speye(dim))
        end
        i_sys += 1
    end
    return bra * ρ * ket
end
A = [0.09942819 0.29923607; 0.29923607 0.90057181]
B = [1 0; 0 0]
println(A == sum(term(kron(A, B), 2, j, [2; 2]) for j in 1:2))
println(B == sum(term(kron(A, B), 1, j, [2; 2]) for j in 1:2))
```
So it is really weird that when you try to solve the above problem, it reshuffles the subsystems.

Ayush-iitkgp:
Hello @peterwittek , I just found out that the first argument in kron will be a value and the second would be a expression.
In such case, the test would become:
```
A = Semidefinite(2)
B = [1 0; 0 0]
facts("Partial trace tests") do
  context("Qubit tensor product") do
    ρ = kron(B, A)
    constraints = [partialtrace(ρ, 1, [2; 2]) == [0.09942819 0.29923607; 0.29923607 0.90057181]
                   ρ in :SDP]
    p = satisfy(constraints)
    solve!(p)
    @fact evaluate(ρ) --> roughly([0.09942819 0.29923607 0 0; 0.299237 0.900572 0 0; 0 0 0  0; 0 0 0 0], TOL)
  end
end
```

Am I right?

ericphanson:
Closing since #183 implements this

PR #148, opened 2016-07-08
don't return elementwise square for matrix A*A

tkelman:
`*` is matrix multiplication, not elementwise

todo: add a test


madeleineudell:
Thanks for the catch, Tony!

tkelman:
Looks like I need to fix up a few of the tests. Should've marked this WIP, I made this change from a phone and hadn't run tests locally first, sorry.

PR #151, opened 2016-07-10
Reorganize atoms

madeleineudell:
Reorganized atoms into folders. 

Added methods for ./ and / when the divisor is a variable. Added method .^2 to perform elementwise square.

Fixed deprecated syntax in tests.


madeleineudell:
tests pass on .4

Issue #145, opened 2016-06-25
broken on julia 0.5


mlubin:
Convex.jl will be broken on 0.5 until some thought is put into the concatenation behavior and fixing [this](https://github.com/JuliaOpt/Convex.jl/blob/ebd1acb80eb97875550b6b64222da3d0675319fb/src/atoms/affine/stack.jl#L122). Now ``[1,2,3]`` lowers to ``Base.vect(1,2,3)`` instead of ``vcat(1,2,3)``. I spent a bit of time trying to figure out a quick fix but couldn't come up with anything.

madeleineudell:
Miles, I think I understand what's going on now. 

1) We can't overload hcat and vcat without clobbering the default methods from base, which we cleverly get around by rerouting them to cat(., 1) and cat(., 2). It's not great, but not terrible. 

2) You're saying that there's no similar way to find a substitute for vect. So instead, you substituted vect by vcat, and that worked ok despite being ugly.

3) @andreasnoack and I worked on this problem yesterday. We tried and failed to find a nicer solution: we could do it if we could specify variable length arguments at the beginning of a call as well as at the tail - but the language doesn't support it. We could solve the problem by recursively breaking off the first argument until the first argument is an AbstractExpr, but would make these functions *super* slow (n recursive calls) when called on non-AbstractExprs, slowing down all of julia after loading Convex. 

4) Just deleting the version check in your hack seemed to work for us, although it's still ugly, and reroutes vect to vcat. Do you know what the bad consequences of that are? 

At this point, I'm just looking to minimize collateral damage in the way we fix this issue.

Madeleine

mlubin:
> Just deleting the version check in your hack seemed to work for us, although it's still ugly, and reroutes vect to vcat. Do you know what the bad consequences of that are?

Yes, it globally breaks julia:
```julia
julia> a = [1,2,3];

julia> [a,1]
2-element Array{Any,1}:
  [1,2,3]
 1       

julia> Base.vect(args...) = Base.vcat(args...)
WARNING: Method definition vect(Any...) in module Base at abstractarray.jl:24 overwritten in module Main at REPL[3]:1.

julia> [a,1]
4-element Array{Int64,1}:
 1
 2
 3
 1

```

mlubin:
What about making ``vect`` a generated function? You could check if any of the arguments are an ``AbstractExpr`` and if not then fall back to the base definition.

madeleineudell:
[...]

PR #154, opened 2016-08-09
fix for julia 0.5

mlubin:
I might be having a brain fart, but does this address #145? Tests seem to pass locally

### DIFF
--- a/src/atoms/affine/stack.jl
+++ b/src/atoms/affine/stack.jl
@@ -111,14 +111,15 @@ hcat(args::Value...) = Base.cat(2, args...)
 vcat(args::AbstractExpr...) = HcatAtom([arg' for arg in args]...)'
 vcat(args::AbstractExprOrValue...) = HcatAtom([convert(AbstractExpr, arg)' for arg in args]...)'
 vcat(args::Value...) = Base.cat(1, args...) # Note: this makes general vcat slower for anyone using Convex...
-if VERSION >= v"0.4.0"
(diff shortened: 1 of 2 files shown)


src/atoms/affine/stack.jl
```diff
@@ -111,14 +111,15 @@ hcat(args::Value...) = Base.cat(2, args...)
 vcat(args::AbstractExpr...) = HcatAtom([arg' for arg in args]...)'
 vcat(args::AbstractExprOrValue...) = HcatAtom([convert(AbstractExpr, arg)' for arg in args]...)'
 vcat(args::Value...) = Base.cat(1, args...) # Note: this makes general vcat slower for anyone using Convex...
-if VERSION >= v"0.4.0"
-    Base.vect{T<:AbstractExpr}(args::T...) = HcatAtom([arg' for arg in args]...)'
-    Base.vect(args::AbstractExpr...) = HcatAtom([arg' for arg in args]...)'
-    Base.vect(args::AbstractExprOrValue...) = HcatAtom([convert(AbstractExpr,arg)' for arg in args]...)'
-    if Base._oldstyle_array_vcat_
```
madeleineudell: Do we still need this check?

mlubin: Yes, for compatibility with 0.4

madeleineudell:
This looks great! Check whether you want to remove the check for old style vcat, and then go ahead and merge!

PR #165, opened 2016-10-19
SCS verbosity

EQt:
if verbose=false tell it to the SCS solver


madeleineudell:
We try to observe a separation between the modeling language code (Convex) and the solver code (eg, SCS), so I'd like to preserve the ability to turn off verbosity in Convex without turning it off in the solver.

Issue #168, opened 2016-11-09
Partial transpose


peterwittek:
The `HermitianSemidefinite` variable is a phenomenal addition for the quantum information community. It would even be more useful if `kron` was extended to it, and I also wonder if you plan to implement the partial trace and partial transpose functions. Thanks.

madeleineudell:
These functions are not familiar to me, but I'd welcome PRs! Even PRs with
pure documentation: we'd have a much easier time implementing them given a
concise mathematical description of these functions.

On Nov 9, 2016 11:55 AM, "Peter Wittek" <notifications@github.com> wrote:

The HermitianSemidefinite variable is a phenomenal addition for the quantum
information community. It would even be more useful if kron was extended to
it, and I also wonder if you plan to implement the partial trace and
partial transpose functions. Thanks.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/JuliaOpt/Convex.jl/issues/168>, or mute the thread

peterwittek:
Thank you. I just started to pick up Julia, so even the elementary task of writing the pure documentation will take me a while. The kron function would be the same as the basic linear algebra operation [implemented in Julia](http://docs.julialang.org/en/release-0.5/stdlib/linalg/#Base.kron). We implemented the [partial trace](http://picos.zib.de/tools.html#picos.tools.partial_trace) and the [partial transpose](http://picos.zib.de/tools.html#picos.tools.partial_transpose) in [PICOS](http://picos.zib.de/) in Python. I would love to see these functions in Convex.jl, and I will do my best to help in my limited capacity.

Ayush-iitkgp:
Hi @peterwittek thanks for appreciating our work. We implemented the support for complex variables as part of my GSoC'16 project under the mentorship of @madeleineudell and @dvij. Kron and partial trace are in my to-do list too. I am busy looking for a job these days hence can not fulfill your needs immediately but yes certainly in a month or so, I will come back and implement those functions with your help.

peterwittek:
Many thanks, that would be awesome! In that time span, I hope to pick up some Julia skills to be more useful.

Ayush-iitkgp:
Hi @peterwittek are we ready to get going?

peterwittek:
Absolutely. I think the easiest start is the partial transpose. This bit of Python code gives a reference 

```python
import numpy as np
from qutip import partial_transpose, Qobj
[...]

PR #171, opened 2017-01-06
Phase recovery using relaxation similar to MaxCut 

Ayush-iitkgp:
Adding more applications of optimization with complex variables.
@peterwittek and @dvij I hope I have solved the problem correctly.
I also request the experts in complex-domain optimization to add more examples before we release the new version of the package.


peterwittek:
Quantum steering is one area where having complex Hermitian matrices is crucial. There are a handful of examples in MATLAB in [this repo](https://github.com/paulskrzypczyk/steeringreview). This, unfortunately, needs #168 resolved first.

A much simpler is in [this Python notebook](https://github.com/peterwittek/ipython-notebooks/blob/master/Simulating_POVMs.ipynb). This can be done now. I will adapt the qubit case and send a pull request.

peterwittek:
Done: PR #172 implements the qubit POVM simulation problem.

madeleineudell:
@peterwittek , another very useful contribution would be to write tests and documentation for #147 - I think that's all that's required for us to have a partial trace implementation.

Ayush-iitkgp:
Suggestion incorporated.

madeleineudell:
Thanks, Ayush! 

Also, it would be nice to motivate why that SDP recovers the answer. I've usually seen a different setup for the phase retrieval problem. we minimize sum_{i=1}^m (|<a_i, u>|^2 - b_i)^2 . We use the fact that |<a_i, u>|^2 = (u^* a_i) (a_i^* u) = Tr[ (u^* a_i) (a_i^* u) ] = Tr[ a_i a_i^* u u^* ]. Then let U = u u^*, so that sum_{i=1}^m (|<a_i, u>|^2 - b_i)^2 = sum_{i=1}^m (Tr[ a_i a_i^* U ] - b_i)^2 is convex in U!

You should actually be able to get exact recovery of the input signal if you have more measurements than the dimension of the signal. In other words, b is an m-vector, and A is a m x n matrix, with m > n. Using m = 4n should be sufficient.

You can generate problem data by picking measurement vectors a_i to be random n-vectors, picking a true signal x to be a random n-vector, and setting each coordinate b_i = |<a_i, x>|^2 + small gaussian noise. See if you find that u is close to x!

Ayush-iitkgp:
@madeleineudell, the original phase retrieval problem is:
>>>> find x

>>>> such that |Ax| = b

>>>> where x in {C}^p, A in C}{n×p} and b in R}^n

After going the research paper, I found that we should rather assume b and then find u using formulation above and then construct x.
I didn't get what you meant by "u is close to x". They have different dimensions. And if you meant
`atan(sum(imag(u))/sum(real(u))) == atan(sum(imag(x))/sum(real(x)))`, they don't seem to be close enough. I don't know where is the mistake that I am commiting

Ayush-iitkgp:
ping @madeleineudell

Ayush-iitkgp:
@madeleineudell, I am sorry for the misunderstanding. All this time we have been solving different problems. **This example is to retrieve the phase of Ax given it's magnitude**. We do not intend to find the original signal in this example. It uses the classical [MaxCut](http://www-math.mit.edu/~goemans/PAPERS/maxcut-jacm.pdf) relaxation to form a complex semidefinite programming problem.

In coming days, I would create another PR to retrieve the original signal using the methodology defined above.

madeleineudell:
Ayush, thanks for the clarification! The writeup is wonderfully clear on this iteration, and I think I understand now. There is one thing you could do to make the solution even more nice. So far,

1. you start by generating an A and x
2. you solve for u
3. you hope that the phase of u is the phase of Ax

why don't you check?

4. compute norm( mod( phase (u) - phase (Ax), 2*pi ) )? 

(That's pseudocode; you'll want to figure out how to compute phase and which kind of norm is best.) If the problem is big enough and nicely conditioned, we should see that this number is small.

Ayush-iitkgp:
`phase_u = atan(sum(imag(u)/sum(real(u))))
`
`phase_Ax = atan(sum(imag(A*x)/sum(real(A*x))))`

I am not sure but I am using **atan** to calculate the phase. For some random values of A and x, u and Ax point to the same direction but for some values they are not.

madeleineudell:
Try a bigger example (more observations, ie b gets longer) and see if the
fraction of correct answers increases?

Ayush-iitkgp:
@madeleineudell I think we should merge this PR. This example is just intended to show how to use our implemnetation. I think we can improve it once domain specific people start taking interest in our implementation. I want a new version release for Convex.jl and publicize it and this unmerged PR is one of the reasons why I am unable to proceed. I am planning to add one or two more examples and then we are ready for new Convex.jl version.

madeleineudell:
Ayush, sounds like a fine plan.

Just a few typos first:

* the magnitude vector |Ax|, not the phase of Ax
* \mathbb{C}^{n×p} should be \mathbb{C}^{n \times p}

On Fri, Mar 17, 2017 at 12:13 PM, Ayush Pandey <notifications@github.com> wrote:
> @madeleineudell I think we should merge this PR. This example is just
> intended to show how to use our implemnetation. I think we can improve it
> once domain specific people start taking interest in our implementation. I
> want a new version release for Convex.jl and publicize it and this unmerged
> PR is one of the reasons why I am unable to proceed. I am planning to add
> one or two more examples and then we are ready for new Convex.jl version.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or mute the thread.



-- 
Madeleine Udell
Assistant Professor, Operations Research and Information Engineering
Cornell University
https://people.orie.cornell.edu/mru8/
(415) 729-4115

Ayush-iitkgp:
Typo fixed.

PR #176, opened 2017-02-03
Explicitly provide stable and latest docs links.

pkofod:
This is how we do it in Optim, and it can avoid this problem: https://discourse.julialang.org/t/convex-optimisation-over-complex-variables/1388 when the front page readme is the "latest" readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
 
 **Installation**: `julia> Pkg.add("Convex")`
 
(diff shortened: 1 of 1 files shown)


Ayush-iitkgp:
I encountered a similar problem when I was new to Github.
+1

madeleineudell:
Beautiful. Thanks!

PR #182, opened 2017-03-10
Norm_1 correction

Ayush-iitkgp:
As a result of #181.


madeleineudell:
Excellent! Glad we found this bug.

PR #185, opened 2017-03-12
[WIP]: Adding Partial transpose atom

Ayush-iitkgp:
This is an attempt to add partial transpose.
@peterwittek  @madeleineudell any idea how we can leverage the partial trace implementation to get the partial transpose implementation?


peterwittek:
It shouldn't transpose all subsystems, only the requested one, and then kron the components. In any case, I think the [mechanism](https://github.com/qutip/qutip/blob/master/qutip/partial_transpose.py) I [mentioned](https://github.com/JuliaOpt/Convex.jl/issues/168#issuecomment-285012144) earlier is easier.

Edenharder:
@Ayush-iitkgp It seems that `Partial transpose` is still missing in the new release of `Convex.jl`.

ericphanson:
Closing in favor of #316

PR #188, opened 2017-03-17
Adding Fidelity in Quantum Information Theory example

Ayush-iitkgp:
This example is inspired from a lecture of John Watrous in the course on Theory of Quantum Information


madeleineudell:
Cool!

PR #205, opened 2017-07-19
Fixed robustness of test in "test_params.jl"

mfalt:
The regularization-parameters in "fix multiplication" was badly chosen, resulting in numerical problems (`o2-o1  ≈ 3e-12` when `eps=1e-12`was set for SCS). The regularization only works reasonable in the range 0.8 to 0.85 as shown in the plot. I changed the test to reflect the expected behavior for larger and smaller regularization parameters.
Plot


madeleineudell:
Cool, thanks!

PR #206, opened 2017-07-20
Avoid OutOfMemoryError

mfalt:
These lines causes `OutOfMemoryError` on large problems because of the broadcast (I believe sparse broadcast is still suffering from performance issues). The proposed change uses an explicit iterator over the elements instead. The problem is easily reproduced by, for example:
```julia
n = 50000
x0 = randn(2n)
A = sprandn(n, 2n, 0.0001)
b = A*x0
x = Variable(2n)
p = minimize(norm(A*x-b), sum(x) == sum(x0))
```
all tests passes locally on
Julia Version 0.6.1-pre.0
Commit dcf39a1* (2017-06-19 13:06 UTC)

### DIFF
--- a/src/constant.jl
+++ b/src/constant.jl
@@ -22,9 +22,9 @@ type Constant <: AbstractExpr
     if check_sign
       if !isreal(x)
         return Constant(x, ComplexSign())
-      elseif all(x .>= 0)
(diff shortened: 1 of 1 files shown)


madeleineudell:
great, thanks.

PR #221, opened 2018-01-14
updated struct and type parameter syntax to julia 0.6

mbesancon:
mostly removed deprecated syntax  
* type X 
* f{T}(x::T)

Deleted exported but undefined expressions:
* VcatAtom
* domain

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ os:
     - linux
     - osx
 julia:
-    - 0.5
     - 0.6
     - nightly
 notifications:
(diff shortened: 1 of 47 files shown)


mlubin:
Please bump the minimum version to 0.6 and update the travis config.

mlubin:
`struct` is not a drop-in replacement for `type` (try `mutable struct`).

mbesancon:
Sorry, I had forgotten type was converted to both. I added `mutable` based on the tests

PR #222, opened 2018-01-27
use testsets

jw3126:
Replace `FactCheck` by testsets.

### DIFF
--- a/test/test_socp.jl
+++ b/test/test_socp.jl
@@ -1,242 +1,242 @@
 using Convex
-using FactCheck
+using Base.Test
 
 TOL = 1e-3
 
-facts("SOCP Atoms") do
+@testset "SOCP Atoms" begin
 
-  context("norm 2 atom") do
+  @testset "norm 2 atom" begin
     x = Variable(2, 1)
     A = [1 2; 2 1; 3 4]
     b = [2; 3; 4]
     p = minimize(norm2(A * x + b))
-    @fact vexity(p) --> ConvexVexity()
+    @test vexity(p) == ConvexVexity()
     solve!(p)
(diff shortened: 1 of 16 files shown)


test/test_socp.jl
```diff
@@ -1,243 +1,243 @@
...
     x.value = A
-    @fact evaluate(norm(x)) --> roughly(norm(A), TOL);
-    @fact evaluate(norm(x, 1)) --> roughly(norm(A, 1), TOL);
-    @fact evaluate(norm(x, 2)) --> roughly(norm(A, 2), TOL);
-    @fact evaluate(norm(x, Inf)) --> roughly(norm(A, Inf), TOL);
-    @fact evaluate(vecnorm(x, 1)) --> roughly(norm(vec(A), 1), TOL);
-    @fact evaluate(vecnorm(x, 2)) --> roughly(norm(vec(A), 2), TOL);
-    @fact evaluate(vecnorm(x, 7)) --> roughly(norm(vec(A), 7), TOL);
-    @fact evaluate(vecnorm(x, Inf)) --> roughly(norm(vec(A), Inf), TOL);
+    $(Expr(:toplevel, :(@test isapprox(evaluate(norm(x)), norm(A), atol=TOL))))
```
mlubin: Why does this need to be wrapped in `toplevel`? Shouldn't `@test isapprox(evaluate(norm(x)), norm(A), atol=TOL)` work?

jw3126: Good catch. This was not intended! The reason is that this was done by a hacky script that proceeds line by line and `parse("foo;")` (note the `;`) gives `Expr(:toplevel, :foo)`.

PR #237, opened 2018-10-24
Seed random number stream to improve test reliability

ararslan:
We've seen some inconsistent test failures, seemingly always involving tests that use values computed via `rand`. To make these tests more reliable, we can seed the stream using `srand`. Note that it's `Random.seed!` on 0.7 and up, but Convex's current master branch does not support anything other than 0.6. On a related note, this also removes Travis testing on Julia master, as there is no chance of it working as-is.

I will re-tag v0.7.0 once this is merged. The 0.6 failure on JuliaCIBot for the current v0.7.0 tag failed due to these issues.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,13 @@ os:
     - osx
 julia:
     - 0.6
-    - nightly
+    - 0.7
 notifications:
     email: false
+matrix:
+    allow_failures:
+        - julia: 0.7
+    fast_finish: true
 sudo: false
 addons:
     apt_packages:
--- a/test/REQUIRE
+++ b/test/REQUIRE
@@ -1,3 +1,3 @@
 GLPKMathProgInterface
 ECOS
-SCS
+SCS 0.3.0 0.4.0
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,6 +1,9 @@
 using Convex
 using Base.Test
 
+# Seed random number stream to improve test reliability
+srand(2)
+
 solvers = Any[]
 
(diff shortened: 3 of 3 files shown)


ararslan:
Amusingly, every seed I've tried seems to produce a different set of test failures.

mlubin:
Feel free to eyeball the answers and tighten solver convergence tolerances or loosen equality tests as needed. I don't think they were set in any principled way.

ararslan:
As it turns out, the problem is with the version of SCS used. Pinning SCS at 0.3.3 allows all tests to pass successfully, whereas using 0.4.0 or later causes them to fail.

mlubin:
The default parameters changes between SCS releases. You may need to tighten SCS's convergence tolerances to get similar results as before.

ararslan:
It seems that setting the convergence tolerance from 1e-5 to 1e-12(!!) fixes the tests. Any idea why such a huge jump would be necessary?

blegat:
I have encountered situations where setting the tolerance too tight makes the solver fail on occasions.

ararslan:
We've checked with a few convergence tolerances for SCS, and 1e-12 was the largest that got the tests working at all as-is on Julia 0.6.

I've spoken with Miles and for now we'll leave the band-aid solution of pinning SCS to an older version for the purpose of testing on Julia 0.6, then release a (likely final) Julia 0.6-compatible version, and the root cause of the failures can be address amidst updates for Julia 0.7 and later.

Does anyone have any qualms with that plan?

blegat:
> Does anyone have any qualms with that plan?

Looks good to me

PR #242, opened 2018-11-06
Fix use of norm and remove vecnorm

### LINKED ISSUE #231: norm(matrix) -> opnorm(matrix) in Julia 0.7
It looks like may be using `norm(matrix)`.  In Julia 0.7, this will compute the Frobenius norm (`vecnorm` in Julia 0.6), due to JuliaLang/julia#27401.   If you want the induced/operator norm as in Julia 0.6, use `opnorm(matrix)` instead, or `Compat.opnorm(matrix)` to work in 0.6 and 0.7 (JuliaLang/Compat.jl#577).

Note that, for testing purposes, rather than `@test norm(A - B) ≤ tol`, it is usually preferred to do `@test A ≈ B` or `@test A ≈ B rtol=...` (which uses `isapprox`).

ararslan:
The behavior of `norm(x, p)` for matrices changed in https://github.com/JuliaLang/julia/pull/27401 to be the same as `norm(vec(x), p)` and `vecnorm` was removed. The previous behavior of `norm` for matrices is now captured by `opnorm`. This updates our extensions to these functions to match, and emits deprecation warnings as appropriate.

As a somewhat related change, this also deprecates the positional dimension argument to `sum` in favor of keyword arguments, in keeping with Base.

Properly fixes #231, which was already closed, and should fix an error in SchattenNorms observed with JuliaCIBot (see [here](https://juliarun-ci.s3.amazonaws.com/1ab1a9c14bbccf791c72724313bf4b57c7f479f9/SchattenNorms_with_pull_request_of_Convex_on_julia_0_7.log)).

### DIFF
--- a/src/atoms/affine/sum.jl
+++ b/src/atoms/affine/sum.jl
@@ -57,9 +57,13 @@ function conic_form!(x::SumAtom, unique_conic_forms::UniqueConicForms=UniqueConi
   return get_conic_form(unique_conic_forms, x)
 end
 
-sum(x::AbstractExpr) = SumAtom(x)
+# Dispatch to an internal helper function that handles the dimension argument in
+# the same manner as Base, with dims=: denoting a regular sum
+sum(x::AbstractExpr; dims=:) = _sum(x, dims)
 
-function sum(x::AbstractExpr, dimension::Int)
+_sum(x::AbstractExpr, ::Colon) = SumAtom(x)
+
+function _sum(x::AbstractExpr, dimension::Integer)
   if dimension == 1
     return Constant(ones(1, x.size[1]), Positive()) * x
   elseif dimension == 2
@@ -68,3 +72,5 @@ function sum(x::AbstractExpr, dimension::Int)
     error("Sum not implemented for dimension $dimension")
   end
 end
+
+Base.@deprecate sum(x::AbstractExpr, dim::Int) sum(x, dims=dim)
--- a/src/atoms/sdp_cone/operatornorm.jl
+++ b/src/atoms/sdp_cone/operatornorm.jl
@@ -6,7 +6,7 @@
 # Please read expressions.jl first.
 #############################################################################
 import LinearAlgebra: opnorm
-export operatornorm, sigmamax
+export sigmamax
 
 ### Operator norm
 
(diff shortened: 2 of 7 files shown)


ararslan:
cc @stevengj, who opened #231.

src/atoms/affine/sum.jl
```diff
@@ -57,9 +57,13 @@ function conic_form!(x::SumAtom, unique_conic_forms::UniqueConicForms=UniqueConi
   return get_conic_form(unique_conic_forms, x)
 end
 
-sum(x::AbstractExpr) = SumAtom(x)
+# Dispatch to an internal helper function that handles the dimension argument in
+# the same manner as Base, with dims=: denoting a regular sum
+sum(x::AbstractExpr; dims=:) = _sum(x, dims)
```
matbesancon: Should we put the comment above as docstring for `sum(::AbstractExpr)`?

ararslan: I don't think so, since the fact that this dispatches to a helper function which isn't exported is an implementation detail.

matbesancon: That no indeed, but the
```julia
handles the dimension argument in
the same manner as Base, with dims=: denoting a regular sum
```
Could be a docstring

ararslan: Yeah true. I'd rather make documentation-related changes and other cleanup tasks in a separate PR though.

matbesancon: In that case the comment lines are added in the PR, so it wouldn't be out of scope to make it a docstring right away

test/test_socp.jl
```diff
@@ -226,18 +226,18 @@ TOL = 1e-3
     @test norm(g, 2) ^ 2 ≈ 0 atol=TOL
   end
 
-  @testset "norm consistent with Base" begin
+  @testset "matrix norm consistent with Base" begin
```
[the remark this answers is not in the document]
ararslan: Yeah, come to think of it this change is actually more confusing.

src/atoms/sdp_cone/operatornorm.jl
```diff
@@ -42,6 +43,21 @@ end
 operatornorm(x::AbstractExpr) = OperatorNormAtom(x)
```
jrevels: Looks like these are exported; should `operatornorm(x)` be deprecated for `opnorm(x)`?

ararslan: Good idea. I've implemented that in a separate commit.

PR #244, opened 2018-11-07
RFC/WIP: Load solver packages using Requires.jl

ararslan:
Currently the package implements its own version of optional dependencies which checks the current user's package installation directories for the given solver packages. This has issues both with precompilation and with Pkg in Julia 0.7 and higher.

Now, historically the Requires package has been a hack, but a recent overhaul makes it more legit to use. It is currently the best, if only, solution to optional dependencies.

This change moves all solver selection logic into Requires. This should be largely user-invisible, with the very notable exception that the first matching solver package is NOT automatically loaded; the user will now need to explicitly load a solver package in order to use Convex. The order of `using Convex` and `using SomeSolver` does not matter, however, and also note that this change requires no changes to the tests.


src/solver_info.jl
```diff
@@ -1,54 +1,21 @@
...
+  DEFAULT_SOLVER[] = solver
 end
 
 function get_default_solver()
-  if DEFAULT_SOLVER == nothing
+  if DEFAULT_SOLVER[] === nothing
     error("The default solver is set to `nothing`
-         You must have at least one solver installed to use Convex.
+         You must have at least one solver installed and loaded to use Convex.
          You can install a solver such as SCS by running:
```
oxinabox:
```suggestion
         You can load a solver such as SCS (once installed) by running 
         using SCS,
        before or after loading Convex.
```

It is probably not required to tell them how to install a package,
since if they try to load a package that is not installed, julia itself will tell them how to do a `Pkg.add`.
But it is perhaps worth explaining that loading is done with the `using` command (at least usually).

ararslan:
As @mlubin pointed out in Slack, another possible approach is the one taken by JuMP: do not provide a default nor load any solvers automatically and instead require that the user specify the solver they wish to use.

oxinabox:
That does seem quite clean.
And being like JuMP is good for setting user expectations of how things would work.

test/runtests.jl
```diff
@@ -14,15 +14,15 @@ push!(solvers, ECOSSolver(verbose=0))
...
-if isinstalled("Gurobi")
-    using Gurobi
-    push!(solvers, GurobiSolver(OutputFlag=0))
-end
-
-if isinstalled("Mosek")
-    using Mosek
-    push!(solvers, MosekSolver(LOG=0))
-end
+#if isinstalled("Gurobi")
```
mlubin: Why the commented block of code?

ararslan: We don't currently install Mosek and Gurobi for the tests, so these tests were only triggered if the user happened to have the packages installed. I commented them out because I didn't know whether we should remove the code or actually require and test those solvers along with the rest.

oxinabox: We can't test the Gurobi (and I assume Mosek) without a CI server with liscenses.

REQUIRE
```diff
@@ -1,3 +1,4 @@
 julia 0.7
 MathProgBase 0.7 0.8
 DataStructures
+Requires
```
rofinn: I'll note that Requires tends to significantly slow down package load times (I looking into it for Memento and FilePaths) and I'm not sure how well it'll play with system images.

oxinabox: Requires is supposed to speed up load times, that was it's original purpose.

iamed2: "supposed to"

rofinn: https://github.com/MikeInnes/Requires.jl/issues/39

PR #245, opened 2018-11-07
Remove vecdot

ararslan:
The `vecdot` function was deprecated in favor of `dot` in https://github.com/JuliaLang/julia/pull/27401; `dot` on matrices now treats them like vectors.

The change here updates our extension of `dot` to match this behavior and deprecates `vecnorm`.

### DIFF
--- a/src/atoms/affine/dot.jl
+++ b/src/atoms/affine/dot.jl
@@ -1,26 +1,23 @@
 import LinearAlgebra.dot
-export vecdot, dot
+export dot
 
+ismatrix(x::AbstractExpr) = (s = size(x); length(s) == 2 && s[1] > 1 && s[2] > 1)
+ismatrix(::AbstractMatrix) = true
+ismatrix(::Any) = false
 
-vecdot(x::AbstractExpr, y::AbstractExpr) = sum(broadcast(*, x, y))
-vecdot(x::Value, y::AbstractExpr) = sum(broadcast(*, Constant(x), y))
-vecdot(x::AbstractExpr, y::Value) = sum(broadcast(*, x, Constant(y)))
+# NOTE: Using asvec avoids broadcast-specific behaviors that we want to avoid, such
+# as extending singleton dimensions. We need to ensure that the inputs have the same
+# length, which broadcast will check for us if both inputs are vectors.
+asvec(x) = convert(AbstractExpr, ismatrix(x) ? vec(x) : x)
+_vecdot(x, y) = sum(broadcast(*, asvec(x), asvec(y)))
 
-dot(x::AbstractExpr, y::AbstractExpr) = (ismatrix(x) || ismatrix(y)) ? error("dot not implemented for matrices. perhaps you're looking for vecdot?") : vecdot(x, y)
-dot(x::Value, y::AbstractExpr) = (ismatrix(x) || ismatrix(y)) ? error("dot not implemented for matrices. perhaps you're looking for vecdot?") : vecdot(x, y)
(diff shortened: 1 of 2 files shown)


src/atoms/affine/dot.jl
```diff
@@ -1,26 +1,19 @@
 import LinearAlgebra.dot
-export vecdot, dot
+export dot
 
+ismatrix(x::AbstractExpr) = (s = size(x); length(s) == 2 && !any(==(1), s))
```
oxinabox: Perhaps,

```suggestion
function ismatrix(x::AbstractExpr)
    s = size(x)
    return length(s) == 2 && s[1] != 1 && s[2] != 1
```

The `(a; b)` form, I find hard to read.
And using `any` on something of length 2 seems both harder to read and slow.

src/atoms/affine/dot.jl
```diff
@@ -1,26 +1,19 @@
...
+ismatrix(x::AbstractMatrix) = true
```
oxinabox:
```suggestion
ismatrix(::AbstractMatrix) = true
```

src/atoms/affine/dot.jl
```diff
@@ -1,26 +1,19 @@
...
+ismatrix(x) = false
```
oxinabox:
```suggestion
ismatrix(::Any) = false
```

src/atoms/affine/dot.jl
```diff
@@ -1,26 +1,19 @@
...
 
-vecdot(x::AbstractExpr, y::AbstractExpr) = sum(broadcast(*, x, y))
-vecdot(x::Value, y::AbstractExpr) = sum(broadcast(*, Constant(x), y))
-vecdot(x::AbstractExpr, y::Value) = sum(broadcast(*, x, Constant(y)))
+asvec(x) = ismatrix(x) ? convert(AbstractExpr, vec(x)) : convert(AbstractExpr, x)
 
-dot(x::AbstractExpr, y::AbstractExpr) = (ismatrix(x) || ismatrix(y)) ? error("dot not implemented for matrices. perhaps you're looking for vecdot?") : vecdot(x, y)
-dot(x::Value, y::AbstractExpr) = (ismatrix(x) || ismatrix(y)) ? error("dot not implemented for matrices. perhaps you're looking for vecdot?") : vecdot(x, y)
-dot(x::AbstractExpr, y::Value) = (ismatrix(x) || ismatrix(y)) ? error("dot not implemented for matrices. perhaps you're looking for vecdot?") : vecdot(x, y)
+dot(x::AbstractExpr, y::AbstractExpr) = sum(broadcast(*, asvec(x), asvec(y)))
```
iamed2: Can we do this without `asvec`? That works on normal matrices. Or are you trying to ensure length is the same to avoid extending singleton dimensions?

ararslan: I don't really know what I was doing. I made this PR right before I went to bed last night. Most of this logic is completely unnecessary.

ararslan: Oh wait, yes, the lengths need to be the same and we need to avoid broadcast-specific behaviors

omus:
@ararslan what's left to do here?

ararslan:
I need to change this back to something like it was when I first pushed the branch to ensure that we don't end up using broadcast behaviors like extending singleton dimensions, as Eric noted.

iamed2:
What if the `Value` passed in is a 3+-dimensional array?

ararslan:
Those aren't supported by Convex, which is documented and assumed but not strictly enforced.

iamed2:
Uh. Okay then I guess.

PR #249, opened 2018-12-06
Deprecated default solvers

rofinn:
This seems like the first step to dropping the `DEFAULT_SOLVER` behaviour and getting rid of the following warning:

```
┌ Warning: Package Convex does not have ECOS in its dependencies:
│ - If you have Convex checked out for development and have
│   added ECOS as a dependency but haven't updated your primary
│   environment's manifest file, try `Pkg.resolve()`.
│ - Otherwise you may need to report an issue with Convex
```

which causes precompilation to fail on 1.0. Related to #244

### DIFF
--- a/src/problems.jl
+++ b/src/problems.jl
@@ -40,7 +40,7 @@ end
 
 # constructor if model is not specified
 function Problem(head::Symbol, objective::AbstractExpr, constraints::Array=Constraint[],
-                 solver::MathProgBase.AbstractMathProgSolver=get_default_solver())
+                 solver::MathProgBase.AbstractMathProgSolver=DEFAULT_SOLVER)
     Problem(head, objective, MathProgBase.ConicModel(solver), constraints)
 end
 
--- a/src/solver_info.jl
+++ b/src/solver_info.jl
@@ -4,16 +4,28 @@ export can_solve_mip, can_solve_socp, can_solve_sdp, can_solve_exp
 export set_default_solver, get_default_solver, isinstalled
 
 function set_default_solver(solver::MathProgBase.AbstractMathProgSolver)
-    Base.depwarn("Default solvers are deprecated and will not be provided in a future " *
-                 "release.\nLoad any necessary solvers manually.", :set_default_solver)
+    Base.depwarn(
+        "Default solvers are deprecated and will not be provided in a future release.
+        You can still manually load and use solvers such as ECOS by running:
+        using ECOS
+        p = minimize(...);
+        solve!(p, ECOSSolver())",
+        :set_default_solver
+    )
 
     global DEFAULT_SOLVER
     DEFAULT_SOLVER = solver
 end
 
 function get_default_solver()
-    Base.depwarn("Default solvers are deprecated and will not be provided in a future " *
-                 "release.\nLoad any necessary solvers manually.", :get_default_solver)
+    Base.depwarn(
+        "Default solvers are deprecated and will not be provided in a future release.
+        You can still manually load and use installed solvers such as ECOS by running:
+        using ECOS
+        p = minimize(...);
+        solve!(p, ECOSSolver())",
+        :get_default_solver
+    )
 
     if DEFAULT_SOLVER == nothing
         error("The default solver is set to `nothing`
--- a/src/solver_info.jl
+++ b/src/solver_info.jl
@@ -4,11 +4,17 @@ export can_solve_mip, can_solve_socp, can_solve_sdp, can_solve_exp
 export set_default_solver, get_default_solver, isinstalled
 
 function set_default_solver(solver::MathProgBase.AbstractMathProgSolver)
+    Base.depwarn("Default solvers are deprecated and will not be provided in a future " *
+                 "release.\nLoad any necessary solvers manually.", :set_default_solver)
+
     global DEFAULT_SOLVER
     DEFAULT_SOLVER = solver
 end
 
 function get_default_solver()
+    Base.depwarn("Default solvers are deprecated and will not be provided in a future " *
+                 "release.\nLoad any necessary solvers manually.", :get_default_solver)
+
     if DEFAULT_SOLVER == nothing
         error("The default solver is set to `nothing`
               You must have at least one solver installed to use Convex.
@@ -16,6 +22,7 @@ function get_default_solver()
               Pkg.add(\"SCS\").
               You will have to restart Julia after that.")
     end
+
     return DEFAULT_SOLVER
 end
 
@@ -29,6 +36,7 @@ function isinstalled(pkg)
             return true
         end
     end
+
     return false
 end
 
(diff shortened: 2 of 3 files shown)


src/solver_info.jl
```diff
@@ -9,13 +9,18 @@ function set_default_solver(solver::MathProgBase.AbstractMathProgSolver)
 end
 
 function get_default_solver()
+    Base.depwarn(
+        "Default solvers will be deprecated in a future release.", :get_default_solver
```
ararslan: Just call them deprecated now?
```julia
function get_default_solver()
    Base.depwarn("Default solvers are deprecated and will not be provided in a future " *
                 "release.\nLoad any necessary solvers manually.", :get_default_solver)
    ...
end
```

ararslan:
Ah, it seems to be because it's used in the default constructor for `Problem`. This might help:
```diff
--- a/src/problems.jl
+++ b/src/problems.jl
@@ -40,7 +40,7 @@ end
 
 # constructor if model is not specified
 function Problem(head::Symbol, objective::AbstractExpr, constraints::Array=Constraint[],
-                 solver::MathProgBase.AbstractMathProgSolver=get_default_solver())
+                 solver::MathProgBase.AbstractMathProgSolver=DEFAULT_SOLVER)
     Problem(head, objective, MathProgBase.ConicModel(solver), constraints)
 end
```

rofinn:
I guess if we deprecate the `set_default_solver` function too then that should be fine.

ararslan:
The warning still pops up a lot but it's at least less spammy than before it seems.

rofinn:
Yeah, looks like the test code is calling `get_default_solver()` pretty often.

ararslan:
That's probably not representative of external user/package code, so maybe fine? :sweat_smile:

src/problems.jl
```diff
@@ -40,7 +40,7 @@ end
 
 # constructor if model is not specified
 function Problem(head::Symbol, objective::AbstractExpr, constraints::Array=Constraint[],
-                 solver::MathProgBase.AbstractMathProgSolver=get_default_solver())
+                 solver::MathProgBase.AbstractMathProgSolver=DEFAULT_SOLVER)
```
oxinabox: We probably should introduce a function

```
check_default_solver(::Any)=nothing
check_default_solver(::Nothing)=`error("DEFAULT_SOLVER not set, you can ...."`
```

since right now the helpful error message is in `get_default_solver`,
which is no longer called.

src/solver_info.jl
```diff
@@ -4,18 +4,25 @@ export can_solve_mip, can_solve_socp, can_solve_sdp, can_solve_exp
...
+    Base.depwarn("Default solvers are deprecated and will not be provided in a future " *
+                 "release.\nLoad any necessary solvers manually.", :set_default_solver)
+
     global DEFAULT_SOLVER
     DEFAULT_SOLVER = solver
 end
 
 function get_default_solver()
+    Base.depwarn("Default solvers are deprecated and will not be provided in a future " *
+                 "release.\nLoad any necessary solvers manually.", :get_default_solver)
```
iamed2: I would give a helpful message that you can run, e.g., `using SCS` if you have it installed. I think it's important to be excessively helpful here, just as the error below is.

src/problems.jl
rofinn: That's true, but I think that's the desired behaviour to avoid spam. Currently, `get_default_solver` will be run on package load anyways.

https://github.com/JuliaOpt/Convex.jl/blob/04afdf5b23cfcd77c6f28f74457594df850f160e/src/solver_info.jl#L51

rofinn:
I'm not sure why the depwarn was causing problems, but I have noticed that it isn't very good at only warning once. Maybe it's related to where/how the deprecated function is called? :/

oxinabox:
It is because `maxlog` is broken, no?
https://github.com/JuliaLang/julia/issues/28786

It is now fixed, but that fix is only in master, I believe.

src/problems.jl
oxinabox: Shouldn't this be in `__init__`  as else it will not work with precompilation?

(the whole block https://github.com/JuliaOpt/Convex.jl/blob/04afdf5b23cfcd77c6f28f74457594df850f160e/src/solver_info.jl#L43-L60)

iamed2:
Looks like that made it in to 1.0.2

rofinn:
Hmmm, I'd rather not wait until that change is tagged in an official release, but if folks would prefer that then I can just work off my other branch for a while.

oxinabox:
Ah, so it is.

src/solver_info.jl
rofinn: How do folks feel about this warning?
```
Base.depwarn(
        "Default solvers are deprecated and will not be provided in a future release.
        You can still manually load and use installed solvers such as ECOS by running:
        using ECOS
        p = minimize(...);
        solve!(p, ECOSSolver())",
        :get_default_solver
 )
```

src/problems.jl
rofinn: That's true. If folks want we could just move that code into an `__init__`, which would be a less breaking change. At this point, I'd rather just deprecate this interface anyways.

rofinn: Actually, I guess we could do both. I could move the code into an `__init__` (probably a separate PR?) for an earlier release, but still deprecate the default solver functionality. This would help move the interface closer to JuMP, but allow us to provide a longer deprecation cycle.

src/solver_info.jl
```diff
@@ -4,18 +4,37 @@ export can_solve_mip, can_solve_socp, can_solve_sdp, can_solve_exp
 export set_default_solver, get_default_solver, isinstalled
 
 function set_default_solver(solver::MathProgBase.AbstractMathProgSolver)
+    Base.depwarn(
+        "Default solvers are deprecated and will not be provided in a future release.
+        You can still manually load and use solvers such as ECOS by running:
+        using ECOS
+        p = minimize(...);
+        solve!(p, ECOSSolver())",
```
ararslan: Perhaps use multi-line strings, since these are strings across multiple lines? e.g.
```julia
Base.depwarn("""
    Default solvers are deprecated and will not be provided in a future release.
    You can still manually load and use solvers such as ECOS by running:
        using ECOS
        p = minimize(...)
        solve!(p, ECOSSolver())
    """,
    :set_default_solver
)
```

rofinn: I was mostly trying to follow whatever style was being used below, but that does seem cleaner.

rofinn:
```
error("The default solver is set to `nothing`
          You must have at least one solver installed to use Convex.
          You can install a solver such as SCS by running:
          Pkg.add(\"SCS\")
          You will have to restart Julia after that.")
```

rofinn: Opted to make an issue to cleanup the formatting more generally. https://github.com/JuliaOpt/Convex.jl/issues/251

PR #250, opened 2018-12-07
WIP: Completely dropped the DEFAULT_SOLVER logic

rofinn:
- Dropped get_default_solver and set_default_solver
- Reorganized tests to not depend on get_default_solver (the giant @testset for loop can be changed).
- Only API change is that if you didn't define a problem with a solver/model then `solve!(p)` throws an argument
  (e.g., `solve!(p, solver)` is preferred now. NOTE: The existing `solve!(p, solver)` just assigns the problem model anyways
  with `problem.model = MathProgBase.ConicModel(s)`

NOTE: Depends on #249 getting merged and tagged first. I'll need to do a rebase after that.


src/solution.jl
```diff
@@ -18,6 +18,12 @@ function solve!(problem::Problem;
                 check_vexity=true,
                 verbose=true)
 
+    if problem.model === nothing
+        throw(ArgumentError(
+            "The provided problem hasn't been initialized with a conic model."
```
iamed2: Maybe suggest calling the other method of `solve!`?

iamed2:
I don't mind the for loop but maybe split the sections inside it into different files? They can still reference the `solver` variable if they're `include`d.

rofinn:
Weird because I kept getting `UndefVarError`s when I tried that.

src/problems.jl
```diff
@@ -39,9 +40,11 @@ mutable struct Problem
...
 # constructor if model is not specified
-function Problem(head::Symbol, objective::AbstractExpr, constraints::Array=Constraint[],
-                 solver::MathProgBase.AbstractMathProgSolver=get_default_solver())
-    Problem(head, objective, MathProgBase.ConicModel(solver), constraints)
+function Problem(head::Symbol, objective::AbstractExpr,
+                 constraints::Array=Constraint[],
+                 solver::Union{MathProgBase.AbstractMathProgSolver, Nothing}=nothing)
+    model = solver !== nothing ? MathProgBase.ConicModel(solver) : solver
+    Problem(head, objective, model, constraints)
 end
```
omus: Given the inner constructor I'm not following why we don't just define this as:
```julia
function Problem(head::Symbol, objective::AbstractExpr, constraints::Array,
                 solver::MathProgBase.AbstractMathProgSolver)
    Problem(head, objective, MathProgBase.ConicModel(solver), constraints)
end
```

PR #256, opened 2018-12-20
Deprecate get_vectorized_size in favor of length

ararslan:
They're defined to be the same thing but the implementation of the former is 2x slower than the latter. Rather than improving the code we can just combine them.

Further, we can simplify the implementation of `length` by calling `length` directly on the underlying value for `Constant`s.

### DIFF
--- a/src/expressions.jl
+++ b/src/expressions.jl
@@ -34,7 +34,6 @@ export vexity, sign, size, evaluate, monotonicity, curvature, length, convert
 export conic_form!
 export lastindex, ndims
 export Value, ValueOrNothing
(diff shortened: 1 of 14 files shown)


src/expressions.jl
```diff
@@ -95,7 +94,7 @@ function size(x::AbstractExpr)
 end
 
 function length(x::AbstractExpr)
-    return prod(x.size)
+    return x.size[1] * x.size[2]
```
iamed2: This change generates the exact same code. I would leave it as is.

ararslan: Okay

Issue #254, opened 2018-12-14
Massive RAM for moderate problem


cossio:
I have an example of a moderately sized problem where problem formulation is consuming massive memory. I have no idea why this happens. Is there something inherently difficult to this kind of problem? Or is this a bug in Convex.jl?

Note that this is not an issue with any solver. The RAM spike occurs at the problem formulation step, the line `problem = maximize(...)` below.

To run this code, you need to load the variables `N, Adj` from the `.jld2` file (using JLD2.jl) here:

https://github.com/cossio/cvx_example_data

`N` is an float 3-dimensional Array, and `Adj` a sparse matrix.

**WARNING:** The following code took so much memory that I had to reboot my laptop. Run this in a contained environment where you can constrain memory consumption. (For example, using https://github.com/pshved/timeout, start julia with `timeout -m 4000000 julia` to limit memory usage to 4GB.)

```julia
x = Variable(2730)
S, V, T = size(N)
lN = log.(N)
M = vec(sum(N[:,:,2:T]; dims=(2,3)))
H = Adj * x
problem = maximize(-sum(logsumexp(lN[:,v,t] - H) for t = 1:T-1, v = 1:V) - dot(M, H), [x ≥ -1e2, x ≤ 1e2])
solve!(problem, ECOSSolver())
```

**UPDATE:** I'm now getting the RAM spike at the solve phase, instead of the formulation phase. Both ECOS and SCS have the same problem.

cossio:
@ararslan I'm now seeing the memory spike in the solve phase instead of the formulation. I'll close this unless something similar comes up again.

ararslan:
Okay, thanks for reporting back. Did you see the spike only with a particular solver? If so, it would be good to know which.

cossio:
@ararslan I just tested with ECOS and SCS, both consume > 4GB RAM (at which point I cancelled to avoid having to reboot my laptop again)

ararslan:
If it's both of them, that suggest to me that the issue may still be Convex's fault. Have you by chance done any profiling using the Profile stdlib package?

cossio:
@ararslan Problem is my laptop dies before I can complete the profiling. Any suggestions?

ararslan:
Oh that makes sense. Hm, I don't know what to suggest offhand... I wonder if maybe the work could be put into a `Task` then after a certain amount of time, send an interrupt to the task. Not sure if that would work though.

cossio:
[...]

PR #258, opened 2018-12-21
WIP: Use FillArrays and LazyArrays to reduce memory footprint

ararslan:
Currently sparse vectors and matrices are used in many places, but this can cause memory usage to skyrocket in certain circumstances. See for example issue #254.

This change uses LazyArrays and FillArrays to express matrix structures which don't need to be mutated or immediately materialized, leading to significant memory savings.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,5 @@
 julia 0.7
 MathProgBase 0.7 0.8
 DataStructures
+LazyArrays
+FillArrays
--- a/src/Convex.jl
+++ b/src/Convex.jl
@@ -4,6 +4,8 @@ module Convex
 import DataStructures
 using LinearAlgebra
 using SparseArrays
+using FillArrays
+using LazyArrays
 
 global DEFAULT_SOLVER = nothing
 ### modeling framework
--- a/src/atoms/affine/conv.jl
+++ b/src/atoms/affine/conv.jl
@@ -11,7 +11,7 @@ function conv(x::Value, y::AbstractExpr)
     end
     m = length(x)
     n = y.size[1]
-    X = spzeros(m+n - 1, n)
+    X = zeros(m+n - 1, n)
     for i = 1:n
         X[i:m+i-1, i] = x
     end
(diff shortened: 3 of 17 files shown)


ararslan:
Note that this requires tags of both LazyArrays and FillArrays. Local testing was done using the master branches of both packages.

iamed2:
Any measurements?

ararslan:
Right now this is just "throw laziness at the wall and see what sticks"; I'm mostly making sure that this approach doesn't full-on break anything. Doing long-running computations and monitoring `top` shows drastically decreased memory usage on this branch, but I don't have individual benchmarks at the moment. I'll do that more formally as I fine-tune this.

ararslan:
Unfortunately, for the simplest problem:
```julia
x = Variable(1)
p = minimize(2.0 * x, [x >= 2, x <= 4])
```
this PR provides a pretty modest benefit in terms of memory use but a regression in terms of execution time.

Master:
```
julia> @benchmark Convex.conic_problem($p)
BenchmarkTools.Trial: 
  memory estimate:  38.77 KiB
  allocs estimate:  703
  --------------
  minimum time:     34.333 μs (0.00% GC)
  median time:      42.053 μs (0.00% GC)
  mean time:        138.870 μs (68.14% GC)
  maximum time:     504.959 ms (99.98% GC)
  --------------
  samples:          10000
  evals/sample:     1
```
PR:
```
julia> @benchmark Convex.conic_problem($p)
BenchmarkTools.Trial: 
  memory estimate:  30.17 KiB
  allocs estimate:  613
  --------------
  minimum time:     66.901 μs (0.00% GC)
  median time:      76.468 μs (0.00% GC)
  mean time:        107.625 μs (25.10% GC)
  maximum time:     24.867 ms (99.13% GC)
  --------------
  samples:          10000
  evals/sample:     1
```
Pretty disappointing. I'll keep at this.

ararslan:
Observation: LazyArrays has some facilities to act as a computation graph in some scenarios, e.g. with `Mul` representing a multiplication. There is a direct analog here with Convex's own `MultiplyAtom`. I wonder if there's a way to more fundamentally combine these two.

odow:
Closing as stale. 

Moving more to MOI (e.g., https://github.com/jump-dev/Convex.jl/pull/393) seems like a better way of dealing with the memory issues.

PR #260, opened 2018-12-22
Make ConicObj its own type

ararslan:
* Make ConicObj its own type. This reduces the amount of type piracy in the package. Currently the package defines multiplication of a sparse matrix and an ordered dictionary, as one example.

* Replace the dependency on DataStructures, which is quite large, with one on OrderedCollections. This is a lighter weight package and we're only using one piece of it anyway.

* Change some `Array` definitions to `Vector` as appropriate (the compiler likes it better) and add `@inbounds` annotations to a few loops that fill the arrays.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,3 @@
 julia 0.7
 MathProgBase 0.7 0.8
-DataStructures
+OrderedCollections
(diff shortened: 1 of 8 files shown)


ararslan:
Hm, I'm not quite sure what's going on with coverage here.

ararslan:
I'm going to call the coverage change a fluke. I'll eventually work on getting this package to 100% coverage (or close to it).

PR #264, opened 2019-01-08
Bug in sign detection of complex constants

mikolajpp:
The following snippet
```
z = ComplexVariable(2,2)
[1.0 0.0*im; 0.0 1.0]*z
```

will fail with 
```
MethodError: no method matching isless(::Int64, ::Complex{Float64})
Closest candidates are:
  isless(!Matched::Missing, ::Any) at missing.jl:66
  isless(::Real, !Matched::AbstractFloat) at operators.jl:149
  isless(::Real, !Matched::Real) at operators.jl:338
  ...
```

Quick fix is to force the type to actually be real, but I have no idea if this will or will not 
impact any code down the line.

### DIFF
--- a/src/constant.jl
+++ b/src/constant.jl
@@ -22,9 +22,9 @@ struct Constant <: AbstractExpr
         if check_sign
             if !isreal(x)
                 return Constant(x, ComplexSign())
-            elseif all(xi >= 0 for xi in x)
+            elseif all(real(xi) >= 0 for xi in x)
                 return Constant(x, Positive())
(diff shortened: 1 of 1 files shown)


ararslan:
Sign-checked complex constants should always have `ComplexSign()`, so this looks like a bug in the `isreal` check.

ararslan:
I've created #265, which I think will be a more complete solution to this issue, so I'll close this. Thank you for the contribution and willingness to dive in to fix the issue!

PR #265, opened 2019-01-08
Parameterize Constant on the type of its value

ararslan:
This allows for more expressive condition checking using dispatch, as complex numbers, real numbers, or arrays thereof can be handled as separate methods.

~~This also limits the types that can be accepted as constant values to scalars, vectors, and matrices. Previously this was assumed but not actually checked.~~

Fixes #264.

### DIFF
--- a/src/constant.jl
+++ b/src/constant.jl
@@ -5,54 +5,56 @@
 export Constant
 export vexity, evaluate, sign, conic_form!
 
-struct Constant <: AbstractExpr
+ispos(x::Real) = x >= 0
+ispos(v::AbstractVecOrMat{<:Real}) = all(ispos, v)
+isneg(x::Real) = x <= 0
+isneg(v::AbstractVecOrMat{<:Real}) = all(isneg, v)
+
+_size(x::Number) = (1, 1)
+_size(x::AbstractVector) = (length(x), 1)
+_size(x::AbstractMatrix) = size(x)
+
+_sign(x::Union{Complex,AbstractVecOrMat{<:Complex}}) = ComplexSign()
+function _sign(x::Value)
+    if ispos(x)
+        Positive()
+    elseif isneg(x)
+        Negative()
+    else
+        NoSign()
+    end
+end
+
+struct Constant{T<:Value} <: AbstractExpr
     head::Symbol
     id_hash::UInt64
-    value::Value
+    value::T
     size::Tuple{Int, Int}
-    vexity::Vexity
     sign::Sign
 
-    function Constant(x::Value, sign::Sign)
-        sz = (size(x, 1), size(x, 2))
-        return new(:constant, objectid(x), x, sz, ConstVexity(), sign)
-    end
-
-    function Constant(x::Value, check_sign::Bool=true)
-        if check_sign
-            if !isreal(x)
-                return Constant(x, ComplexSign())
-            elseif all(xi >= 0 for xi in x)
-                return Constant(x, Positive())
-            elseif all(xi <= 0 for xi in x)
-                return Constant(x, Negative())
-            end
-        end
-        return Constant(x, NoSign())
-    end
+    Constant(x::Value, sign::Sign) = new{typeof(x)}(:constant, objectid(x), x, _size(x), sign)
+    Constant(x::Value, check_sign::Bool=true) = Constant(x, check_sign ? _sign(x) : NoSign())
 end
+
 #### Constant Definition end     #####
 
-function vexity(x::Constant)
-    return x.vexity
-end
+vexity(::Constant) = ConstVexity()
 
-function evaluate(x::Constant)
-    return x.value
-end
+evaluate(x::Constant) = x.value
 
-function sign(x::Constant)
-    return x.sign
-end
+sign(x::Constant) = x.sign
 
+# `real(::Real)` is a no-op and should be optimized out for `Constant{<:Real}`
+real_conic_form(x::Constant{<:Number}) = [real(x.value)]
+real_conic_form(x::Constant{<:AbstractVecOrMat}) = vec(real(x.value))
 
-function real_conic_form(x::Constant)
-    return vec([real(x.value);])
-end
-
-function imag_conic_form(x::Constant)
-    return im*vec([imag(x.value);])
-end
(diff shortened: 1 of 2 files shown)


src/constant.jl
```diff
@@ -5,54 +5,46 @@
...
-            elseif all(xi <= 0 for xi in x)
-                return Constant(x, Negative())
-            end
-        end
-        return Constant(x, NoSign())
-    end
-end
-#### Constant Definition end     #####
+    Constant(x::T, sign::Union{Positive,Negative,NoSign}) where {T<:Value} =
+        new{T}(:constant, objectid(x), x, _size(x), sign)
```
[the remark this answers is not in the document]
omus: I didn't noticed that at first. My only complaint now is:

> single-line function syntax being used over multiple lines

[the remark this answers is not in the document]
ararslan: Those are great though :D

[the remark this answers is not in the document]
ararslan: Updated.

src/constant.jl
```diff
@@ -5,54 +5,47 @@
...
-            if !isreal(x)
-                return Constant(x, ComplexSign())
-            elseif all(xi >= 0 for xi in x)
-                return Constant(x, Positive())
-            elseif all(xi <= 0 for xi in x)
-                return Constant(x, Negative())
-            end
-        end
-        return Constant(x, NoSign())
+    function Constant(x::T, sign::Union{Positive,Negative,NoSign}) where T<:Value
```
ararslan: While likely uncommon, I just realized that we could(/should?) support something like
```julia
Constant(x::Real, sign::ComplexSign) = Constant(Complex(x), sign)
```

ararslan: So currently Convex is fine with whatever sign you tell it that a `Constant` has and it stores the value with no conversion or checking against the given sign. This PR removes that entirely; mismatched value types and signs result in `MethodError`s. I'm not sure whether that's a bug or feature.

src/constant.jl
```diff
@@ -5,54 +5,47 @@
...
-struct Constant <: AbstractExpr
+const ComplexValue = Union{Complex,AbstractVecOrMat{<:Complex}}
+
+_size(x::Number) = (1, 1)
+_size(x::AbstractVector) = (length(x), 1)
+_size(x::AbstractMatrix) = size(x)
+
+_sign(x::ComplexValue) = ComplexSign()
+_sign(x::Number) = x >= 0 ? Positive() : Negative()
+_sign(v::AbstractVecOrMat) = all(x->x >= 0, v) ? Positive() : Negative()
```
ararslan: I just realized this isn't the same behavior as before. On master, when the sign is detected, it's only `Negative()` if _all_ values are less than or equal to 0. Currently in this PR, I'm calling it negative if any are not greater than or equal to 0, which is wrong.

test/test_utilities.jl
```diff
@@ -42,6 +42,36 @@
...
+        @test evaluate(x) === 1+0im
+        @test sign(x) == ComplexSign()
+
+        y = Constant(0+0im, Positive())
+        @test y isa Constant{Int}
+        @test evaluate(y) === 0
+        @test sign(y) == Positive()
+
+        @test_throws ArgumentError Constant(1+2im, Positive())
+        @test_throws ArgumentError Constant([-1,0], Positive())
```
ararslan: Oh, the conditions I wrote were specific to complex-valued inputs, so this doesn't error. Whoops.

src/constant.jl
```diff
@@ -5,54 +5,49 @@
...
+ispos(v::AbstractVecOrMat{<:Real}) = all(ispos, v)
+isneg(x::Real) = x <= 0
+isneg(v::AbstractVecOrMat{<:Real}) = all(isneg, v)
+
+_size(x::Number) = (1, 1)
+_size(x::AbstractVector) = (length(x), 1)
+_size(x::AbstractMatrix) = size(x)
+
+_sign(x::Union{Complex,AbstractVecOrMat{<:Complex}}) = ComplexSign()
+_sign(x::Value) = ispos(x) ? Positive() :
```
rofinn: I think this would be clearer as:
```
function _sign(x::Value)
    if ispos(x)
        Positive()
    elseif isneg(x)
        Negative()
    else
        NoSign()
    end
end
```

src/constant.jl
```diff
@@ -5,54 +5,49 @@
...
+real_conic_form(x::Constant{<:Number}) = [real(x.value)]
+real_conic_form(x::Constant{<:AbstractVecOrMat}) = vec(real(x.value))
 
-function imag_conic_form(x::Constant)
-    return im*vec([imag(x.value);])
-end
+# `imag(::Real)` always returns 0, so we can avoid the implicit conversion to `Complex`
+# by multiplication with `im` and just use an explicit call to `zeros` with the appropriate
+# length
+imag_conic_form(x::Constant{<:Real}) = zeros(Complex{typeof(x.value)}, 1)
```
rofinn: I think this an below would be clearer as:

```
imag_conic_form(x::Constant{T}) where T<:Real = zeros(Complex{T}, 1)
imag_conic_form(x::Constant{<:AbstractVecOrMat{T}}) where T<:Real = zeros(Complex{T}, length(x))
...
```

src/constant.jl
```diff
@@ -5,54 +5,49 @@
...
+imag_conic_form(x::Constant{<:AbstractVecOrMat{<:Real}}) = zeros(Complex{eltype(x.value)}, length(x))
```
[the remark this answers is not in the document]
oxinabox: Ah I see Rory already said this.
(This was from the review I forgot to click submit on yesterday)

test/test_utilities.jl
```diff
@@ -42,6 +42,35 @@
         @test size(y) == (2, 1)
     end
 
+    @testset "Parametric constants" begin
+        z = Constant([1.0 0.0im; 0.0 1.0])
+        @test z isa Constant{Matrix{Complex{Float64}}}
+
+        # Helper functions
```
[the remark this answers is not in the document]
ararslan: ¯\\\_(ツ)\_/¯ Doesn't seem necessary

PR #269, opened 2019-01-18
Add a Project file, remove vestigial Pkg uses

ararslan:
This adds a Project.toml file which records dependencies and compatibility information and it removes the vestigial uses of Pkg for hacking optional dependencies.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,0 +1,25 @@
+name = "Convex"
+uuid = "f65535da-76fb-5f13-bab9-19810c17039a"
+
+[deps]
+LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+MathProgBase = "fdba3010-5040-5b88-9595-932c9decdf73"
+OrderedCollections = "bac558e1-5e72-5ebc-8fee-abe8a469f55d"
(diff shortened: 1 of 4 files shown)


Project.toml
```diff
@@ -0,0 +1,25 @@
...
+[deps]
+LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+MathProgBase = "fdba3010-5040-5b88-9595-932c9decdf73"
+OrderedCollections = "bac558e1-5e72-5ebc-8fee-abe8a469f55d"
+SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
+
+[compat]
+MathProgBase = "0.7"
+OrderedCollections = ">=1.0.0"
+julia = ">=0.7"
```
iamed2: Should be `"0.7, 1.0"`

ararslan: It should work fine with 1.1 and nightly (1.2) as well though.

iamed2: `1.0` means `1.x`

iamed2: Read https://julialang.github.io/Pkg.jl/v1/compatibility/#Version-specifier-format-1

PR #271, opened 2019-01-23
Updated docs to reflect that one has to pass a solver to `solve!`.

ericphanson:
I tried to update the docs to reflect that `solve!` requires a solver be passed now. I used `SCSSolver()` in the code for concreteness.


ararslan:
Thanks!

PR #277, opened 2019-03-20
match neg(x) docs definition

### LINKED ISSUE #276: neg is backwards
Note that `neg`, [as documented](https://convexjl.readthedocs.io/en/latest/operations.html) (and according to its behavior in other versions of CVX) should be convex. In the docs, it says `neg(x) = max(-x, 0)`, and in reality, I believe it's doing `-max(-x, 0)`, and therefore returns ConcaveVexity.
```julia
julia> x = Variable(10);

julia> neg(x)
AbstractExpr with
head: min
size: (10, 1)
sign: Negative()
vexity: ConcaveVexity()
```

tomerarnon:
fixes #276.

### DIFF
--- a/src/atoms/lp_cone/min.jl
+++ b/src/atoms/lp_cone/min.jl
@@ -78,4 +78,4 @@ end
 min(x::AbstractExpr, y::AbstractExpr) = MinAtom(x, y)
 min(x::AbstractExpr, y::Value) = min(x, Constant(y))
 min(x::Value, y::AbstractExpr) = min(Constant(x), y)
-neg(x::AbstractExpr) = max(-x, 0)
+neg(x::AbstractExpr) = max(-x, Constant(0, Positive()))
--- a/src/atoms/lp_cone/min.jl
+++ b/src/atoms/lp_cone/min.jl
@@ -78,4 +78,4 @@ end
 min(x::AbstractExpr, y::AbstractExpr) = MinAtom(x, y)
 min(x::AbstractExpr, y::Value) = min(x, Constant(y))
 min(x::Value, y::AbstractExpr) = min(Constant(x), y)
-neg(x::AbstractExpr) = min(x, Constant(0, Negative()))
+neg(x::AbstractExpr) = max(-x, 0)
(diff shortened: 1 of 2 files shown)


src/atoms/lp_cone/min.jl
```diff
@@ -78,4 +78,4 @@ end
 min(x::AbstractExpr, y::AbstractExpr) = MinAtom(x, y)
 min(x::AbstractExpr, y::Value) = min(x, Constant(y))
 min(x::Value, y::AbstractExpr) = min(Constant(x), y)
-neg(x::AbstractExpr) = min(x, Constant(0, Negative()))
+neg(x::AbstractExpr) = max(-x, 0)
```
ararslan:
```suggestion
neg(x::AbstractExpr) = max(-x, Constant(0, Negative()))
```
This ensures that the result is `Negative()` even if the maximum is 0. IIRC, 0 is treated as `Positive()`, which is likely why Convex thinks it's not DCP-compliant.

tomerarnon: Comparing the two different cases at the command line yielded no difference. Despite that fact, for the sake of pedantry, I believe Negative() should now be Positive() given the change since `neg` returns positive values (-x negates x, and the max returns the new positive entries.)

It is also possible to define this function as `pos(-x)`, which also rightly returns `Positive()`.

tomerarnon:
Looks like the tests relied on the concave definition, so I reversed them. Now `neg(x) <= 4` is DCP and `neg(x) >= 4` is not.

src/atoms/lp_cone/min.jl
iamed2: I agree, I think it should be `Positive` now

tomerarnon: Is it important to enforce `Positive()`, or is the `0` literal positive by default?

ararslan: Doesn't hurt to be explicit, so I think `Constant(0, Positive())` would be reasonable here.

Issue #209, opened 2017-09-01
Dimension mismatch in kron


Rocknokokoro:
I am running a fresh install of Julia 0.52 and cloned the most recent version of Convex.

It appears that the object kron returns have different sizes depending on the order of the arguments.

Let me define two matrices:
X = [0 0;0 0];
Y = Variable(2,2);

If I investigate what kron does depending on the arguments I get the following:
kron(X,Y)
AbstractExpr with
head: transpose
size: (4, 4)
sign: Convex.NoSign()
vexity: Convex.AffineVexity()

kron(Y,X)
AbstractExpr with
head: transpose
size: (2, 2)
sign: Convex.NoSign()
vexity: Convex.AffineVexity()

I care about this because I would like to take the trace of the resulting matrix multiplied with a third matrix Z:

Z = [0 0 0 0; 0 0 0 0; 0 0 0 0; 0 0 0 0];

Then depending on the order of kron, the multiplication is well defined or I get an error:

Example OK:
trace(Z * kron(X, Y));

Example NOK:
trace(Z * kron(Y, X));
Cannot multiply two expressions of sizes (4,4) and (2,2)

in Convex.MultiplyAtom(::Convex.Constant, ::Convex.TransposeAtom) at /Users/localadmin/.julia/v0.5/Convex/src/atoms/affine/multiply_divide.jl:29
in *(::Array{Int64,2}, ::Convex.TransposeAtom) at /Users/localadmin/.julia/v0.5/Convex/src/atoms/affine/multiply_divide.jl:104
in include_string(::String, ::String) at ./loading.jl:441
in include_string(::String, ::String) at /Applications/Julia-0.5.app/Contents/Resources/julia/lib/julia/sys.dylib:?

Is this a bug or I am missing something about how to use kron?

Ayush-iitkgp:
@Rocknokokoro thanks for reporting this issue. Theoretically, both should return AbstractExpr of size (4,4). I am looking into [here](https://github.com/JuliaOpt/Convex.jl/blob/master/src/atoms/affine/kron.jl). I will update as soon as I come to some conclusion.

Rocknokokoro:
@Ayush-iitkgp playing with the code, the dimensions seem to be fine again if I change line 26 in kron.jl from

push!(row, a[i, j] .* b)

to

push!(row, a[i, j] * b)

but I have no idea what is the difference between .* and * so I might well be breaking something.

ararslan:
Fixed by #279.

PR #279, opened 2019-03-22
Fix `kron` dimension bug, add simple test

### LINKED ISSUE #267: Bug in kron
The following gives (2,2) dimensional variable
```
z = ComplexVariable(2,2)
kron(z, [1 0;  0 1])
```

Switching the order of the arguments gives the correct (4,4) variable result. 
Unless I am misunderstanding  the meaning of dimensions, the result should be the same in both cases. In either way, both expressions should be multiply-capable with 4x4 matrices.

ericphanson:
Fixes #267, #209. The dimensions of `kron(A,B)` should be the same as the dimensions of `kron(B,A)`. I just copied the working method to fix the broken one, and put the `isreal` check on the value again. That check is mentioned in <https://github.com/JuliaOpt/Convex.jl/issues/204#issuecomment-315590295>, seems to be working around the strange behavior `Convex.jl` has with complex-typed but `isreal`-true values.

### DIFF
--- a/src/atoms/affine/kron.jl
+++ b/src/atoms/affine/kron.jl
@@ -23,7 +23,11 @@ function kron(a::AbstractExpr, b::Value)
     for i in 1:size(a)[1]
         row = AbstractExpr[]
         for j in 1:size(a)[2]
-            push!(row, a[i, j] .* b)
+            if isreal(b)
+                push!(row, a[i, j] * real(b))
+            else
+                push!(row, a[i, j] * b)
+            end
         end
         push!(rows, foldl(hcat, row))
     end
--- a/test/test_affine.jl
+++ b/test/test_affine.jl
@@ -8,6 +8,13 @@
         @test evaluate(-x) ≈ 0 atol=TOL
     end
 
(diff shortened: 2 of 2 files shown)


src/atoms/affine/kron.jl
```diff
@@ -23,7 +23,11 @@ function kron(a::AbstractExpr, b::Value)
     for i in 1:size(a)[1]
         row = AbstractExpr[]
         for j in 1:size(a)[2]
-            push!(row, a[i, j] .* b)
+            if isreal(b)
+                push!(row, a[i, j] * real(b))
```
ararslan: You should just be able to use this definition only and ditch the conditional, since `real` is a no-op for real inputs.

ericphanson: If `b` is complex, we don't want to take the real part though. We just don't want the type to be complex if the entries of `b` are all real, otherwise it complains later. (Not my fix, just copying the workaround from the other method).

ararslan: Oh sorry, I read the logic backwards for some reason.

Issue #266, opened 2019-01-08
SDP constraint not satisfied 


osjacky430:
Hi, I'm trying to implement the simulation in the paper, and I've encountered the following convex problem

![image](https://user-images.githubusercontent.com/11375975/50851449-d67d7580-13b7-11e9-99db-2781486e9783.png)

which is a SDP constraint, where 

Q_i is a n x n, real, and semidefinite matrix
r_i is a n x 1 complex matrix
s_i, d_i are scalars, and d_i is non-negative

The code is 
```
using LinearAlgebra
using Distributions
using Random
using Convex
using SCS
using Mosek
using Printf

Nt = K = 3;
rng = MersenneTwister(1234);
chiSqDis = Chisq(2 * Nt);

S = ntuple(x->Variable(3, 3), K);

γ = 10^(1/10); # linear power ratio, 1dB
σ = 0.1;
C = Diagonal(fill(0.002, Nt)); # CSI error
ρ = 0.1;

ϵ = 1e-6;
feasible_num = 0;
realization = 2;

for iter = 1:realization
    global feasible_num;
    problem = minimize(tr(sum(S)));

    t = [];
    hBars = [];

    for i = 1:K
        hBar = randn(rng, ComplexF64, (Nt, 1));
        middle = S[i]/γ - sum(S[1:end .!= i]);

        Qi = sqrt(C) * middle * sqrt(C);
        ri = sqrt(C) * middle * hBar;
        si = real(hBar' * middle * hBar) - σ;
        di = sqrt(quantile(chiSqDis, 1 - ρ) / 2.0);
        ti = Variable();
        eye = Diagonal(fill(1, Nt));

        push!(t, ti);
        push!(hBars, hBar);

        problem.constraints += (S[i] == transpose(S[i]))
        problem.constraints += (S[i] ⪰ 0);
        problem.constraints += (ti ≥ 0);
        problem.constraints += ([Qi + ti * eye  ri; ri'   si - ti * di^2] ⪰ 0);
    end

    solve!(problem, MosekSolver());

    if problem.status == :Optimal
        feasible_num += 1;

        for j = 1:K
            global S;

            SumJ = S[j].value / γ - sum(Si.value for Si in S[1:end .!= j]);
            Qj = sqrt(C) * SumJ * sqrt(C);
            rj = sqrt(C) * SumJ * hBars[j];
            sj = real((hBars[j])' * SumJ * hBars[j])[1] - σ;
            tj = t[j].value;
[...]

PR #280, opened 2019-03-24
Fix right-multiply bug, add simple test

### LINKED ISSUE #274: Bug in norm/abs
Consider the following program
```julia
effs = Vector{ComplexF64}[[1/sqrt(2), 0], [1/sqrt(6), 1/sqrt(3)], [(-sqrt(2)-1im*sqrt(6))/(4sqrt(3)), 1/sqrt(3)], [(-sqrt(2)+1im*sqrt(6))/(4sqrt(3)), 1/sqrt(3)]]
effsperm = Vector{ComplexF64}[effs[2], effs[3], effs[4], effs[1]]

X = hcat(effs...)
Y = hcat(effsperm...)

d = size(X, 1)
ρ = ComplexVariable(d, d)
constraints = [ρ in :SDP]
constraints += [tr(ρ)==1]
target = real(norm(diag(X'*ρ*Y), 1))
problem = minimize(target, constraints)
solve!(problem, SCSSolver(verbose=1))
v = problem.optval
```
We get `v=0.660`. Now, lets calculate `sum(abs.(diag(X'*ρ.value*Y)))`. This evaluates to `0.590`. Shouldn't this be equal to `v`?

ericphanson:
Fixes #274 and #266.

I saw a suspicious adjoint and changed it to a transpose, and suddenly #274 worked (and the tests pass). 

I have to say, I don't entirely understand how this works. I think what's going on is that matrices are stored as long vectors, which I think corresponds to the map in (1.127) of [TQI], and then matrix multiplication can be represented on the vectorized space by the formula 1.132 of [TQI], which indeed uses the transpose, not the adjoint.

    [TQI]: https://cs.uwaterloo.ca/~watrous/TQI/TQI.1.pdf

But I'd appreciate a more appropriate reference!

### DIFF
--- a/src/atoms/affine/multiply_divide.jl
+++ b/src/atoms/affine/multiply_divide.jl
@@ -86,7 +86,7 @@ function conic_form!(x::MultiplyAtom, unique_conic_forms::UniqueConicForms=Uniqu
         # right matrix multiplication
         else
             objective = conic_form!(x.children[1], unique_conic_forms)
-            objective = kron(x.children[2].value', sparse(1.0I, x.size[1], x.size[1])) * objective
+            objective = kron(transpose(x.children[2].value), sparse(1.0I, x.size[1], x.size[1])) * objective
         end
         cache_conic_form!(unique_conic_forms, x, objective)
     end
--- a/test/test_affine.jl
+++ b/test/test_affine.jl
@@ -47,6 +47,12 @@
         @test vexity(p) == AffineVexity()
         solve!(p, solver)
         @test p.optval ≈ 3 atol=TOL
+
+        # Check #274
+        x = ComplexVariable(2,2)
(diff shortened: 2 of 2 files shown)


ericphanson:
It seems like #266 was caused by the same bug. In that issue, a matrix had negative eigenvalues, violating an SDP condition. That seems to not happen with this fix.

Output from https://github.com/JuliaOpt/Convex.jl/issues/266#issue-397038986  on master:
```
[-1.27951, 0.0300499, 0.0302101, 0.0303364]
[0.00346596, 0.00824788, 0.00830066, 0.0854156]
[-0.257762, 0.00577054, 0.00597361, 0.00630559]
feasble rate: 100.000000
```


versus from my branch:
```
[6.10046e-9, 0.0133301, 0.0139489, 0.0483067]
[4.74103e-10, 0.0137494, 0.0174826, 0.12606]
[-1.63983e-9, -1.00358e-9, 0.00651302, 0.0141659]
feasble rate: 100.000000
```
No more negative eigenvalues.

ericphanson:
Thanks!

PR #284, opened 2019-05-06
Fix and simplify partial trace, add more tests

### LINKED ISSUE #281: Bug in `partialtrace`
```julia
julia> using Convex

julia> A = rand(6,6);

julia> expr = partialtrace(Constant(A), 1, [2, 3])
AbstractExpr with
head: partialtrace
size: (3, 3)
sign: Positive()
vexity: ConstVexity()


julia> size(expr)
(3, 3)

julia> size(evaluate(expr))
(2, 2)
```
Here, of course, we should have `size(expr) == size(evaluate(expr))`.

I'm thinking we should try to do some automated testing, such as trying to randomly generate trivial problems (like https://github.com/JuliaOpt/Convex.jl/issues/274#issuecomment-475865220 where the only variable is constrained to be the identity matrix) and check that their `optval` equals the true solution. Such testing would have been able to catch the bugs from #279, #280, and this one.

ericphanson:
The partial trace was implemented using just `kron`, `*`, and `sum`, which are already atoms, so we actually don’t need `PartialTraceAtom`. Additionally, `evaluate` and `conic_form!` were implemented separately but via the same algorithm (and the version in `evaluate` was incorrect), so I’ve replaced it all with a generic function following the same method. I also fixed the test (to see why the new version of the test is right, note the solution to the SDP should satisfy the constraint), and added some additional tests.

Fixes #281.

### DIFF
--- a/src/atoms/affine/partialtrace.jl
+++ b/src/atoms/affine/partialtrace.jl
@@ -1,116 +1,43 @@
-import Base.sign
 export partialtrace
-export sign, curvature, monotonicity, evaluate
 
-struct PartialTraceAtom <: AbstractExpr
-    head::Symbol
-    id_hash::UInt64
-    children::Tuple{AbstractExpr}
-    size::Tuple{Int, Int}
-    sys::Int
-    dims::Vector
-
-    function PartialTraceAtom(x::AbstractExpr, sys::Int, dims::Vector)
-        if x.size[1] ≠ x.size[2]
-            error("Only square matrices are supported")
-        end
-        if ! (1 ≤ sys ≤ length(dims))
-            error("Invalid system, should between 1 and ", length(dims), " got ", sys)
-        end
-        if x.size[1] ≠ prod(dims)
-            error("Dimension of system doesn't correspond to dimension of subsystems")
-        end
-        children = (x, )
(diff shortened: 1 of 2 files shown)


src/atoms/affine/partialtrace.jl
```diff
@@ -1,116 +1,43 @@
...
-    sub_systems = [subsystem(i) for i in 1:length(dims)]
-    a = Matrix(1.0I, 1, 1)
-    for i in 1:length(dims)
-        if i == x.sys
-            continue
-        else
-            a = kron(a,sub_systems[i])
-        end
+    if ! (1 ≤ sys ≤ length(dims))
+        throw(ArgumentError("Invalid system index, should between 1 and ", length(dims), " got ", sys))
```
ararslan: There actually isn't a vararg constructor for `ArgumentError`s
```suggestion
        throw(ArgumentError("Invalid system index, should between 1 and $(length(dims)), got $sys"))
```

ararslan:
Looks good overall. Would you be willing to add some `@test_throws` for the argument checking here?

src/atoms/affine/partialtrace.jl
ericphanson: Oops! Thanks, I'll fix that and add some `@test_throws` tests for these

Issue #234, opened 2018-09-14
parameter change not detected


stumarcus576:
I am trying to solve two problems, one after the other, by varying a single parameter p. But, according to Convex.jl, the solution to the second problem is the same as the first, which I know should not be the case, so Convex.jl is not detecting the parameter change when it solves the problem the second time. That is, in the two problems below, Convex.jl reports the solution to be 4; but for the second problem the solution should be 5.

lam = Convex.Variable(9,1)
p = 4 # p is a parameter.
convex_problem = minimize( p*sum(lam)+p , [lam >= 0])
Convex.solve!(convex_problem,GurobiSolver()) 
println("p: $p. Convex Status: $(convex_problem.status). Objective Value of Lagrangian Relaxation: $(convex_problem.optval).")
p += 1 # Increment the parameter p.
Convex.solve!(convex_problem,GurobiSolver(),warmstart=true)
println("p: $p. Convex Status: $(convex_problem.status). Objective Value of Lagrangian Relaxation: $(convex_problem.optval).")

ericphanson:
The problem is that in the call `minimize( p*sum(lam)+p , [lam >= 0])`, when `p = 4` is just a value, internally Convex forgets that name `p` and simply uses the value `4` in its formulation of the problem. Instead, you can make a variable `p`, and then `fix!` it's value. In your case, that would be
```julia
using Convex, SCS
lam = Convex.Variable(9,1)
p = Variable()
fix!(p, 4)
convex_problem = minimize( p*sum(lam)+p , [lam >= 0])
Convex.solve!(convex_problem,SCSSolver(verbose=0))
println("p: $p. Convex Status: $(convex_problem.status). Objective Value of Lagrangian Relaxation: $(convex_problem.optval).")
fix!(p,5)
Convex.solve!(convex_problem,SCSSolver(verbose=0),warmstart=true)
println("p: $p. Convex Status: $(convex_problem.status). Objective Value of Lagrangian Relaxation: $(convex_problem.optval).")
```
(I used SCS instead of Gurobi, since I don't have Gurobi installed on this computer, but it should be the same).

I believe the problem is with the documentation. The example in <https://convexjl.readthedocs.io/en/stable/advanced.html#warmstarting> is exactly like your problem, and it incorrectly does not set `lambda` to be a parameter. In that case, one needs to use
```julia
lambda = Variable(Positive())
```
[...]

Issue #229, opened 2018-06-02
Warm Start example does not work as expected


rakeshvar:
I am running the example given in the [docs](http://convexjl.readthedocs.io/en/latest/advanced.html). 
As I change the value of lambda the doc seems to imply that we get different results; but we do not.  
What is the best way to solve this problem for different values of lambda? 
using `lambda=Variable(); fix!(lambda, 9)`? 

```julia
n = 9
y = 1:n
x = Variable(n)

lambda = 1
problem = minimize(sumsquares(y - x) + lambda * sumsquares(x - 10))
@time solve!(problem)
@show x.value  # Expected = 5.5:.5:9.5

lambda = 9
@time solve!(problem, warmstart=true)
@show x.value # Expected = 9.1:.1:9.9
```

ericphanson:
Yes, I believe you should use `fix!`, and I made a pull request to change the docs to this effect; changing `lambda` as the docs suggest after the problem is constructed does not change the problem at all (in fact, since `minimize` is not a macro, I think it would be impossible for that to happen).

Note to use `fix!` you would need `lambda = Variable(Positive())` because Convex does not update the sign automatically when using `fix!` (maybe this should be added?). You could also do
```julia
n = 9
y = 1:n
x = Variable(n)

lambda = 1
obj1 = sumsquares(y - x)
obj2 = sumsquares(x - 10)
problem = minimize( obj1 + lambda * obj2 )
@time solve!(problem, SCSSolver())
@show x.value  # Expected = 5.5:.5:9.5

lambda = 9
problem.objective =  obj1 + lambda * obj2 
@time solve!(problem, SCSSolver(),  warmstart=true)
@show x.value # Expected = 9.1:.1:9.9
```
which avoids `fix!`. Reusing `obj1` and `obj2`, Convex doesn't have to recompute the conic forms, which might be helpful for large problems.

PR #291, opened 2019-05-12
[docs] Fix `warmstart` example

ericphanson:
Changing `lambda` after the problem is constructed doesn't do anything. Resolves https://github.com/JuliaOpt/Convex.jl/issues/234 and resolves https://github.com/JuliaOpt/Convex.jl/issues/229.

Another issue is that this example doesn't seem to work very well; maybe it depends on the random choice of `y`, but when I try it, `SCS` runs out of iterations, and `Mosek` stalls but gets an optimal result. I tried a few other things but didn't find a nicer problem quickly. So it's hard to tell if the warmstart is actually speeding up the solve. Any ideas for a better problem?

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -26,7 +26,7 @@ Warmstarting
 If you're solving the same problem many times with different values
 of a parameter, Convex.jl can initialize many solvers with the solution
 to the previous problem, which sometimes speeds up the solution time.
-This is called a **warm start**.
+This is called a **warm start**. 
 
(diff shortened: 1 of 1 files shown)


raphaelchinchilla:
@ericphanson  the documentation has not yet been updated, at least not on the website

ericphanson:
Yeah, this would need to get merged first. It needs to be reformatted since the docs have changed format since this PR was made. I should do that and merge this so the docs get fixed

Issue #223, opened 2018-02-15
Array of Variable is Unexpected behavior


utotch:
Hello.
I found that I can't make "Array of Variable" by using Convex.jl.
I think it is unexpected behavior but is it a supposed specification?

I think a collection of Variable is useful in some situations and this behavior limits my coding style...

-- simple example
```
using Convex
a = Variable(2)
[a]
```

```
MethodError: no method matching +(::Convex.NoSign)
Closest candidates are:
  +(::Convex.NoSign, ::Convex.NoSign) at /opt/julia/v0.6/Convex/src/dcp.jl:93
  +(::Convex.NoSign, ::Convex.Positive) at /opt/julia/v0.6/Convex/src/dcp.jl:94
  +(::Convex.NoSign, ::Convex.Negative) at /opt/julia/v0.6/Convex/src/dcp.jl:96
  ...

Stacktrace:
 [1] sum(::Tuple{Convex.NoSign}) at ./tuple.jl:323
 [2] sign(::Convex.HcatAtom) at /opt/julia/v0.6/Convex/src/atoms/affine/stack.jl:26
 [3] show(::IOContext{Base.AbstractIOBuffer{Array{UInt8,1}}}, ::Convex.TransposeAtom) at /opt/julia/v0.6/Convex/src/utilities/show.jl:56
 [4] limitstringmime(::MIME{Symbol("text/plain")}, ::Convex.TransposeAtom) at /opt/julia/v0.6/IJulia/src/inline.jl:24
 [5] display_dict(::Convex.TransposeAtom) at /opt/julia/v0.6/IJulia/src/execute_request.jl:28
 [6] (::Compat.#inner#17{Array{Any,1},IJulia.#display_dict,Tuple{Convex.TransposeAtom}})() at /opt/julia/v0.6/Compat/src/Compat.jl:488
 [7] execute_request(::ZMQ.Socket, ::IJulia.Msg) at /opt/julia/v0.6/IJulia/src/execute_request.jl:186
 [8] (::Compat.#inner#17{Array{Any,1},IJulia.#execute_request,Tuple{ZMQ.Socket,IJulia.Msg}})() at /opt/julia/v0.6/Compat/src/Compat.jl:488
 [9] eventloop(::ZMQ.Socket) at /opt/julia/v0.6/IJulia/src/eventloop.jl:8
 [10] (::IJulia.##14#17)() at ./task.jl:335

```
-- and can't make Array of Variable
```
using Convex
[...]

Issue #282, opened 2019-03-28
Convex crashing silently


lpawela:
This is an extended restatement of the following issue https://github.com/ZKSI/QuantumInformation.jl/issues/50
I couldn't find any reason for the crashes on our side. Consider the following code
```julia
using Convex, SCS, LinearAlgebra

⊗ = kron
function norm_diamond_primal(J::Matrix, d1::Int, d2::Int, eps)
    
    X = ComplexVariable(d1*d2, d1*d2)
    t = 0.5*inner_product(X, J) + 0.5*inner_product(X', J')

    ρ₀ = ComplexVariable(d1, d1)
    ρ₁ = ComplexVariable(d1, d1)

    constraints = [ρ₀ in :SDP, ρ₁ in :SDP]
    constraints += tr(ρ₀) == 1
    constraints += tr(ρ₁) == 1
    iden = Matrix{ComplexF64}(I, d2, d2)
    constraints += [iden ⊗ ρ₀ X; X' iden ⊗ ρ₁] in :SDP

    problem = maximize(t, constraints)
    solve!(problem, SCSSolver(verbose=0, eps=eps))
    problem.optval
end

function ket(::Type{T}, val::Int, dim::Int) where T<:Number
    dim > 0 ? () : throw(ArgumentError("Vector dimension has to be nonnegative"))
    1 <= val <= dim ? () : throw(ArgumentError("Label have to be smaller than vector dimension"))
    ψ = zeros(T, dim)
    ψ[val] = one(T)
    ψ
end
ket(val::Int, dim::Int) = ket(ComplexF64, val, dim)
proj(ψ::AbstractVector{<:Number}) = ψ * ψ'


effs = Vector{ComplexF64}[[1/sqrt(2), 0], [1/sqrt(6), 1/sqrt(3)], [(-sqrt(2)-1im*sqrt(6))/(4sqrt(3)), 1/sqrt(3)], [(-sqrt(2)+1im*sqrt(6))/(4sqrt(3)), 1/sqrt(3)]]
effsperm = Vector{ComplexF64}[effs[2], effs[3], effs[4], effs[1]]
J1 = sum(proj(ket(i, 4)) ⊗ proj(ket(j, 4)) ⊗ proj(effs[i]) ⊗ proj(effs[j]) for i=1:4, j=1:4)
[...]

PR #293, opened 2019-05-13
Remove Base.vect override

ericphanson:
Fixes https://github.com/JuliaOpt/Convex.jl/issues/223.

I'm not sure if this change is OK or not. On the one hand, the tests pass. On the other hand, it must be here for a reason; maybe this change will break some obscure edge case. In the latter case though, it would be nice if we had a test to show we need this override (or even better, find a different resolution to such an edge case without overriding `Base.vect`).

I thought I'd make the pull request to open a discussion at least.

P.S. sorry for the barrage of pull requests; no rush to address these, I'm just trying to work through the issue backlog a bit.

### DIFF
--- a/src/atoms/affine/stack.jl
+++ b/src/atoms/affine/stack.jl
@@ -131,9 +131,3 @@ function hvcat(rows::Tuple{Vararg{Int}}, args::AbstractExprOrValue...)
     end
     return vcat(rs...)
 end
-
-Base.vect(args::AbstractExpr...) = transpose(HcatAtom(map(transpose, args)...))
-Base.vect(args::AbstractExprOrValue...) = transpose(HcatAtom(map(arg -> transpose(convert(AbstractExpr, arg)), args)...))
-
-# XXX: Reimplementation of the Base method
-Base.vect(args::Value...) = copyto!(Vector{Base.promote_typeof(args...)}(undef, length(args)), args)
--- a/test/test_utilities.jl
+++ b/test/test_utilities.jl
@@ -88,6 +88,11 @@
         @test Convex.imag_conic_form(Constant([1.0, 2.0])) == [0.0, 0.0]
     end
 
+    @testset "Base.vect" begin
+    # Issue #223: ensure we can make vectors of variables
+    @test size([Variable(2), Variable(3,4)]) == (2,)
+    end
+
     @testset "Iteration" begin
         x = Variable(2,3)
         s = sum([xi for xi in x])


ararslan:
That extension was included to ensure that the `[A, B]` syntax produces an `HcatAtom`. Is that still the case if these definitions are removed?

ericphanson:
No, I don't think so:
```julia
julia> typeof([Variable(2), Variable(2)])
Array{Variable,1}
```
To be fair,
```julia
julia> typeof([rand(2), rand(2)])
Array{Array{Float64,1},1}
```
isn't an hcat either, is it?

ericphanson:
On this branch, we have
```julia
julia> [Variable(2), Variable(2)]
2-element Array{Variable,1}:
 Variable of
size: (2, 1)
sign: NoSign()
vexity: AffineVexity()
 Variable of
size: (2, 1)
sign: NoSign()
vexity: AffineVexity()

julia> [Variable(2) Variable(2)]
AbstractExpr with
head: hcat
size: (2, 2)
sign: NoSign()
vexity: AffineVexity()


julia> [Variable(2); Variable(2)]
AbstractExpr with
head: transpose
size: (4, 1)
sign: NoSign()
vexity: AffineVexity()
```
This behavior seems to match how usual vectors behave (e.g. substituting `Variable` for `rand`). Likewise for matrice:
```julia
julia> [Variable(2,3) Variable(2,3)]
AbstractExpr with
head: hcat
size: (2, 6)
sign: NoSign()
vexity: AffineVexity()


julia> [Variable(2,3); Variable(2,3)]
AbstractExpr with
head: transpose
size: (4, 3)
sign: NoSign()
vexity: AffineVexity()


julia> [Variable(2,3), Variable(2,3)]
2-element Array{Variable,1}:
 Variable of
size: (2, 3)
sign: NoSign()
vexity: AffineVexity()
 Variable of
size: (2, 3)
sign: NoSign()
vexity: AffineVexity()
```

So we do have `hcat` and `vcat` working when you separate with a space or semicolon, as with usual vectors or matrices, and it creates a list when separating with a comma. To me, this seems like natural behavior. Why do we want the comma separation to result in hcat too?

ericphanson:
If my understanding is right, I think the proposed changed here is analogous to https://github.com/JuliaLang/julia/pull/8599 in Base, which seems to have introduced `Base.vect`, leading to https://github.com/JuliaOpt/Convex.jl/issues/145 and these lines in Convex in the first place. So if I'm understanding it right, the proposed change here is to match the behavior in Base instead of overriding it for `Variable` types. This is thus a breaking change to Convex, to match the associated one in Base. 

But I think I am missing some context, because  I'd think the resolution in https://github.com/JuliaOpt/Convex.jl/issues/145 would have been to match base behavior if that had been an acceptable option at that time. Perhaps now it is an acceptable change? Or maybe those those reasons that kept it from being a solution to https://github.com/JuliaOpt/Convex.jl/issues/145 are still present.

ericphanson:
It seems like this change also fixes https://github.com/JuliaOpt/Convex.jl/issues/282 (see https://github.com/JuliaOpt/Convex.jl/issues/282#issuecomment-493688722).

ararslan:
Tbh I don't really know what the right thing to do here is, but your explanation with regards to the change in Base is compelling. Let's go for it and see what happens.

ericphanson:
Okay, exciting!

Issue #298, opened 2019-06-06
fix! and free! do not work as expected


boborbt:
I have a sample project in which fix! and free! do not seem to work correctly. In particular, it seems that once fixed variables never change again even after calling free!.

You can get the sample project at: https://pastebin.com/DX1mp5NN

In that sample project I implement the same minimisation problem two times. In the first function I do use the fix! and free! functions to implement an alternating least square approach. 
In the second function I do the same, but avoid the fix! and free! functions and resort to creating new problems where I substitute constants in place of variables that should be fixed.
While I would have expected the two functions to behaves identically, this is not the case. The version without the fix! and free! functions works as expected. The one with fix! and free! does not.

Here you can find the log of the output of the program: https://pastebin.com/fWWmRCFP
(the first 2226 lines are logs of the first function, the following ones are logs of the second function).

Here is an excerpt of the log output for the first few iterations of the implementation using the fix and free functions (α was initialised with value [0.3; 0.7; 1.0]; β was initialised with value [-3.; 3]):

```
┌ Debug: optval (β fixed)
│   p.optval = 103.47459779886627
│   α.value =
│    3×1 Array{Float64,2}:
│     0.0885974060292705
│     ⋮
│   β.value =
│    2-element Array{Float64,1}:
│     -3.0
│      3.0
└ @ Main ~/Research/LIMPET/block-least-squares/issue_example.jl:38
┌ Debug: optval (α fixed)
│   p.optval = 103.47459535939427
│   α.value =
│    3×1 Array{Float64,2}:
│     0.0885974060292705
│     ⋮
│   β.value =
│    2-element Array{Float64,1}:
│     -3.0
│      3.0
└ @ Main ~/Research/LIMPET/block-least-squares/issue_example.jl:44
┌ Debug: optval (β fixed)
│   p.optval = 103.47459779886627
│   α.value =
│    3×1 Array{Float64,2}:
│     0.0885974060292705
│     ⋮
│   β.value =
│    2-element Array{Float64,1}:
│     -3.0
│      3.0
└ @ Main ~/Research/LIMPET/block-least-squares/issue_example.jl:38
```

As you can see the β never changes after initialisation (since it is fixed before having the chance to be optimised in the first iteration), while α changes only the first time. This despite the calls to the free! functions.

Thanks in advance for any help.

ericphanson:
It seems to work on the master branch (you can try this with `] add Convex#master`). Can you confirm?

[...]

PR #299, opened 2019-06-08
Allow calling `fix!` twice in a row

ericphanson:
Currently, if you `fix!` twice and then `free!`, you end up with a variable with `AffineVexity` but `:fixed`, because `free!` only removes the last `:fixed` and still changes the convexity. 

On release and on master with ECOS, the test I added segfaults (which seems strange), and on release and master with SCS, the third test simply fails (the solver returns `NaN`).

### DIFF
--- a/src/variable.jl
+++ b/src/variable.jl
@@ -88,7 +88,7 @@ end
 
 function conic_form!(x::Variable, unique_conic_forms::UniqueConicForms=UniqueConicForms())
     if !has_conic_form(unique_conic_forms, x)
-        if x.vexity == ConstVexity()
+        if vexity(x) == ConstVexity()
             # do exactly what we would for a constant
             objective = ConicObj()
             objective[objectid(:constant)] = (vec([real(x.value);]),vec([imag(x.value);]))
--- a/src/variable.jl
+++ b/src/variable.jl
@@ -127,8 +127,7 @@ end
 fix!(x::Variable, v::Number) = fix!(x, fill(v, (1, 1)))
 
 function free!(x::Variable)
-    # TODO this won't work if :fixed appears other than at the end of x.sets
-    x.sets[end] == :fixed && pop!(x.sets)
+    deleteat!(x.sets, findall(==(:fixed), x.sets))
     x.vexity = AffineVexity()
     x
 end
--- a/src/variable.jl
+++ b/src/variable.jl
@@ -88,7 +88,7 @@ end
 
 function conic_form!(x::Variable, unique_conic_forms::UniqueConicForms=UniqueConicForms())
     if !has_conic_form(unique_conic_forms, x)
-        if :fixed in x.sets
+        if x.vexity == ConstVexity()
             # do exactly what we would for a constant
             objective = ConicObj()
             objective[objectid(:constant)] = (vec([real(x.value);]),vec([imag(x.value);]))
@@ -115,7 +115,6 @@ end
 # fix variables to hold them at their current value, and free them afterwards
 function fix!(x::Variable)
     x.value === nothing && error("This variable has no value yet; cannot fix value to nothing!")
-    push!(x.sets, :fixed)
     x.vexity = ConstVexity()
     x
 end
@@ -127,7 +126,6 @@ end
 fix!(x::Variable, v::Number) = fix!(x, fill(v, (1, 1)))
 
 function free!(x::Variable)
-    deleteat!(x.sets, findall(==(:fixed), x.sets))
     x.vexity = AffineVexity()
     x
 end
--- a/test/test_const.jl
+++ b/test/test_const.jl
@@ -31,5 +31,24 @@
         @test prob.optval ≈ 0.0 atol = TOL
     end
 
+    @testset "Test double `fix!`" begin
+        x = Variable()
+        y = Variable()
+        fix!(x, 1.0)
+        prob = minimize(y*x, [y >= x, x >= 0.5])
+        solve!(prob, solver)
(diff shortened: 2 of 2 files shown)


src/variable.jl
```diff
@@ -127,8 +127,7 @@ end
 fix!(x::Variable, v::Number) = fix!(x, fill(v, (1, 1)))
 
 function free!(x::Variable)
-    # TODO this won't work if :fixed appears other than at the end of x.sets
-    x.sets[end] == :fixed && pop!(x.sets)
+    deleteat!(x.sets, findall(==(:fixed), x.sets))
```
ararslan: I'm not super familiar with the `fix!`/`free!` stuff, but don't we only want to delete the last occurrence of `:fixed`?

ericphanson: My understanding was `:fixed` should be basically a boolean toggle, either the variable is `:fixed` or it isn't: if it has `:fixed` in its `sets` field, then it should also have `ConstVexity()` and should be effectively treated like a constant. So if you `fix!(x,3.0)` and then `fix!(x, 2.0)`, you are just changing the fixed value of `x`, but if you `free!(x)`, you're saying it's just a regular variable again (and so it shouldn't have `:fixed` in it's `.sets` field anymore). 

In fact, I think we don't need the concept of `:fixed` at all: for a `Variable` `x`, `:fixed` should be in `x.sets` if and only if `x.vexity == ConstVexity()` (and otherwise `x.vexity == AffineVexity()`). So we actually have two redundant ways to express `x` being fixed to a constant, and the bug that causes the segfault or NaN result with that test I added is what happens when these two concepts get out of sync: by calling `fix!` twice and then `free!`, the vexity gets reset to `AffineVexity()`, but it still has `:fixed` in its `.sets` field. Besides in `fix!` and `free!`, `:fixed` only shows up once in Convex.jl's source code, which is in `conic_form!(x::Variable,...)`, which has the lines
```julia
 if :fixed in x.sets
            # do exactly what we would for a constant
            ...
```
This causes the  twice `fix!`'d then `free!`'d `x` to hit this code path despite having `AffineVexity()`. So in part of the problem formulation `x` is treated like a variable, but here it is treated like a constant. And I think that's what causes the bug that this PR's test checks.

I'll push another commit that actually removes `:fixed` altogether, so we just have 1 toggle to mean constant or not: `ConstVexity()` or `AffineVexity()`, which would hopefully remove the possibility for this kind of bug.

ericphanson: So all the tests pass locally with this change (removing `:fixed` altogether). Maybe there is a usage of `fix!` and `free!` that is not covered by the tests (which are quite sparse regarding `fix!` and `free!`) which I don't understand where you don't want to remove all instances of `:fixed` though. I certainly don't want to be pushing for changes that will break people's code. To me, fixed or not as a boolean makes sense (either the variable is constant or it isn't), but maybe there is some more complicated usage. 

However, I somewhat doubtful that there are lots of active users of `fix!` and `free!` currently, since until https://github.com/JuliaOpt/Convex.jl/pull/294 (which hasn't made it into a release yet), the functionality was completely broken on any code hitting that path (e.g. https://github.com/JuliaOpt/Convex.jl/issues/298). So me, it makes sense to take the opportunity to have a simple semantic (treated like a constant or treated like a variable; double `fix!`ing just updates the value of the constant, `free!`ing makes it a regular variable again).

ararslan: Thanks for the very detailed and thoughtful analysis here. I think you're right, doing everything in terms of the vexity instead of a `Symbol`-based toggle makes more sense. I feel a little uncomfortable accepting the change unilaterally only because I have so little knowledge of the `fix!`/`free!` stuff, but then again, not many people really look at this package nowadays...

src/variable.jl
```diff
@@ -127,8 +126,6 @@ end
 fix!(x::Variable, v::Number) = fix!(x, fill(v, (1, 1)))
 
 function free!(x::Variable)
-    # TODO this won't work if :fixed appears other than at the end of x.sets
-    x.sets[end] == :fixed && pop!(x.sets)
     x.vexity = AffineVexity()
```
ararslan: This is mostly unrelated to the PR, but I noticed something: isn't it possible that `x` could have been something other than affine before fixing it as a constant? If that can occur, in such cases `x` would not actually have its vexity reset by freeing but rather changed entirely.

ericphanson: I see what you mean, but I think `Variable`'s only have `ConstVexity()` or `AffineVexity()`, where `ConstVexity()` corresponds to the fixed case. I think other vexities don't quite make sense for a `Variable` (but instead an expression involving a variable might have a different vexity, e.g. `log(x)` would be concave). To double check this, I searched the source for `.vexity`, since that would be the way to set a `Variable` to another vexity (since also searching for `field` and `property` doesn't show anything relevant), and only saw `.vexity` show up in a getter, in `free!`, in a show method, and in the line I changed,
```julia
        if x.vexity == ConstVexity()
            # do exactly what we would for a constant
            ...
```
so I think indeed, Variable's vexities are not set to anything but `AffineVexity()` (and `ConstVexity()` in `fix!`).

src/variable.jl
ericphanson: Thanks! Yeah, makes sense.

src/variable.jl
ericphanson: Just to say, I realized that actually I probably should have written `vexity(x)` instead of `x.vexity` in that if condition (we only need direct access for setting it, not accessing), so I pushed one (hopefully last) commit to change that.

ararslan:
Clearly no one else is going to look at this, so yolo

PR #302, opened 2019-07-26
Fixing bug for complex semidefinite cases

raphaelsaavedra:
Addressing #198. I believe this check is incorrectly checking for `ComplexVariable()` rather than `ComplexSign()` which results in some complex values in the constraints list.

Besides this small fix, I have added the bugged case from #198 in the tests. Please let me know if you want me to change or add something.

### DIFF
--- a/test/test_sdp.jl
+++ b/test/test_sdp.jl
@@ -250,6 +250,7 @@
                 solve!(p, SCSSolver())
                 @test p.status == :Optimal
                 @test p.solution.primal ≈ [0.; 1.; 0.; 0.; 1.; zeros(4)] atol=TOL
+                @test p.optval ≈ 0 atol=TOL
             end
 
             @testset "norm2 atom" begin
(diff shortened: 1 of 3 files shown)


test/test_sdp.jl
```diff
@@ -243,6 +243,15 @@
                 #@fact x1==x2 --> true
             end
 
+            @testset "" begin
```
ericphanson: Maybe `@testset "Issue #198" begin`?

ericphanson:
Good find! Looks good to me, besides maybe naming the test to point to the issue for future clarity. (I don't have merge rights though)

test/test_sdp.jl
raphaelsaavedra: Whoops I forgot to name the test! I'm about to take a train but I'll fix that later today.

raphaelsaavedra: Added it in 9392ca909cf2464513802484829fb29d9b037aeb.

ericphanson:
I have been given merge rights and I'll go ahead and merge this. Thanks a lot @raphaelsaavedra! That was a sneaky bug, and that issue had annoyed me for some time.

PR #304, opened 2019-08-09
Convert docs to Documenter.jl

ericphanson:
# Why?

* Consistency with base Julia and other packages to make it easier for contributors (like me) to add to the docs (reStructuredText's code and math formatting is pretty confusing if you aren't used to it).
* Remove a python build dependency, which is convenient (I had to make a virtual environment, install Sphinx and install the readthedocs theme before I could build locally, which took a fair bit of trial and error).

I hope to make improvements to Convex.jl, and being able to easily update the documentation alongside would be great :).

# Will it work?

It works locally when I `include("make.jl")` in the `docs` environment. I copied some bits from `PkgTemplates` so that Travis will try to build the docs. But Travis needs to have push access via an environmental variable. Maybe JuliaOpt already set that stuff up?

# How?

Pandoc mostly. The big tables in `operations.md` were annoying because Documenter seems to like what Pandoc calls "pipe tables", which wrap lines automatically but text cannot be entered into them on multiple lines, and pandoc will not convert multiline tables to pipe tables. In the end, what worked was to build the html files (via Sphinx), then in `_build/html/operations.html`, remove all `<p>` tags and `<colgroup>`'s, and do a

```
pandoc -t markdown+pipe_tables-grid_tables-simple_tables-multiline_tables _build/html/operations.html -o _build/html/operations.md
``` 
Then copy out those the resulting pipe tables from `docs/_build/html/operations.md` into the `docs/src/operations.md` file generated by directly converting the `.rst` via pandoc. Then regex-fix some css classes and extra `\`'s, run `Format Document` in vscode and viola.


ararslan:
This is amazing, thanks for doing it! I've added the requisite deployment keys to the repository and to Travis.

docs/make.jl
```diff
@@ -0,0 +1,28 @@
+using Documenter, Convex
+
+makedocs(;
+            modules = [Convex],
```
ararslan: These could all probably just be reduced to 4 spaces of absolute indentation rather than relative to a particular thing

docs/make.jl
```diff
@@ -0,0 +1,28 @@
...
+                "Optimizing in a Loop" => "loop.md",
+                "Advanced" => "advanced.md",
+                "Contributing" => "contributing.md",
+                "Credits" => "credits.md"
+            ],
+            repo="https://github.com/JuliaOpt/Convex.jl/blob/{commit}{path}#L{line}",
+            sitename="Convex.jl")
+
+
+deploydocs(;
```
ararslan:
```suggestion
deploydocs(repo="https://github.com/JuliaOpt/Convex.jl")
```
It's short enough, might as well have it on one line

docs/make.jl
ericphanson: Ok, let me know if the new commit addresses this or if I misunderstood

docs/make.jl
```diff
@@ -0,0 +1,25 @@
...
+        "Examples" => "examples.md",
+        "Complex-domain Optimization" => "complex-domain_optimization.md",
+        "Solvers" => "solvers.md",
+        "FAQ" => "faq.md",
+        "Optimizing in a Loop" => "loop.md",
+        "Advanced" => "advanced.md",
+        "Contributing" => "contributing.md",
+        "Credits" => "credits.md"
+    ],
+    repo = "https://github.com/JuliaOpt/Convex.jl/blob/{commit}{path}#L{line}",
```
ararslan: I haven't seen this before. What is it for? I think the line number part is confusing me.

docs/make.jl
ararslan: Yep, thanks!

ararslan:
I want to merge it and see it deployed, then we can fix anything that needs to be fixed if applicable.

docs/make.jl
ericphanson: I think it’s for the “edit this on github” links on the generated docs

PR #306, opened 2019-08-10
Fix repo specification for doc deployment

### LINKED ISSUE #305: Documentation deployment fails
```
fatal: protocol 'git@https' is not supported
┌ Error: Git failed to fetch git@https://github.com/JuliaOpt/Convex.jl
│ This can be caused by a DOCUMENTER_KEY variable that is not correctly set up.
│ Make sure that the environment variable is properly set up as a Base64-encoded string
│ of the SSH private key. You may need to re-generate the keys with DocumenterTools.
└ @ Documenter ~/.julia/packages/Documenter/zbb48/src/Documenter.jl:692
```
https://travis-ci.org/JuliaOpt/Convex.jl/jobs/570104348

ararslan:
Should fix #305, cc @ericphanson

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -22,4 +22,4 @@ makedocs(;
     sitename = "Convex.jl")
 
 
-deploydocs(repo = "https://github.com/JuliaOpt/Convex.jl")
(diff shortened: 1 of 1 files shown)


ericphanson:
Thanks for the fix!

PR #307, opened 2019-08-10
Add docs badge

ericphanson:
Can be convenient. Also, I think the "stable" links won't work until we tag a new release.


ararslan:
The docs are actually linked a little further down, but the badges are nice.

PR #314, opened 2019-08-16
Revert "Make global variables constant"

ericphanson:
Reverts JuliaOpt/Convex.jl#286

`empty!` doesn't actually free all the memory used by the dictionary, causing problems reported in https://github.com/JuliaOpt/Convex.jl/issues/83.

### DIFF
--- a/src/Convex.jl
+++ b/src/Convex.jl
@@ -84,10 +84,9 @@ include("utilities/broadcast.jl")
 
 #Temporary workaround for memory leak (https://github.com/JuliaOpt/Convex.jl/issues/83)
 function clearmemory()
-    global id_to_variables
-    empty!(id_to_variables)
-    global conic_constr_to_constr 
-    empty!(conic_constr_to_constr)
+    global id_to_variables = Dict{UInt64, Variable}()
+    global var_to_ranges = Dict{UInt64, Tuple{Int, Int}}()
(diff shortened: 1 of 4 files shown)


ericphanson:
I'm not sure this is actually a problem in practice. With the script
```julia
using Convex
using SCS

function testmem()
       x = Variable(10)
       A = rand(100, 10)
       z = randn(1000, 1000, 100)
       for i=1:1000, j = 1:1000
           b = z[i, j, :]
           problem = minimize(norm(A*x - b, 1))
           if (i*1000 + j) % 1000 == 0
                @info "Before clearing" i j Base.summarysize(Convex.id_to_variables)/1e6 Base.summarysize(Convex.conic_constr_to_constr)/1e6
               Convex.clearmemory()
               GC.gc()
               x = Variable(10)
               @info "After clearing" i j Base.summarysize(Convex.id_to_variables)/1e6 Base.summarysize(Convex.conic_constr_to_constr)/1e6
           end
       end
end

testmem()
```
we see that clearing is quite effective even if it doesn't completely empty the dictionaries. With ECOS there still seems to be a memory leak, but since it's solver specific, I suspect that's a bug in ECOS.

ericphanson:
I'll close this; if someone objects or #83 resurfaces and this seems to be the problem, I can reopen. I'd prefer to just get rid of the globals all together, which I have some ideas for but little time...

PR #316, opened 2019-08-19
[WIP] Partial transpose atom

joerenes:
An updated attempt at a partial transpose atom. Thanks to @ericphanson for much advice.


ericphanson:
Travis failed on Mac OS for Julia 1.0, but not on Mac Julia 1.1 or Linux, and just because of a timeout. That's strange but seems unlikely to be related to this PR. I'll try closing and reopening the pull request to restart Travais and see if it works.

src/atoms/affine/partialtranspose.jl
```diff
@@ -0,0 +1,104 @@
...
+    end
+    n = length(dims)
+    d = prod(dims)
+    s = n - sys + 1
+    p = collect(1:2n)
+    p[s] = n + s
+    p[n + s] = s
+
+    rdims = reverse(dims)
+    r = reshape(x, [rdims;rdims]...)
```
ericphanson: It's the smallest of nitpicks, but I think `[rdims;rdims]...` is better written as `(rdims..., rdims...)`:
```julia
using BenchmarkTools
julia> @btime reshape($x, [$rdims; $rdims]...);
  584.874 ns (5 allocations: 304 bytes)

julia> @btime reshape($x, ($rdims..., $rdims...));
  477.585 ns (3 allocations: 176 bytes)
```
Vectors (e.g. created by `[ ; ]`) are mutable and need memory allocations, whereas tuples (used in the second option) are immutable and don't. In this case it probably doesn't matter though.

src/atoms/affine/partialtranspose.jl
```diff
@@ -0,0 +1,104 @@
...
+    return reshape(permutedims(r,p),(d,d))
+end
+
+
+function permutedims_matrix(dims, p)
```
ericphanson: Let's add the docstring from https://github.com/ericphanson/QuantumSDPs.jl/blob/97225a7616f08f46de568091dcf63956b2125b1c/src/choi2so.jl#L4-L11

src/atoms/affine/partialtranspose.jl
```diff
@@ -0,0 +1,104 @@
...
+        n = length(x.dims)
+        d = prod(x.dims)
+        s = n - x.sys + 1
+        p = collect(1:2n)
+        p[s] = n + s
+        p[n + s] = s
+
+        rdims = reverse(x.dims)
+
+        partialtranspose_matrix = permutedims_matrix([rdims;rdims],p)
```
ericphanson: same comment here, to use `(rdims..., rdims...)`

test/test_affine.jl
```diff
@@ -417,4 +417,36 @@
...
+        Mt2 = kron(A,transpose(B),C)
+        Mt3 = kron(A,B,transpose(C))
+
+        Rt1 = ComplexVariable(d,d)
+        Rt2 = ComplexVariable(d,d)
+        Rt3 = ComplexVariable(d,d)
+        S = rand(ComplexF64,d,d)
+        solve!(satisfy(partialtranspose(Rt1, 1, dims) == S ),SCSSolver(verbose=false))
+        solve!(satisfy(partialtranspose(Rt2, 2, dims) == S ),SCSSolver(verbose=false))
+        solve!(satisfy(partialtranspose(Rt3, 3, dims) == S ),SCSSolver(verbose=false))
```
ericphanson: here, it should be `solver` instead of `SCSSolver(verbose=false)`, since this testset is in a loop that will try several different solvers

test/test_affine.jl
```diff
@@ -417,4 +417,36 @@
             @test all(abs.(p.constraints[1].dual - dual) .<= TOL)
         end
     end
+
```
ericphanson: Could we add the test for `permutedims_matrix` from here: https://github.com/ericphanson/QuantumSDPs.jl/blob/97225a7616f08f46de568091dcf63956b2125b1c/test/runtests.jl#L18-L28? It should work by just changing `QuantumSDPs.permuteddims_matrix` to `Convex.permuteddims_matrix`.

ericphanson:
I added partial transpose to the docs in https://github.com/JuliaOpt/Convex.jl/pull/317, which I'll merge after this one, by the way

src/atoms/affine/partialtranspose.jl
joerenes: done! Thanks for the tip.

src/atoms/affine/partialtranspose.jl
joerenes: done.

src/atoms/affine/partialtranspose.jl
joerenes: done.

test/test_affine.jl
joerenes: done.

test/test_affine.jl
joerenes: done.

ericphanson:
Thanks a lot for contribution!

Issue #320, opened 2019-08-31
Dev plan


ericphanson:
I'm planning on making some changes and improvements, and I thought I'd write out my plan, in case there are objections and just so these changes aren't coming out of nowhere. It's hard to say how much of this I'll actually get done, but I've consistently been feeling pretty motivated on weekends to work on this, so hopefully it's not too wildly ambitious.

*Ordered priorities:*

- [ ] Add some benchmarking. This is the first priority, because I want to make sure subsequent changes don't hurt performance. I'll probably put up a PR today or tomorrow and merge it fairly quickly, so that subsequent PRs can get benchmarked against master. This won't involve code changes to the Convex.jl `src` folder so there should be little chance of breaking anything.
- [ ] Get rid of global dictionary of variables (`id_to_variables`).  This is half the cause of #83, and I think there's a relatively easy fix. The hard bit is to make sure to not hurt perf too much, since we will have to do the work locally in the problem instead of once at variable construction time, so benchmarking is key.
- [ ] Get rid of the other global variable? I haven't looked into this, but it's used for dual values so it might be nice to do this before the next point. This would then close #83.
- [ ] Switch to MOI. This is maybe 80% done in my branch https://github.com/ericphanson/Convex.jl. I still need to populate dual variables, decide what to do with the various `can_solve_sdp` type methods, and clean up the code a bit (either get rid of MBP or make Convex usable with either MBP or MOI). Benchmarking is important here too, to make sure we haven't hurt model formulation speed with the new internals generating MOI functions and constraints. Ideally, we might get improvements.

*Optional:* 
- [ ] Add `AbstractTrees` dependency + methods, so a Convex.jl problem can be seen as an `AbstractTree` (?). I thought I might need this for (2), but I think I won't now, so it's kind of optional. It's nice because you can use `AbstractTrees.print_tree(p::Problem)` to generate a nice representation of the problem. This might actually give a nicer way to show the problem in the REPL which we could use for `show`. You can also use GraphPlots master and do `plot(TreePlot(p))` where `p::Problem` and get a visual representation of the problem (which was my motivation for https://github.com/JuliaPlots/GraphRecipes.jl/pull/80). The only downside is the dependency, which is 700 lines of Julia with no dependencies itself, which doesn't seem too bad. I might PR this one soon because it's already done locally.
[...]

PR #321, opened 2019-08-31
Add benchmarks

ericphanson:
Uses PkgBenchmark to run benchmarks on each PR, using some nice code from @tkf's Transducers.jl. 

The benchmarks themselves are written with an `@add_benchmark` macro taken from MOI. What this does is allow defining benchmarks which take a callback, and then using them to generate different benchmark suites easily:
```julia
include("benchmarks/benchmarks.jl") # defines module Benchmarks, with the actual benchmarks
const SUITE = BenchmarkGroup()
SUITE["SCS"] = Benchmarks.suite(p -> Convex.solve!(p, SCSSolver(verbose=0)))
SUITE["ECOS"] = Benchmarks.suite(p -> Convex.solve!(p, ECOSSolver(verbose=0)); exclude = [r"sdp", r"SDP"])
SUITE["formulation"] = Benchmarks.suite(Convex.conic_problem)
```
Here, you pass a callback for what to do with the problem (solve it with ECOS, with SCS, or just generate the MBP standard form with `Convex.conic_problem`). This runs all the benchmarks except those excluded by a regex search. This could also be used for comparing the speed of solvers on Convex.jl problems; maybe at some point we could even incorporate benchmark data into the docs, all via travis.

The actual benchmark problems themselves are mostly taken from Convex.jl's tests, and are certainly incomplete: they cover some of the affine and sdp tests only. I think they might be enough however to see the effect of large changes which affect all atoms more-or-less the same. Ideally at some point we could have benchmarks for every atom, like we have tests for each atom.

Hopefully I've got the CI bits done correctly. I suspect the tests will fail on this PR, because it will try to compare benchmarks to master, which doesn't have any, but that doesn't mean there's something wrong. The real test will be the next PR after this one is merged.

Edit: no, this PR passed because Travis doesn't use the PR's .travis.yml. That makes sense :). 

This is priority 1 on https://github.com/JuliaOpt/Convex.jl/issues/320.


ericphanson:
I'll go ahead and merge this! Only thing that can go wrong is CI and I'll fix that if it's wrong.

PR #326, opened 2019-09-03
Sort eigenvalues to ensure consistent behaviour

JiazhengZhu:
Currently GenericLinearAlgebra is necessary to evaluate eigenvalues of BigFloat arrays. But its resultant values do not follow the same convention as eigvals for Float64 arrays.

### DIFF
--- a/src/atoms/sdp_cone/lambda_min_max.jl
+++ b/src/atoms/sdp_cone/lambda_min_max.jl
@@ -39,7 +39,8 @@ function curvature(x::LambdaMaxAtom)
 end
 
 function evaluate(x::LambdaMaxAtom)
-    eigvals(evaluate(x.children[1]))[end]
+    #eigvals(evaluate(x.children[1]))[end]
+    sort(eigvals(evaluate(x.children[1])), by = x -> real(x))[end]
 end
 
 lambdamax(x::AbstractExpr) = LambdaMaxAtom(x)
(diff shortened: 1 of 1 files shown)


src/atoms/sdp_cone/lambda_min_max.jl
```diff
@@ -39,7 +39,8 @@ function curvature(x::LambdaMaxAtom)
 end
 
 function evaluate(x::LambdaMaxAtom)
-    eigvals(evaluate(x.children[1]))[end]
+    #eigvals(evaluate(x.children[1]))[end]
+    sort(eigvals(evaluate(x.children[1])), by = x -> real(x))[end]
```
ararslan:
```suggestion
    sort!(eigvals(evaluate(x.children[1])), by=real)[end]
```
Same below. Should also delete the line above instead of leaving it commented out.

ararslan:
Could you add a test for a case that this change fixes?

ericphanson:
I think maybe it's better to just PR GenericLinearAlgebra to match sorting as base does. I'm not really opposed to making the change here, but it seems like we could fix it at the source.

ararslan:
I agree, that does seem like a better course of action.

ericphanson:
Just to say @JiazhengZhu and I discussed this (we're working together over at https://github.com/ericphanson/SDPA_GMP.jl, which is where the problem came up) and he's ok with closing this in favor of waiting for resolution over at GenericLinearAlgebra (or making a PR there). So I'll close this.

PR #330, opened 2019-09-11
Switch to MathOptInterface

### LINKED ISSUE #262: Switch to MathOptInterface
It is intended to supersede MathProgBase. JuMP is also making this transition.

### LINKED ISSUE #288: `conv` fails for complex inputs

### LINKED ISSUE #337: Dual values of complex constraints wrong, poisons later dual values
E.g.
```julia
using Convex, SCS, LinearAlgebra
solver = SCSSolver(verbose=0)
x = ComplexVariable(2, 2)
p = minimize( lambdamax(x), [ x[1,2] == im, x[2,2] == 1.0, x ⪰ - I(2) ] )
Convex.solve!(p, solver)

@show p.optval
@show p.constraints[1].dual
@show p.constraints[2].dual

using MATLAB
mat"""
cvx_begin sdp quiet
variable x(2, 2) complex hermitian
dual variable cvx_dual1
dual variable cvx_dual2
minimize( lambda_max(x) )
subject to
    cvx_dual1 : x(1,2) == j
    cvx_dual2 : x(2,2) == 1.0
    x >= -eye(2)
cvx_end
$(cvx_optval) = cvx_optval;
$(cvx_dual1) = cvx_dual1;
$(cvx_dual2) = cvx_dual2;
"""
@show cvx_optval
@show cvx_dual1
@show cvx_dual2;
```
yields
```julia
p.optval = 1.4999999802464283
(p.constraints[1]).dual = 5.587176841458554e-16
(p.constraints[2]).dual = 0.9999999500234388
cvx_optval = 1.4999999265147386
cvx_dual1 = 0.0 + 1.0000954705004044im
cvx_dual2 = 0.7499522176931694
```
Why? In

https://github.com/JuliaOpt/Convex.jl/blob/e0513acc7a2fa3c22325c7ea9e528b0a340f068b/src/solution.jl#L194-L214

the dual values are taken out of a big vector of dual values returned by MPB by using the constraint's size to determine the indexing. However, for complex dual values, they have been written as a long vector, and we only take the first half-- the real parts. Then the next dual value takes the next bit of the big vector of dual values, which is then can include the complex part of the previous dual value. So here, `(p.constraints[1]).dual + im*(p.constraints[2]).dual` is actually the true dual value for the first constraint (and hence matches `cvx_dual1`). And we've totally lost the dual value for the 2nd constraint (i.e. `cvx_dual2`).

I found this because MOI doesn't return a big vector of dual values, but instead returns the dual values for each constraint as you ask for it. Then when I ported the tests from #332 to the MOI branch (#330), I found I was missing a reshape in the MOI code. Adding the reshape added a size check, and I found MOI was returning double the number of values than the size expected (since it returns the real and complex parts). When I went to check how we handled this with MPB, I found that we did not, hence this issue.

This will be fixed with #330 (once I push some commits up), and I'll add this example as a test there.

ericphanson:
Still needed to match features under MPB:

- [x] Dual values for affine problems
- [x] Warmstarts

~~Some things are left to do, hence the (WIP)~~, but I thought I'd open this now to get CI on it and in case anyone wanted to look it over already.

Closes #262, and addresses the last high priority point of #320. Also fixes #288 and fixes #337.

~~The diff looks really big, but that's because this PR is built on the ProblemDepot (i.e. the first commit here), so skip that to see the MOI changes. The ProblemDepot is in https://github.com/JuliaOpt/Convex.jl/pull/328.~~ The ProblemDepot was merged and this is rebased on that, so the diff isn't quite so gigantic.

The commit "Add initial MOI support …" reuses the MPB standard form to send the problem to MOI; this is a bit wasteful but it gets things working. The next commit, "Rewrite `load_problem` to not construct MBP form" instead directly sends the problem to MOI, which lets us avoid constructing more than we need to. I include both because it's harder to understand the final code without seeing the transition.

Some decisions and the reasoning behind them:

- Can we support both MPB and MOI at the same time? I thought yes at first, but then I remembered they have different representations of SDP variables, which means we'd need to support two separate `conic_form!`s, which would mean we need a mechanism to choose the `conic_form!` based on the solver. That seemed too complicated.
- Why base this PR on the Problem Depot PR? It made writing the tests easier, because with MOI we need to use a new solver object for each solve, so it doesn't work to have a global `solvers = [SCS.Optimizer(), ECOS.Optimizer()]`; we could pass around a vector of functions `() -> SCS.Optimizer()` and then call it with `solver()` instead of `solver`, but that seems less elegant than the Problem Depot callback. Moreover, this let me get rid of the `can_solve_*` functions which otherwise would need to be updated for MOI. I think they are only used for the Convex.jl tests (although getting rid of them is definitely breaking), and Problem Depot solves the same problem in a more nuanced way.


chriscoey:
Wow, nice. IMO not worth supporting MPB.

ericphanson:
I came to the same conclusion— actually, the commit called “Try allowing both MBP and MOI” was supposed to be called “Remove MathProgBase”, I just messed up the rebase.

src/problems.jl
```diff
@@ -107,121 +87,34 @@ function conic_form!(p::Problem, unique_conic_forms::UniqueConicForms=UniqueConi
     return objective, objective_var.id_hash
 end
 
-function conic_problem(p::Problem)
```
rofinn: I may have missed it, but does this function not exist anymore? It's still referenced in the updated docs, but that appears to be it?

test/runtests.jl
```diff
@@ -1,45 +1,55 @@
...
-    include("test_affine.jl")
-    include("test_lp.jl")
-    include("test_socp.jl")
-    include("test_sdp.jl")
-    include("test_exp.jl")
-    include("test_sdp_and_exp.jl")
-    include("test_mip.jl")
+
+    @testset "SCS" begin
+        run_tests(; exclude=[   r"mip",
```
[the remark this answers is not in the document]
ericphanson: Thanks very much for the review! I kind of prefer the current way, since it makes it clear that the bug is only with `sdp_matrix_frac_atom`; as you might know, the fact that SCS cannot handle mixed-integer problems is just a limitation of SCS (which we used to gate around by `can_solve_mip` conditionals in the tests-- now with the new system, we have to be explicit, which I think is an advantage). What is the advantage of putting the comment above?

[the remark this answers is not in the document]
rofinn: > What is the advantage of putting the comment above?

Just that it makes the actual code easier to read and doesn't require lining up the elements of the array in a certain way.

src/problems.jl
rofinn: Cool, that all seems reasonable. I just wanted to clarify that `load_MOI_model!` was the closest thing to what this previously did and that we want to remove that reference from the docs. Thanks for fixing that up already 👍

test/runtests.jl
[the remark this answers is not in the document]
ericphanson: Ok, fair enough. I moved the comment up but added a few words to say that it's the `sdp_matrix_frac_atom` which is the test which hits the SCS bug. Hopefully then we have the best of both.

ericphanson:
I added variable warmstarts just now, as well as some comments and cleanup[

Actually, I had variable warmstarts working about a month ago but I hoped to add constraint dual warmstarts as well, but I had a bit of trouble getting it working (minimal example: https://gist.github.com/ericphanson/9ab9f8460bbe5200be82d0b0cb55ecd5), either due to a bug in MOI or a bug in my understanding :). However, as @mlubin pointed out on the JuMP-dev gitter, we don't support constraint dual warmstarts under MPB (we can see this directly in the code, shown below), so we probably shouldn't consider this as blocking for merging this PR.

Hence, to my knowledge this PR now matches the current functionality we have, except for the obvious changes of not supporting MPB-only solvers such as Pajarito (currently), while adding support for MOI-only solvers, such as SDPAFamily.jl.

The only user-facing change should be accessing solvers via e.g. `SCS.Optimizer()` instead of `SCSSolver()`.


----

No constraint warmstarts: See

https://github.com/JuliaOpt/Convex.jl/blob/e0513acc7a2fa3c22325c7ea9e528b0a340f068b/src/solution.jl#L41-L43

and

https://github.com/JuliaOpt/Convex.jl/blob/e0513acc7a2fa3c22325c7ea9e528b0a340f068b/src/solution.jl#L55-L88

src/problems.jl
```diff
@@ -1,81 +1,41 @@
...
-                        var_to_ranges[id] = (index + 1, index + 2*length(var))
-                        index += 2*length(var)
-                    else
-                        var_to_ranges[id] = (index + 1, index + length(var))
-                        index += length(var)
-                    end
-                end
-            end
-            constr_size += constraint.sizes[i]
+function Base.getproperty(p::Problem, s::Symbol)
```
ericphanson: I'm open to feedback on this one. The reason I made `optval` lazy in this way is so that we don't need to query `MOI.ObjectiveValue()` during the `solve!` command (which would be eagerly populating `optval`). That way if the solver did not find a solution and hence may throw if you ask for its objective value, we don't error during `solve!` and you still get the status and termination info, along with whatever else the solver returned. This also matches how JuMP does it.

We could query to ask MOI if there is a primal feasible solution etc and only populate under those circumstances as well. It wasn't entirely clear to me exactly what conditions will guarantee `get(model, MOI.ObjectiveValue())` will not throw, however.

src/utilities/show.jl
```diff
@@ -176,7 +176,13 @@ end
 
 function show(io::IO, p::Problem)
     TreePrint.print_tree(io, p, MAXDEPTH[])
-    print(io, "\ncurrent status: $(p.status)")
+    if p.status == MOI.OPTIMIZE_NOT_CALLED
+        print(io, "\nstatus: `solve!` not called yet")
+    else
+        print(io, "\ntermination status: $(p.status)")
+        print(io, "\nprimal status: $(primal_status(p))")
+        print(io, "\ndual status: $(dual_status(p))")
```
ericphanson: MOI returns us lots of status information (termination status, primal status, dual status), and we might as well give it all to the user, otherwise they probably won't know it's there.

test/test_utilities.jl
```diff
@@ -1,5 +1,12 @@
 @testset "Utilities" begin
 
+    @testset "`solve!` does not return anything" begin
```
ericphanson: This isn't a new test; it was just moved around during the problem depot changeover and ended up here

test/runtests.jl
```diff
@@ -1,20 +1,26 @@
...
+const MOI = MathOptInterface
+const MOIU = MOI.Utilities
 
 # Seed random number stream to improve test reliability
 using Random
 Random.seed!(2)
 
 @testset "ProblemDepot" begin
-    @testset "Problems can run without `solve!`ing if `test==false`" begin
+    @testset "Problems can run without `solve!`ing if `test==false`; T=$T" for T in (Float64, BigFloat)
```
ericphanson: We check that our tests run even when the problem is specified with high precision numeric types

ericphanson:
Ok, I tried to help make this more reviewable by merging #328 and adding some review comments. This is good to go from my side.

nickrobinson251:
Thanks so much for all your work on this, @ericphanson!

We use Convex in some private code, and I will try switching it to use this branch to double check everything still runs as expected :)

nickrobinson251:
Will this is released as v0.13?

ericphanson:
@nickrobinson251 awesome! Please let me know if there are any issues.

Yes, my plan was to release this as v0.13.0, maybe also with #336 fixed (`lambdamin` and `lambdamax` deprecated in favor of `eigmin` and `eigmax`) to try to group the breaking changes together. And maybe I can add a deprecation for `solve!(problem, solver)` for `MathProgBase` solver arguments to guide users to the MOI syntax.  It might be good to then release all that without deprecations as 1.0.

nickrobinson251:
> We use Convex in some private code, and I will try switching it to use this branch to double check everything still runs as expected :)

hmm, I switched to this branch and updated the private code from `ECOSSolver <: MathProgBase.SolverInterface.AbstractMathProgSolver`  to instead use `ECOS.Optimizer <: MathOptInterface.AbstractOptimizer` (on a quadratic optimization problem) 

I get an error from `moi_populate_solution!`:
```julia
  DimensionMismatch("new dimensions (0, 1) must be consistent with array size 659")
  Stacktrace:
   [1] (::Base.var"#throw_dmrsa#197")(::Tuple{Int64,Int64}, ::Int64) at ./reshapedarray.jl:41
   [2] reshape at ./reshapedarray.jl:45 [inlined]
   [3] unpackvec(::Array{Float64,1}, ::Tuple{Int64,Int64}, ::Bool) at /Users/nick/repos/Convex.jl/src/solution.jl:311
   [4] moi_populate_solution!(::MathOptInterface.Bridges.LazyBridgeOptimizer{MathOptInterface.Utilities.CachingOptimizer{ECOS.Optimizer,MathOptInterface.Utilities.UniversalFallback{MathOptInterface.Utilities.Model{Float64}}}}, ::Convex.Problem{Float64}, ::OrderedCollections.OrderedDict{UInt64,Convex.Variable}, ::Dict{Convex.ConicConstr,Convex.Constraint}, ::Array{Convex.ConicConstr,1}, ::Dict{UInt64,Array{MathOptInterface.VariableIndex,1}}, ::Array{MathOptInterface.ConstraintIndex{MathOptInterface.VectorAffineFunction{Float64},S} where S,1}) at /Users/nick/repos/Convex.jl/src/solution.jl:287
   [5] #solve!#14(::Bool, ::Bool, ::Bool, ::typeof(Convex.solve!), ::Convex.Problem{Float64}, ::ECOS.Optimizer) at /Users/nick/repos/Convex.jl/src/solution.jl:231
   [6] solve!(::Convex.Problem{Float64}, ::ECOS.Optimizer) at /Users/nick/repos/Convex.jl/src/solution.jl:209

```

The problem seems to be that `constr.size == (0, 1)`.

Adding an `@show` to Convex.jl/src/solution.jl:283
```julia
constr = conic_constr_to_constr[conic_constr] = <= constraint (affine)
├─ index (affine; real)
│  └─ 659-element real variable (id: 657…592)
└─ 0
```

ericphanson:
Thanks very much for the report. Strange-- I think `constr.size == (0, 1)` should never happen. The size should be set here:

https://github.com/JuliaOpt/Convex.jl/blob/cba179feeaa7934e0e70011ef74c181a7e88ccbd/src/constraints/constraints.jl#L75-L78

so in this case it should be `(659, 1)`. Now, `LtConstraint` is a mutable struct so it could get changed somewhere, somehow. Just to check, `constr.size == (0, 1)` and `constr isa LtConstraint` both hold? And `constr.lhs.size == (659, 1)` and `constr.rhs.size == (1,1)`?

ericphanson:
Before merging this, it might be good to think a bit about the user-facing API. Right now, if a user does eg
```julia
optimizer = SCS.Optimizer(verbose=0)
p = ...
solve!(p, optimizer)
q = ...
solve!(q, optimizer)
```
This won't work, since `solve!` expects a fresh optimizer instance for each solve. JuMP currently handles this by the `with_optimizer` construction which holds the details needed to construct an optimizer (i.e. the constructor and its arguments). This is being discussed to be changed in https://github.com/JuliaOpt/JuMP.jl/pull/2090 to instead just pass the constructor and then instead of using arguments to the constructor, set them via calls like `set_parameter(model, verbose, 0)`.

We could just ask the user to always give us a fresh instance. In other words, the example above should be written eg
```julia
optimizer() = SCS.Optimizer(verbose=0)
p = ...
solve!(p, optimizer())
q = ...
solve!(q, optimizer())
```
and we just document / redirect users to that way of interacting with Convex. Or we can think about other ways, such something like what JuMP is doing.

ericphanson:
Also, @nickrobinson251 if you have any more details or followup, I'd be quite happy to help get to the bottom of it, particularly if there's a bug in the MOI code above. You can also message me on the Julia slack (I'm `@ericphanson` there too) if that's easier.

nickrobinson251:
Thanks! I'll try to look at this again tomorrow :)

src/problems.jl
mlubin: Is this the only instance where you override `getproperty`? Why not deprecate `p.optval` in favor of `objective_value` that's implemented below?

src/solution.jl
```diff
@@ -1,214 +1,332 @@
...
-    m = problem.model
-    load_problem!(m, c, A, b, cones, vartypes)
-    if warmstart
-        set_warmstart!(m, problem, length(c), var_to_ranges, id_to_variables)
-    end
-    # optimize MPB conic problem
-    MathProgBase.optimize!(m)
+# This function generates a MOI set and VectorAffineFunction corresponding to a constraint that
+# Convex represents with a `ConicConstr` object.
+# This is part of what used to be `Convex.conic_problem` (under MPB instead of MOI).
```
mlubin: Historical comments referencing how things used to be in MPB won't be too useful for someone reading this code in the future. These comments are throughout this file.

ericphanson: Good point. I added those to help with review, but I probably should have put them as review comments instead of code comments.

ericphanson:
> Thanks! I'll try to look at this again tomorrow :)

Friendly ping @nickrobinson251 :)

nickrobinson251:
Yeah, sorry about that! It's possible I'll be able to investigate this before the weekend but unlikely, so if not I think just go ahead - don't let me hold up the pr.

ericphanson:
> Yeah, sorry about that! It's possible I'll be able to investigate this before the weekend but unlikely, so if not I think just go ahead - don't let me hold up the pr.

Ok, sounds good. I think this might sit on master for a bit anyway even after it's merged, because we'll want to finish all the [v0.13-tagged stuff](https://github.com/JuliaOpt/Convex.jl/issues?q=is%3Aopen+is%3Aissue+milestone%3Av0.13.0) first anyway before tagging a release.

ericphanson:
Ah good, CI is failing because the examples need to be updated for MOI. (Wasn't sure if that would fail the docs build or not, but glad to see it does).

ericphanson:
Ok, I think all review comments have been addressed (modulo #346 which I'll leave for a separate PR), and with the docs and examples updated and codecov passing, I think we're good to go here.

iamed2:
Fantastic!

PR #333, opened 2019-10-04
remove eye from examples; closes #331

### LINKED ISSUE #331: eye is no longer supported in new julia
examples-->basic_usage， first linear programming, eye() function still used here.

ranocha:


ericphanson:
Thanks very much for the PR!

Reviewing this made me think we should do the examples differently (https://github.com/JuliaOpt/Convex.jl/issues/335), but that's not really related to the content of this PR.

For this PR-- I think we should make sure to support the LTS version (Julia 1.0) in these notebooks otherwise it could cause confusion for users if things don't work. I saw two changes we need, one for identity matrices and one for diagonal matrices. Can you add the lines
```julia
if VERSION < v"1.2.0-DEV.0"
    (I::UniformScaling)(n::Integer) = Diagonal(fill(I.λ, n))
     LinearAlgebra.diagm(v::AbstractVector) = diagm(0 => v)
end
```
after `using LinearAlgebra` in each of those notebooks you're updating? That will backport the change. (The `diagm` isn't need in `basic_usage` though so it could be skipped there).

Also, in `basic_usage`, we need `using SCS` instead of `import SCS`, or else need to qualify each `SCSSolver` with `SCS.SCSolver` (only the first one is qualified currently).

ranocha:
That's a good idea. I've made the corresponding changes.

ericphanson:
Great, thanks very much!

ranocha:
You’re welcome!

Issue #335, opened 2019-10-10
Switch examples to Weave or Literate


ericphanson:
It's a bit hard to review PRs to jupyter notebooks (eg https://github.com/JuliaOpt/Convex.jl/pull/333); I think we should switch to generating them from Weave like DifferentialEquations.jl did (https://github.com/JuliaDiffEq/DiffEqTutorials.jl) or to Literate (using eg https://github.com/oxinabox/ipynb2Literate.jl). Then the notebooks get generated from the code which makes the diffs much simpler, and we could also automatically test that the files run without errors to make sure they don't get out of date.

Issue #338, opened 2019-10-30
helping with updating Examples


PaulSoderlind:
Hi,

I would like to make a PR for updating some of the finance examples (portfolio_optimization...) you have.  However, I ask for some instructions.

1. Do you prefer a miniminal update, or a bit more (improving the style, fixing the LaTeXing, and perhaps even improving the link to standard "finance" approaches)?
2. The current examples use both SCS and ECOS. I would probably start with SCS and only use ECOS if needed. OK?
3...and one more thing: the examples include plots (using various packages). Should I perhaps standardise this to PyPlot.jl or Plots.jl?

ericphanson:
I'm very glad to hear it! For point 1, I think either way is fine. The most important thing is getting the code running on Julia 1.0 so that users can play around with it themselves without running into errors. Besides that, I think any extra references or styles fixes are nice to have too.

For the solvers, I think any open source solver that can be automatically installed is fine (that way the users don't have to do anything more than add the packages). So ECOS, SCS, GLPK, etc-- as you wish.

For plotting packages, I don't have a strong preference, I would lean a bit towards standardizing towards one package. It would be nice to have a single `Project.toml` file that can be used for all the examples to install all the packages we need (which is what JuMP does: https://github.com/JuliaOpt/JuMP.jl/tree/master/examples), and if that has a bunch of different packages then that's more upfront loading time to get it all installed.

What do you think? I think if you're the one putting in the time to update examples, your opinion probably counts the most :). 

One thing that I think is important, however, is https://github.com/JuliaOpt/Convex.jl/issues/335. I think it would be great to generate the notebooks instead of writing them directly, since that makes reviewing PRs and seeing the changes much easier, and also allows writing to multiple output formats. I think it could even be nice to have the examples in a way that we could include them directly in the documentation (under a subsection), along with links to notebooks for users to run and modify themselves. It seems like Literate can do this: https://fredrikekre.github.io/Literate.jl/stable/documenter/, and it seems like Literate is what was used for the JuliaAcademy resources: https://github.com/JuliaComputing/JuliaAcademyMaterials/. I think we could use their `unliterate` script to transform the current notebooks into literate files, or https://github.com/oxinabox/ipynb2Literate.jl (I haven't tried either but both seem to have the same goal).

[...]

PR #261, opened 2018-12-22
WIP: Update examples to 1.0

brilhana:
I've started updating examples to `1.0` as some dated from years ago. This is a work in progress.

PR #339, opened 2019-10-30
Switch examples to use Literate.jl and update to 1.0

### LINKED ISSUE #201: Update examples to reflect new syntax for `dot(*)`
Elementwise operations must work differently in Julia v0.6 due to breaking changes in broadcast; but I'm told we will likely be able to  revert to the previous `.*` syntax in future versions. In light of that, we should update the examples so that 

* they do not use the infix operators `.*`, `./`, or `.^`, but instead use `dot(*)`, `dot(/)`, and `dot(^)`
* it will be easy to revert back to the original syntax when the fix to broadcast happens in Julia v1.0.

### LINKED ISSUE #268: Running error in Control example
I tried to run Control example, but it does not work. 

The error message said 

>MethodError: no method matching +(::Array{Any,1}, ::EqConstraint)
Closest candidates are:
  +(::Any, ::Any, !Matched::Any, !Matched::Any...) at operators.jl:502
  +(::Array, !Matched::Array...) at arraymath.jl:44
  +(::Array, !Matched::SparseArrays.SparseMatrixCSC) at C:\cygwin\home\Administrator\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.0\SparseArrays\src\sparsematrix.jl:1574
  ...
Stacktrace:
 [1] top-level scope at .\In[5]:25

So, I bracketed all the constraints related code. For example

>From
   constraints += position[:, i + 1] == position[:, i] + h * velocity[:, i]
to
  constraints += [ position[:, i + 1] == position[:, i] + h * velocity[:, i] ] 

Then the different error message appeared, like this

>DimensionMismatch("dimensions must match")
Stacktrace:
 [1] promote_shape at .\indices.jl:154 [inlined]
 [2] promote_shape(::Array{Any,1}, ::Array{EqConstraint,1}) at .\indices.jl:145
 [3] +(::Array{Any,1}, ::Array{EqConstraint,1}) at .\arraymath.jl:45
 [4] top-level scope at .\In[6]:25
========================================

I am using Julia 1.0.2.

Is the modifying (adding brackets) appropriate? If it is, what is the problem with the following code?

>using Convex, SCS
#Some constraints on our motion
#The object should start from the origin, and end at rest
initial_velocity = [-20; 100]
final_position = [100; 100]

>T = 100 #The number of timesteps
h = 0.1 #The time between time intervals
mass = 1 #Mass of object
drag = 0.1 #Drag on object
g = [0, -9.8] #Gravity on object

>#Declare the variables we need
position = Variable(2, T)
velocity = Variable(2, T)
force = Variable(2, T - 1)

>#Create a problem instance
mu = 1
constraints = []

>#Add constraints on our variables
for i in 1 : T - 1
  constraints += [ position[:, i + 1] == position[:, i] + h * velocity[:, i] ]
end

>for i in 1 : T - 1
  acceleration = force[:, i]/mass + g - drag * velocity[:, i]
  constraints += [ velocity[:, i + 1] == velocity[:, i] + h * acceleration ]
end

>#Add position constraints
constraints += [ position[:, 1] == 0 ]
constraints += [ position[:, T] == final_position ]

>#Add velocity constraints
constraints += [ velocity[:, 1] == initial_velocity ]
constraints += [ velocity[:, T] == 0 ]

>#Solve the problem
problem = minimize(sumsquares(force), constraints)

ericphanson:
Closes https://github.com/JuliaOpt/Convex.jl/issues/335, closes #268, closes #201 (which I'm not sure is relevant now anyway, but with the examples updated it surely can be closed). Supersedes #261.

I converted the examples to use Literate.jl. The notebook files can be generated by `examples/build.jl`, which is automatically done in `docs/make.jl` so that we can serve the built notebooks in the docs. ~~There may be a better way to do the docs building; currently, Documenter emits warnings for the links to the notebooks~~ (edit: just had some bad paths).

I also would like to render an HTML version so the user doesn't need to open the Jupyter notebooks necessarily. That requires the notebooks to run though! Hence I've updated all the notebooks to 1.0, and standardized plotting on Plots.jl as suggested in #338. I use a mix of ECOS and SCS since ECOS is better suited for some problems. The `docs/make.jl` file now uses Literate to build markdown versions of all the examples which it includes in the docs.

To do:
- [x] Provide a way for the user to download Jupyter notebooks. Currently I generate them during the docs build step, but don't link to them anywhere. The catch is that some examples need auxiliary files.
- [x] Check for more latex / render errors.
- [x] Check that they all run on 1.0. I've been using 1.3-rc3, so possibly I've used features that aren't available in 1.0.
- [x] Update the `examples/build.jl` script so it's easy to work with one example at a time (if you're updating or adding a new one).
- [x] Update readme to no longer say the examples are out of date

---
Code for converting examples:

This was run in the `examples` directory.
```julia
import JSON

function unliterate(path, output)
    @assert !isdir(path)
    @assert !isdir(output)
    endswith(path, ".ipynb") || throw(ArgumentError("only Jupyter notebooks are supported"))
    nb = open(JSON.parse, path, "r")
    out = open(output, "w")
    prev_type = ""
    if !haskey(nb, "cells")
        apath = abspath(path)
		  # need to update notebook format
        run(`python -m nbconvert --to notebook --nbformat 4 $apath --output $apath`)
        @warn "Upgrading to nb4 " path 
        nb = open(JSON.parse, path, "r")

    end
    for cell in nb["cells"]
        type = cell["cell_type"]
        type == prev_type && print(out, "#-\n\n")
        prev_type = type
        
        if type == "markdown"
            for line in cell["source"]
                print(out, "#")
                if !isempty(strip(line))
                    print(out, " ")
                    print(out, line)
                else
                    print(out, "\n")
                end
            end
            print(out, "\n\n")
        elseif type == "code"
            for line in cell["source"]
                startswith(line, "# ") && print(out, "#")
                print(out, line)
            end
            print(out, "\n\n")
        else
            error("unknown cell type $type")
        end
    end
    close(out)
    return
end

function literate_folder(dir, output)
    !isdir(output) && mkdir(output)
    for file in readdir(dir)
        file = joinpath(dir, file)
        if endswith(file, ".ipynb")
            file_out = joinpath(output, split(basename(file), ".")[1] * "_literate.jl")
            @info "Unliterating" file file_out
            unliterate(file, file_out)
        elseif !isdir(file)
            file_out = joinpath(output, basename(file))
            @info "Copying" file file_out
            cp(file, file_out; force=true)
        end
    end
end

examples_path = "."
literate_path = "../literate"
!isdir(literate_path) && mkdir(literate_path)
for dir in filter(isdir, readdir(examples_path))
    dir == literate_path && continue
    output = joinpath(literate_path, dir)
    literate_folder(dir, output)
end

dir = "."
output = joinpath(literate_path, dir)
literate_folder(dir, output)
```
which was modified from <https://github.com/JuliaComputing/JuliaAcademyMaterials/blob/master/unliterate.jl>.


ericphanson:
## Some reasoning for decisions I made in this PR

* Why move examples to a subdirectory of docs? I initially had each folder of examples have its own project. That was kind of nice so you didn't install Dataframes and RDatasets if you were working just with examples that don't need them. But then I needed to activate each folder in order to run the code when generating the documentation. It was pointed out to me on #pkg-usage on the Julia slack that this was a bad idea, loading packages then activating new environments and loading more packages. I didn't want to spawn new Julia processes to run the code (it's slow enough already), so I decided to make a single environment for all the examples. Then we're down to two environments: one for the docs folder and one for the examples, and we have the same problem (want to activate two environments). So instead I realized it was simpler to have a nested structure of `docs/examples/` instead of separate `docs/` and `examples/` and try to copy files out of `examples/` when generating the docs. Since the examples are embedded in the docs now, I think the subdirectory structure makes sense.
* Why provide the Jupyter notebooks as one zip file? I was initially thinking that for each example I would generate a link to the Jupyter-notebook version of that example. The problem is that some examples need auxiliary files, so then I'd need a way to provide a link to a bundle of files. A zip of files was a natural way to do that, but at that point it seems reasonable to just bundle all the examples instead of requiring someone to download each one individually. This also has the advantage that we can provide a README and a Project.toml to try to make it easier to run them.
* Why generate the Jupyter notebooks with `execute=false`? Docs build time is already 20 minutes on Travis (it's only a few minutes on my laptop though; I think part of the difference in speed is that it has to install all the packages each time on Travis), so it would just be longer to execute all the notebooks. The advantage would be if one of the notebooks doesn't run, we would actually get an error in that case, and CI would fail, which would be ideal. Currently, if there is a problem executing the code in the notebooks, we just get warnings from Documenter but CI will still pass. So maybe this should be revisited.

## Followup things to do

* Have CI fail if one of the examples doesn't run; possibly `execute=true` on the notebooks as mentioned above. Another possibility is `strict=true` on `makedocs`; however, we currently generate a few warnings which will then fail CI.
* Have a way to preview the docs from a PR. This has been done here: https://github.com/fredrikekre/Literate.jl/pull/75 and is tracked here: https://github.com/JuliaDocs/Documenter.jl/issues/1131. This way docs PRs can be visually inspected to see if they look right (examples ran, formatting, etc).
* Unify data formats? Currently we use DelimitedFiles, MAT, and hardcoded data in `.jl` files. It would be nice to stick to one.
* Have a way to render single examples to html without having to call `makedocs` (which generates all the examples and hence is slow). Documenter.jl doesn't seem to provide a method to do this; I tried also rendering the markdown files from Literate using Weave, but I ran into a problem with the paths to auxiliary data (`@__DIR__` pointing to the wrong place). It also doesn't look the same as it does coming from Documenter (different CSS etc). However, since the Literate `.jl` files are directly runnable and one can just use an external markdown previewer, the formatting + code of single examples can be each checked separately without needing to call `makedocs`, so the fact that we are lacking this point doesn't seem so troublesome.

PRs welcome :)

ericphanson:
Travis passes; I think the coveralls decrease is spurious since I haven't changed anything in /src or /test.

I'll go ahead and merge this-- it would be nice to preview to make sure everything works (it does locally at least), but I'll just fix it up after if there are problems.

fredrikekre:
Nice LOC net delettion here: `+1,866 −41,342` :)

PR #340, opened 2019-10-31
Finance examples

PaulSoderlind:
Hi,
portfolio_optimization.ipynb is now updated and also rewritten to be closer to standard finance notation. 

portfolio_optimization2.ipynb is also updated and combines the previous files portfolio_optimization2.ipynb and portfolio_optimization3_unconstrained_markowitz_efficient_frontier.ipynb. (There was a lot of overlap.) The latter file is now redundant.


ericphanson:
Thanks very much for these updates. I'll look over this soon, and probably merge this, then update #339 with the changes and then merge that.

ericphanson:
I don't really know enough about finance to judge well, but they seem like good changes to me :).

PR #347, opened 2019-11-28
Update to Documenter 0.24

ericphanson:
Documenter 0.24 uses KaTeX by default instead of mathjax (which is faster and looks nicer but supports a smaller subset of latex), so I updated the docs to not use `mbox` (substituted `text`), and I also fixed up a few things that I saw while going through the docs.


ericphanson:
Preview here: https://www.juliaopt.org/Convex.jl/previews/PR347/

ericphanson:
I don't see any more LaTeX errors or other pressing doc fixes, so I'll go ahead and merge this. Coveralls is complaining again but since these were docs-only changes, I don't understand how coverage could have changed.

Issue #349, opened 2019-12-05
"aux" is not an allowed directory name under windows


reumle:
and this prevents (obviously) git from creating that directory when cloning your repo. 

In case it would be a very small cost (for you and other users), could you consider amending those names (relevant directories are under docs/examples_literate/…)? In the meantime I'll find a workaround.

Here the reference from 2003.....(other names are also invalid: "con" "prin"..) 
https://coderanch.com/t/131585/engineering/Folder-con-Windows-invalid

ericphanson:
Oops! I had no idea. I'll put up a PR shortly to rename them.

reumle:
Thanks and greetings from happy cloner!

PR #350, opened 2019-12-05
Move aux -> aux_files and exclude test

ericphanson:
Fixes https://github.com/JuliaOpt/Convex.jl/issues/349


ericphanson:
Travis failed due to a SCS bug (https://github.com/JuliaOpt/SCS.jl/issues/167), so I'll exclude it until that's fixed (since having tests fail every PR due to a bug we now know about isn't helpful!).

ericphanson:
Coverage went down since we lost a test and SCS is the only SDP solver we test on. I'll have a look at adding another SDP solver to our tests, but I think that's outside the remit of this PR so I'll merge this first to fix #349.

PR #351, opened 2019-12-06
the tutorial is using syntax that now causes errors

danspielman:
Changed "SCS.Optimizer" to "SCSSolver"

### DIFF
--- a/docs/src/quick_tutorial.md
+++ b/docs/src/quick_tutorial.md
@@ -33,7 +33,7 @@ x = Variable(n)
 problem = minimize(sumsquares(A * x - b), [x >= 0])
 
 # Solve the problem by calling solve!
-solve!(problem, SCS.Optimizer(verbose=false))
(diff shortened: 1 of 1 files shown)


ericphanson:
It looks like you’re on the “dev” docs— that’s for the master branch which isn’t released yet (and the syntax there is still in flux but as you see it will be different from that of the current release). The “stable” docs (https://www.juliaopt.org/Convex.jl/stable/) correspond to the current release of Convex.jl, and have the right syntax for that.

ericphanson:
I’ll close this because we don’t want to make this change (I actually just updated these tutorials to the SCS.Optimizer syntax a few days ago on master), but please feel free to write here or in an issue if you have any questions or comments; I’m very willing to answer them if I can.

ericphanson:
Ah, I realized the README was calling the dev docs as “latest”, which is a bit misleading, so I changed that in https://github.com/JuliaOpt/Convex.jl/pull/352.

PR #360, opened 2020-01-16
small update of example

PaulSoderlind:
now that quadform is a scalar. Also, Plots.jl has changed how legends should be formatted (row vector).

### DIFF
--- a/docs/examples_literate/portfolio_optimization/portfolio_optimization2.jl
+++ b/docs/examples_literate/portfolio_optimization/portfolio_optimization2.jl
@@ -1,6 +1,6 @@
-# #  Portfolio Optimization - Markowitz Efficient Frontier 
+# #  Portfolio Optimization - Markowitz Efficient Frontier
 #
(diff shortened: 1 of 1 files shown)


ericphanson:
Looks good, thanks for the update!

ericphanson:
Just to say, I’m very happy with how easy it was to review this now that we’ve changed to Literate.jl. Was the new updating process OK for you, @paulsoderlind?

(One difficulty I can anticipate is the need to generate all the examples just to check how it looks, which would be helped by https://github.com/JuliaDocs/Documenter.jl/issues/1188, but maybe there are other difficulties that are easier to improve on our side).

PaulSoderlind:
Yes, this was easy to work with. 

My work flow was: (1) edit the jl file in my text editor, (2) run `Literate.notebook()` on the jl file to generate a notebook; (c) run the notebook to double check that all was well. 

In particular, I didn't generate any of the other examples.

ericphanson:
Ah, that sounds like a good workflow! Glad to hear it was easy. Thanks again for the examples!

PR #365, opened 2020-02-18
Fix `conv` docs; add test of the definition

### LINKED ISSUE #364: Wrong indices in documentation of `conv(h,x)` ?
[Documentation of `conv(h,x)`][1] currently states:

    h∈R^m x∈R^m h∗x∈R^{m+n−1} entry i is given by ∑_{j=1,…,m} h_j ⋅ x_{i−j}

it looks to me there's two things to fix:

1. dimension of `x` is `n`, not `m`: `x∈R^n`
2. for the definition to make sense, one must have `1 <= i-j <= n` for all `j`, hence `1+m <= i <= n+1`, so `h∗x∈R^{n−m+1}` (signs changed w.r.t. to current docs)

[1]: http://www.juliaopt.org/Convex.jl/stable/operations/#Linear-Program-Representable-Functions-1

ericphanson:
Fixes #364


ericphanson:
@riccardomurri There's a preview for the updated docs here: https://www.juliaopt.org/Convex.jl/previews/PR365/operations/. Does that look OK to you?

In the "files changed" tab for this PR you can see the test I added to check that this definition matches what Convex.jl actually does.

riccardomurri:
Looks good to me.  Many thanks for the clarification!

ericphanson:
Okay, great! Thanks for pointing out the issue and the discussion. I’ll merge this now, but the fix will only be in the “dev” docs until we make another release in a little while.

PR #369, opened 2020-02-28
Remove slow `hash` by defining `isequal`

ericphanson:
See https://discourse.julialang.org/t/convex-jl-slow-in-creating-constraints/34775/13 for context. I'm not 100% sure this is completely correct, so I won't merge it until I know it is. Hopefully someone can chime in;  I'll also try and think it through more.


ericphanson:
https://github.com/JuliaOpt/Convex.jl/pull/378#issuecomment-598180132

Issue #78, opened 2015-05-12
Move away from hashing


karanveerm:
Hashing may result in collisions, and since we use hashes to ascertain object equalities, we should probably move away from this.

madeleineudell:
Here's a proposal for how to make our hashing collision-proof.

We've been caching hashes, rather than just defining a hash function recursively and using a dictionary structure to avoid collisions, to avoid traversing a recursive call to hash multiple times in case we've constructed deep nested expressions, eg (+, (x[1], (+, x[2], (+, x[3], ...)))). 

Consider the following procedure. Maintain a global dictionary unique_exprs of (hash, expr) pairs for each expression constructed so far. When a new expression `expr` is formed, compute a hash  `myhash` of it as we do now and check:

    while true
        if haskey(unique_exprs, myhash)
            if unique_exprs[myhash].head == expr.head && length(unique_exprs[myhash].children == length(expr.children) && all(unique_exprs[myhash].children[i].id_hash == expr.children[i].id_hash for i in 1:length(expr.children))
                # we've already formed this expression
                break
            else
                # hash collision --- rehash
                hash = hash(myhash) # there are lots of ways to do this
            end
        else
            # no collision; add it to the global dictionary
            unique_exprs[myhash] = head
            break
        end
    end
    return myhash

You can prove that this works by induction. (Variables and constants have hashes that are unique b/c we use their `object_id`. Assume all hashes of expressions formed so far are unique, form a new expression, and hash it. If there's a collision, and the heads and hashes of children match, then the arguments of the colliding expressions must also be the same by the inductive hypothesis. If there's no collision then all is well.)

mlubin:
If you have a dictionary where the keys are the expressions themselves and not the hashes, the internal logic of the dictionary will handle hash collisions. That could be a simpler solution.

madeleineudell:
Yes, but we'll still need to define `hash` on expressions to put them into
the dictionary, I presume. It would be nice to cache the hashes to avoid
traversing the whole syntax tree every time the same expression is formed.
Is there a good way to do this? The following fails, for example:

    hash(expr) = (myhash = hash((expr.head, map(c->unique_exprs[c],
expr.children)...)); unique_exprs[expr] = hash)

On Tue, May 12, 2015 at 7:38 PM, Miles Lubin <notifications@github.com>
wrote:

> If you have a dictionary where the keys are the expressions themselves and
> not the hashes, the internal logic of the dictionary will handle hash
[...]

PR #378, opened 2020-03-12
Use different hash function for arrays of `AbstractExpr`.

riccardomurri:
This is an update on top of #369.

Testing with my own code, this definition of `hash` is faster than the one on branch "master" by a good margin (3 sec per iteration vs. 7 sec) and still faster than the hashing variant currently used in #369.

### DIFF
--- a/src/expressions.jl
+++ b/src/expressions.jl
@@ -34,16 +34,18 @@ import Base.sign, Base.size, Base.length, Base.lastindex, Base.ndims, Base.conve
 abstract type AbstractExpr end
 abstract type Constraint end
 
+const hashaa_seed = hash(AbstractExpr)
+
 id_hash(x::AbstractExpr) = getfield(x, :id_hash)
 head(x::AbstractExpr) = getfield(x, :head)
-
-const hashaa_seed = hash(AbstractExpr)
+combinehash(x1, x2) = xor(x1*hashaa_seed, x2)
 
 Base.hash(ex::AbstractExpr, h::UInt) = xor(id_hash(ex), hash(head(ex), h))::UInt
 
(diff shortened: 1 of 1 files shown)


ericphanson:
Thanks for the PR and the benchmarking.

I am concerned about the correctness of the `xor` based hashing (which is also a problem in the other PR). We are getting the same failure as examined in https://discourse.julialang.org/t/convex-jl-slow-in-creating-constraints/34775/26 but also two other ones this time. I'm glad our tests are sensitive enough to catch these; they are pretty subtle issues where with hash collisions cause two expressions to be identified as the same, causing correctness problems if the potential collisions actually occur in one of the problems being tested.

I am a bit reluctant to mess with the hashing because of that-- even if we do find a fast version that doesn't cause test failures, we don't know it won't cause correctness problems in other code out in the wild, unless we are sure that the new implementation is suspectible to collisions, and I'm not entirely sure how to check that.

Is the hashing still a bottleneck in your work?

riccardomurri:
> Is the hashing still a bottleneck in your work?

No, definitely not.  I just thought I would save the work done on the hashing somewhere before I forget, but please feel free to ignore it, if it breaks tests.

ericphanson:
> > Is the hashing still a bottleneck in your work?
> 
> No, definitely not. I just thought I would save the work done on the hashing somewhere before I forget, but please feel free to ignore it, if it breaks tests.

Sure, makes sense :). I will close this for now because I now think we might be able to wait for a better solution to avoid hashing altogether, ref https://github.com/JuliaOpt/Convex.jl/issues/78. Or some other solution... I think the internals definitely need a performance redesign, but I haven't yet figured out how that might work.

PR #379, opened 2020-03-13
Allow setting max print width of IDs via `Convex.MAXDIGITS`.

riccardomurri:
Follow-up to https://discourse.julialang.org/t/convex-jl-how-to-print-full-variable-id/35917

### DIFF
--- a/docs/src/advanced.md
+++ b/docs/src/advanced.md
@@ -143,6 +143,11 @@ only the first `MAXWIDTH` of the constraints will be printed. Vertical dots,
 "⋮", will be printed indicating that some constraints were omitted in the
 printing.
 
+A related setting is [`Convex.MAXDIGITS`](@ref), which controls
+printing the internal IDs of atoms: if the string representation of an
+ID is longer than double the value of `MAXDIGITS`, then it is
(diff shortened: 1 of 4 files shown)


ericphanson:
Thanks, looks great! Any ideas on how to test it? Maybe just check the length of the result of `show_id` before changing the setting is small and see that with `MAXDIGITS` really big we get everything?

ericphanson:
A mention in the docs here would be great too: https://github.com/JuliaOpt/Convex.jl/blob/master/docs/src/advanced.md#printing-and-the-tree-structure

As well as adding `Convex.MAXDIGITS` to the `@docs` block there at the end.

riccardomurri:
I've added the docs and tests.  The test are maybe not optimal, in that they use knowledge that `show_id(x) = string(x.id_hash)` but I couldn't imagine another way of doing it.

ericphanson:
Looks great, thanks!

ericphanson:
0.13.2 will be released with this feature once this PR goes through in ~30 minutes https://github.com/JuliaRegistries/General/pull/10934

PR #381, opened 2020-03-21
Fix `add_constraint!`

### LINKED ISSUE #380: `add_constraint!` does not add constraint to problem
It seems that `add_constraint!` is *not* modifying its `problem` argument in-place:

```
julia> using Convex;

julia> x = Variable(100, 100);

julia> p = minimize(norm_1(x));

julia> y = randn(100, 100);

julia> c = (norm2(x-y) < 1);

julia> add_constraint!(p, c)
1-element Array{Constraint,1}:
 <= constraint (convex)
├─ norm2 (convex; positive)
│  └─ + (affine; real)
│     ├─ 100×100 real variable (id: 849…645)
│     └─ 100×100 Array{Float64,2}
└─ 1

julia> p.constraints
0-element Array{Constraint,1}
```

Am I doing something wrong?

ericphanson:
Closes #380 

The definition of this function hasn't really changed since it was introduced in 2014 in https://github.com/JuliaOpt/Convex.jl/commit/9c346819492f4c3bf551845bc1400fc8eda0c139, but I think it worked at that point since at that point, `+` for constraints called `append!`, so it was more of a `+=` than a `+`, and did the right thing for `add_constraints!` (but I guess the wrong thing for a `+` method). It looks like the `+` method was fixed in https://github.com/JuliaOpt/Convex.jl/commit/337bb96c17d4ce95130473cf7547ba2648a4bed0 (also in 2014) to append to an empty array, which then broke `add_constraints!`.

Quite an old bug! This must have been frustrating for some users over the years. Thanks @riccardomurri for reporting it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@
 Manifest.toml
 benchmark/*.json
 test/Project.toml
+dev
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Convex"
 uuid = "f65535da-76fb-5f13-bab9-19810c17039a"
-version = "0.13.2"
+version = "0.13.3"
 
 [deps]
 AbstractTrees = "1520ce14-60c1-5f80-bbc7-55ef81b5835c"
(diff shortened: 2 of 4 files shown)


ericphanson:
The checks failing are due to Travis maintenance, I believe. I'll retry later once that's done.

ericphanson:
Pretty annoying that travis is bugged, but it errors out on the one setup I've tested locally, so I'm going to merge it.

PR #385, opened 2020-04-09
Try allowing the matrix frac test on SCS

ericphanson:
Let us see if the bug persists.


ericphanson:
Very strange segfault in the docs build; can't reproduce locally on MacOS. I'll retry...

ericphanson:
Let's retry now that https://github.com/JuliaOpt/Convex.jl/pull/386 fixed the docs

PR #387, opened 2020-04-29
Use BibTeX highlighting for citation

zdroid:


ericphanson:
Didn’t know GitHub knew bibtex! Thanks for the fix.

PR #396, opened 2020-08-05
Discuss extended formulations in docs

ericphanson:
I'm not 100% sure this is the best place for it, but I wanted it somewhere people would see it fairly quickly, because I think it makes it more clear what exactly Convex.jl is doing and why DCP warnings are important. As I've worked on #393, I've realized the obvious fact that the central/only thing Convex.jl does is automatic application of extended formulations, and I didn't find that to be completely clear from the docs, so I wanted to write about it somewhere central. 

The other reason for adding this is to make the issue with non-DCP problems more clear. I actually haven't seen too many issues where people are solving non-DCP problems and getting confused by the fact that the results are wrong, but I do think it can be confusing why that happens. For example, I thought for awhile that the main problem was it was a nonconvex problem and you are finding a local solution instead; really the issue is you are solving a completely different problem since the reformulation was invalid, and I hope the `abs` example makes that clear.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -3,13 +3,12 @@ Convex.jl - Convex Optimization in Julia
 
 Convex.jl is a Julia package for [Disciplined Convex
 Programming](http://dcp.stanford.edu/) (DCP). Convex.jl makes it easy to
-describe optimization problems in a natural, mathematical syntax, and to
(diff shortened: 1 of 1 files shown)


ericphanson:
@matbesancon would you be up for reviewing this?

PR #402, opened 2020-09-08
Fix `diagm(x)`

### LINKED ISSUE #401: diagm does not work
`diagm` operation does not work: 

**Code:**

```
using Convex
using LinearAlgebra
test_vec = Convex.Variable(n)
D_test_vec = diagm(test_vec)
```

**Output:**
```
MethodError: no method matching diagm(::Variable)
Closest candidates are:
  diagm(!Matched::Pair{var"#s167",var"#s166"} where var"#s166"<:(StaticArrays.StaticArray{Tuple{N},T,1} where T where N) where var"#s167"<:Val...) at C:\Users\shuvo\.julia\packages\StaticArrays\l7lu2\src\linalg.jl:136
  diagm(!Matched::Pair{var"#s826",var"#s825"} where var"#s825"<:(AbstractArray{T,1} where T) where var"#s826"<:Integer...) at D:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.5\LinearAlgebra\src\dense.jl:279
  diagm(!Matched::Integer, !Matched::Integer, !Matched::Pair{var"#s826",var"#s825"} where var"#s825"<:(AbstractArray{T,1} where T) where var"#s826"<:Integer...) at D:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.5\LinearAlgebra\src\dense.jl:280
  ...

Stacktrace:
 [1] top-level scope at In[54]:1
```
**Source of error**
The source of error seems to be line 55 of the file: [https://github.com/jump-dev/Convex.jl/blob/master/src/atoms/affine/diagm.jl](https://github.com/jump-dev/Convex.jl/blob/master/src/atoms/affine/diagm.jl):

```
Diagonal(x::AbstractExpr) = DiagMatrixAtom(x)
```

which should be

```
diagm(x::AbstractExpr) = DiagMatrixAtom(x)
```

ericphanson:
Fixes #401. The `diagm` method accepting a vector was taken away in Julia 1.0 and added back in Julia 1.1, so we should add it back here too.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# Changes in v0.13.6
+
+* fix [#401](https://github.com/jump-dev/Convex.jl/issues/401) by allowing `diagm(x)`.
+* 
 # Changes in v0.13.5
 
 * fix [#398](https://github.com/jump-dev/Convex.jl/issues/398) by allowing `fix!`'d variables in `quadform`.
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Convex"
 uuid = "f65535da-76fb-5f13-bab9-19810c17039a"
-version = "0.13.5"
+version = "0.13.6"
 
 [deps]
 AbstractTrees = "1520ce14-60c1-5f80-bbc7-55ef81b5835c"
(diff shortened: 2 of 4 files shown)


ericphanson:
The one CI failure is a network issue, so I'll merge this

PR #404, opened 2020-09-10
Add `silent_solver` keyword argument and tweak CI

### LINKED ISSUE #403: Expose a way to set `MOI.Silent`
So that there is a generic way to suppress solver output. This is just a matter of adding a keyword argument to `solve!` and conditionally on that keyword, calling `MOI.set(model, MOI.Silent(), true)` on the MOI model during solving. (Maybe we should reuse the `verbose` keyword?)

ericphanson:
Fixes #403 

Not totally sold on the name, but I wanted to distinguish it from `verbose`...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ os:
   - osx
 julia:
   - 1.0
-  - 1.4
+  - 1.5
 if: branch = master OR tag IS present OR type = pull_request
 notifications:
   email: false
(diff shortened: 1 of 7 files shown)


ericphanson:
Unrelatedly to the other contents of this PR, I removed the benchmark stage from Travis (introduced in #321) since I never look at it and it just adds time. I think a better solution down the line would be to use @tkf's https://github.com/tkf/BenchmarkCI.jl which comments in the PR itself with the benchmark results.

PR #405, opened 2020-11-28
fix: hcat would fail with one block; added unary +(Sign) and +(Comple…

dstahlke:
As the title says, without this I sometimes get errors when using hcat with only one item.  Why would I use hcat with only one item?  Because I'm running the solver in a function that can take a variable sized list (corresponding to something like a block matrix) and sometimes there is only one item.

Though this patch reliably fixes the problem, somehow I have struggled to come up with a standalone test case.  Let me know if a test case is important and I can try to isolate this.

### DIFF
--- a/src/dcp.jl
+++ b/src/dcp.jl
@@ -84,6 +84,7 @@ struct ComplexSign <: Sign                end
 +(v::ConvexVexity, w::ConcaveVexity) = NotDcp()
 
 #+(::Convex.Positive, ::Convex.NoSign)
++(s::Sign) = s
 +(s::Positive, t::Positive) = s
 +(s::Negative, t::Negative) = s
 +(s::Positive, t::Negative) = NoSign()
@@ -95,6 +96,7 @@ struct ComplexSign <: Sign                end
 +(t::Negative, s::NoSign) = s+t
 
 # Any sign + ComplexSign = ComplexSign
++(s::ComplexSign) = s
 +(s::ComplexSign, t::ComplexSign) = s
 +(s::Sign, t::ComplexSign) = t
 +(t::ComplexSign, s::Sign) = s+t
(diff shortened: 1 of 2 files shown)


ericphanson:
Thanks! Yeah, a test would be great. Maybe a simple problem with a single-argument hcat like `x = Variable(); y = hcat(x), ...`? I can try to cook one up if you run into any trouble.

dstahlke:
Test cases added.  The reason I was having trouble reproducing the problem was that none of my changes to Convex.jl were taking effect, even if I quit julia.  I deleted anything called "Convex" in ~/.julia and still no effect.  Finally blew away ~/.julia entirely and that fixed it.  Maybe it's because I manually did git clone rather than pkg dev.

ericphanson:
> Test cases added.

Great thanks! This looks great but I just realized our automated testing (CI) isn’t running; we likely ran out of credits (Travis used to not need credits for open source projects until recently). So I’ll switch CI to GitHub actions first; I just put up a PR for that (#406). 

Once that gets merged, you’ll need to rebase on that in order to pick up the new commits and force push to your branch so the automated tests run. (In case you aren’t too familiar with git, like me until recently, the steps should be to update the master branch of your fork of Convex, then checkout your PRs branch again, run `git rebase master`, and then force push to your branch. Let me know if you need any help!).

> The reason I was having trouble reproducing the problem was that none of my changes to Convex.jl were taking effect, even if I quit julia. I deleted anything called "Convex" in ~/.julia and still no effect. Finally blew away ~/.julia entirely and that fixed it. Maybe it's because I manually did git clone rather than pkg dev.

Hm, sometimes I manually clone Convex, say into `~/Convex.jl` but then to add it to current Julia environment, I dev it from the local path, as `] dev ~/Convex.jl`. I think what is important is not how it gets onto your computer but rather how it’s added to your current environment (dev vs add).

ericphanson:
Closed and open to trigger the newly working CI; if it all passes, I'll merge tomorrow. Thanks again!

dstahlke:
Great, thanks!  Let me know if you need anything from me.

PR #406, opened 2020-12-01
Switch CI to github actions

ericphanson:
Following https://github.com/jump-dev/JuMP.jl/pull/2370, but with a few small changes: extra `nightly` action, match our previous Travis tests choice of OS and architecture (although I added windows), ignore readme changes, run docs with `--depwarn=error` since it includes our examples too.

I believe we need to add the `DOCUMENTER_KEY` to our repo secrets. Maybe @odow can help?


ericphanson:
unfortunately GLPK's build step has a deprecation warning on Julia 1.0, meaning `depwarn=error` just immediately errors out. In the interest of getting CI working again asap let's take it out for now (matching what we had with Travis).

odow:
Added the key.

ericphanson:
It works! https://jump.dev/Convex.jl/previews/PR406/

Thanks @odow

Issue #394, opened 2020-08-01
`eigmin` sign issue


ericphanson:
Currently, `sign(eigmin(A)) == Positive()` always (same for `eigmax`). This is wrong if `A` is not PSD. The code has a comment saying it assumes `A` is PSD but that constraint is not actually added anywhere. And the atom makes sense for general Hermitian matrices, which of course can have negative eigenvalues.

I'm not quite sure what the fix is; we could say `NoSign()` instead, but that might break working code that uses `eigmin` for positive operators only. It would be cool to track the PSDness of AbstractExprs like we track the sign, because then we could use that here for the `sign` and elsewhere too (eg trace of psd is positive). This seems related to the operator vexity discussion of https://github.com/jump-dev/Convex.jl/issues/138#issuecomment-609541914.

PR #409, opened 2020-12-16
Fixed nuclearnorm and sumlargesteigs for complex variables

dstahlke:
- nuclearnorm and sumlargesteigs now support complex variables.  This only required making the auxiliary variables complex when the input is complex.
- sumlargesteigs had a condition A ⪰ 0 which was not appropriate.  This would require the input variable to be positive semidefinite.  A unit test shows it works correctly even when A is not.
- I've also included the unit test for my recent hcat fix.  Looks like this wasn't merged into master.

### DIFF
--- a/docs/src/operations.md
+++ b/docs/src/operations.md
@@ -106,13 +106,14 @@ Semidefinite Program Representable Functions
 An optimization problem using these functions can be solved by any SDP
 solver (including SCS and Mosek).
 
-| operation          | description                       | vexity  | slope         | notes                          |
-| ------------------ | --------------------------------- | ------- | ------------- | ------------------------------ |
-| `nuclearnorm(x)`   | sum of singular values of $x$ | convex  | not monotonic | none                           |
-| `operatornorm(x)`  | max of singular values of $x$ | convex  | not monotonic | none                           |
-| `eigmax(x)`     | max eigenvalue of $x$         | convex  | not monotonic | none                           |
-| `eigmin(x)`     | min eigenvalue of $x$         | concave | not monotonic | none                           |
-| `matrixfrac(x, P)` | $x^TP^{-1}x$                  | convex  | not monotonic | IC: P is positive semidefinite |
+| operation              | description                       | vexity  | slope         | notes                          |
+| ------------------     | --------------------------------- | ------- | ------------- | ------------------------------ |
+| `nuclearnorm(x)`       | sum of singular values of $x$     | convex  | not monotonic | none                           |
+| `operatornorm(x)`      | max of singular values of $x$     | convex  | not monotonic | none                           |
+| `eigmax(x)`            | max eigenvalue of $x$             | convex  | not monotonic | none                           |
+| `eigmin(x)`            | min eigenvalue of $x$             | concave | not monotonic | none                           |
+| `matrixfrac(x, P)`     | $x^TP^{-1}x$                      | convex  | not monotonic | IC: P is positive semidefinite |
+| `sumlargesteigs(x, k)` | sum of top $k$ eigenvalues of $x$ | convex  | not monotonic | IC: P symmetric                |
 
 Exponential + SDP representable Functions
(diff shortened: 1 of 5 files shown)


ericphanson (commented):
Nice work!

For `sumlargesteigs` we should update the comment (https://github.com/jump-dev/Convex.jl/pull/409/files#diff-172ad5804a0872ff3dafd9f53ebfbdc6357701d755bf90861049f3008cc87f9fR51) to remove `A >= 0` (I checked that in the reference it's not there, but they do require `A` to be symmetric), and we should add a check in the `SumLargestEigs` constructor that `A` is either symmetric or Hermitian. It would be nice to add a reference that covers the complex case too.

For the nuclear norm and operator norm, it looks like Recht, Fazel, Parillo 2008 only covers the real case, so it would be great to get a reference for the complex case there too. I checked Watrous's TQI (https://cs.uwaterloo.ca/~watrous/TQI/TQI.pdf) but didn't find it (although maybe I missed it).

ericphanson:
Something else to consider for `sumlargesteigs` is that if we don't require `A >= 0`, then returning `Positive()` as the sign isn't right in general: https://github.com/jump-dev/Convex.jl/blob/621c212b092b86bea3a3e14c4468a6120d4591ce/src/atoms/sdp_cone/sumlargesteigs.jl#L30

We actually have this issue for `eigmin` too: #394. For correctness then it seems like we either have to return `NoSign()` or require `A >= 0`, or figure out a way to keep track of the PSDness of the operator.

As a workaround, we could also introduce some ugly functions like `sumeigslargest_psd` which does the current behavior and return `Positive()`, and `sumeigslargest` which returns `NoSign()` (although that would be breaking, so we probably need to do it the other way around).

dstahlke:
I don't think there is a way to detect that A is Hermitian in the SumLargestEigs constructor (note there is currently no Hermitian variable type or DCP attribute).  But A==A' could be added as a constraint.  And added to the documentation.  Also I noticed sumlargesteigs is not listed on https://jump.dev/Convex.jl/stable/operations.  I just assumed this meant it was experimental/undocumented.  I can add it tonight.

I'll take a look through Recht, Fazel, Parillo 2008 to see if the proof applies to the complex case.  My guess as to why a reference is hard to find is that it's mostly quantum folks who care about complex SDP and their matrices are all Hermitian.  I believe that for Hermitian matrices a simpler implementation of nuclear norm is tr(P+M) where X=P-M, 0 ⪯ P, 0 ⪯ M.

As for positivity of sumlargesteigs, it is my opinion that DCP should be expanded with matrix attributes operator-real (Hermitian), operator-positive, and operator-negative.  Functions could be operator-monotone, operator-concave, operator-convex where ⪯ rather than <= is the relevant ordering.  This would probably be a lot of work though.

dstahlke:
Just pushed some updates.  I found the complex nuclear norm in Watrous: he calls it "trace norm".  I updated sumlargesteigs comments and added it to the operations.md doc.  Hermiticity of the argument is enforced automatically because of the semidefinite constraints.

It sounds like sign of sumlargesteigs is a longer-term issue to be resolved along with #394, so I didn't touch that.

ericphanson:
Sorry, I missed the last notification here! I'll review shortly

ericphanson:
Thanks, references, doc changes, tests, this is great! One small thing is I would prefer to keep the `A >= 0` constraint here, or change the sign of the return to `NoSign()`. I agree that it's a bigger issue to add matrix attribues like you mentioned in your previous comment, but removing `A >= 0` without doing so or weakening the `sign` return feels like introducing a bug, as opposed to just not fixing one.

dstahlke:
Oops, I didn't notice that sign was Positive.  I changed it to NoSign.

ericphanson:
Ok thanks! I think I'll fix #394 the same way, and fix a few other issues, and tag a (breaking) 0.14.0 release.

PR #415, opened 2021-01-19
Fix typos in NEWS

ericphanson:


ericphanson:
Ah 😆. The CI will not run because it uses the `paths` spec to run only when certain files are changed, and the NEWS doesn't affect code or tests. But I asked @odow to add branch protection to require these CI checks to pass to be able to merge.

I really just want to disallow pushing to master, to make everything go through a PR with CI, but I'm not sure that's possible. @odow maybe we can just uncheck the "require CI" boxes but leave the branch protected? I think that will just disallow force pushes which is something at least.

odow:
Done. I think this is what we have in JuMP.

Issue #138, opened 2016-05-04
Matrix exp/log/entropy


bschreck:
Do the exp or log function support the matrix versions? i.e. exp or log of the singular values of the matrix?

I'm looking for a way to implement a bregman divergence term like Tr(Xlog(X/A) - X + A), where A is a constant.

madeleineudell:
Interesting. It's pretty easy to give access to any function of the
eigenvalues; we don't currently expose the functionality, because it's a
bit difficult to certify convexity. But if you're ready for adventure,
check out the `schatten` branch, and try the `eig` function: for example,

using Convex

n = 5
# generate a random PSD matrix M
A = randn(n,n)
U,S,V = svd(A)
d = rand(n)
M = U*d*U'

X = Semidefinite(n)
p = maximize(sum(log(eig(X))), M - X in :SDP)
solve!(p)

@show p.optval - sum(log(d))

Any convex or concave function of the eigenvalues should work, including KL
divergence, which I think is what you want. But exercise caution: I'm not
yet willing to certify correctness of the formulation. Under the hood, if
you try to minimize f(eig(A)), it will solve the problem

minimize  f(d)
subject to U upper triangular
                d == diag(U)
                A :in :PSD
                [diagm(d) U; U' A] :in :PSD

On Tue, May 3, 2016 at 6:12 PM, Ben Schreck <notifications@github.com>
wrote:

> Do the exp or log function support the matrix versions? i.e. exp or log of
> the singular values of the matrix?
>
> I'm looking for a way to implement a bregman divergence term like
> Tr(Xlog(X/A) - X + A), where A is a constant.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/JuliaOpt/Convex.jl/issues/138>
>



-- 
Madeleine Udell
Postdoctoral Fellow at the Center for the Mathematics of Information
California Institute of Technology
*https://courses2.cit.cornell.edu/mru8
<https://courses2.cit.cornell.edu/mru8>*
(415) 729-4115

bschreck:
Thanks I'll check it out! I've actually never used Julia before (I came here because CVXPY doesn't have support for this) so this will be fun.

[...]

PR #418, opened 2021-01-24
ported cvxquad: matrix log, geometric mean, quantum entropy

dstahlke:
This is the port of the Fawzi-Saunderson cvxquad, providing quantum entropy functions.  I think it's mostly ready to go, but I have a few questions.

I added `evaluate(x::Value) = output(x)` in src/constant.jl.  Is that okay?  It allows for uniform syntax to accept either `Constant` or `AbstractMatrix` as arguments to a function.

I have functions named, e.g., `quantum_relative_entropy`, `lieb_ando`, `trace_mpower`, `trace_logm`.  These names are copied from cvxquad.  But it seems in Julia it's preferred to not have the underscores.  What do you think?

Please take a look at `vexity` in geom_mean_hypocone.jl and geom_mean_epicone.jl.  I'm implementing these as constraints to support the syntax `T in GeomMeanHypoCone(A, B, t)`.  But I don't understand what vexity is supposed to mean for a constraint, or what vexity means for a variable.  For variables, I could imagine AffineVexity or ConvexVexity to mean the variable is constrained to an affine or convex set, but I don't know if that's correct.  I sort of copied `vexity(c::GtConstraint)` but I doubt I got it right.

Some background: `T in GeomMeanHypoCone(A, B, t)` implements the constraint T ⪯ A #_t B, for t in [0, 1].  The function A #_t B is operator concave in A and B.  But I don't know how that translates to a vexity of `GeomMeanHypoConeConstraint`, or how `vexity(constraint::GeomMeanHypoConeConstraint)` is supposed to consider the vexity of A, B, and T.  `vexity(c::GtConstraint)` does `vex = -vexity(c.lhs) + (vexity(c.rhs))` but I don't know why.

I'd like to note that this seems to run much slower than the timings quoted in `Efficient optimization of the quantum relative entropy`.  I guess SCS is slower than Matlab's solver.  It's still workable though.

(we had discussed this in #138)


ericphanson:
Wow, this looks great! It’s a pretty big chunk of code so it might take me some time to review, but I’ll try to get started soon.

src/atoms/sdp_cone/geom_mean_epicone.jl
```diff
@@ -0,0 +1,120 @@
...
+    GeomMeanEpiConeConstraint(T::Value, cone::GeomMeanEpiCone) = GeomMeanEpiConeConstraint(Constant(T), cone)
+end
+
+in(T, cone::GeomMeanEpiCone) = GeomMeanEpiConeConstraint(T, cone)
+
+function AbstractTrees.children(constraint::GeomMeanEpiConeConstraint)
+    return (constraint.T, constraint.cone.A, constraint.cone.B, "t=$(constraint.cone.t)")
+end
+
+# FIXME what is the meaning of the vexity of a variable? of a constraint?  Is this correct?
```
ericphanson, line 71: sorry for the delay. So the meaning of vexity of a variable is just:

* ConstVexity iff it's `fix!`'d and should be treated as a constant
* AffineVexity otherwise

That is, we treat a variable `x` like the function `f(x) = x` for the purposes of deciding the vexity (and like `f(x) = c` if it's a constant).

The vexity of a `constraint` is just to determine is the constraint of the form `f(x) <= 0` for convex `f` (where `<=` might be with respect to a general cone, not necessarily just in R or C).

src/atoms/sdp_cone/geom_mean_hypocone.jl
```diff
@@ -0,0 +1,169 @@
...
+struct GeomMeanHypoCone
+    A::AbstractExpr
+    B::AbstractExpr
+    t::Rational
+    size::Tuple{Int, Int}
+    fullhyp::Bool
+
+    function GeomMeanHypoCone(A::AbstractExpr, B::AbstractExpr, t::Rational, fullhyp::Bool=true)
+        if size(A) != size(B)
+            error("A and B must be the same size")
```
ericphanson, line 38: Would be great to get some `@test_throws` tests to check these error cases! (the uncovered lines are marked by codecov)

ericphanson:
```suggestion
            throw(ArgumentError("`A` and `B` must be the same size"))
```

src/atoms/sdp_cone/geom_mean_epicone.jl
ericphanson: so to answer the question of "is this correct", I'm not 100% sure... the question is if `A` and `B` were functions of some underlying vector of variables `x` what conditions are needed on `A` and `B` such that
```
A^{1/2} (A^{-1/2} B A^{-1/2})^t A^{1/2}
```
is a convex function of `x`?

i.e. a possible "condition" would be `A` and `B` must be convex functions (just as an example, I don't know what the conditions are here).

ericphanson (commented):
sorry @dstahlke, this deserves a very thorough review but I've mostly been working on other stuff recently... I left a few comments to answer your questions as a start and will try to continue to engage and help push this through.

One other thing we should do is add the license of the HF code to the Convex.jl LICENSE file, and say something like "the functions X, Y, and Z were ported from CVXQUAD [link], which is availabe under the following license...

> license text...
> license_text...
"

Does that make sense?

chriscoey:
@dstahlke this is very impressive. I have been wanting to compare some new techniques for solving the sorts of problems cvxquad solves, in Hypatia (eg see these examples [1](https://github.com/chriscoey/Hypatia.jl/blob/master/examples/classicalquantum/JuMP.jl), [2](https://github.com/chriscoey/Hypatia.jl/blob/master/examples/relentrentanglement/JuMP.jl), [3](https://github.com/chriscoey/Hypatia.jl/blob/master/examples/entanglementassisted/JuMP.jl), [4](https://github.com/chriscoey/Hypatia.jl/blob/master/examples/nearestcorrelation/JuMP.jl)) and I've been worried I would have to either use Python or reimplement the extended formulations myself. But now that I've seen this PR, it looks like I don't have to! 
@ericphanson @dstahlke is there anything I can do to help get this merged?

ericphanson:
It would be awesome to get this in, I unfortunately haven't had much time to review it more thoroughly and probably won't for the forseeable future either. I think there are some outstanding review items to address and it would be great if anyone could help review it more.

dstahlke:
Sorry, I got busy and this fell off my radar.  I'll address Eric's comments today.

ericphanson:
Thanks for updating it! Looks really great.

I saw I never answered some of the questions in the original post:

> I added evaluate(x::Value) = output(x) in src/constant.jl. Is that okay? It allows for uniform syntax to accept either Constant or AbstractMatrix as arguments to a function.

Yeah, that seems OK to me.

> I have functions named, e.g., quantum_relative_entropy, lieb_ando, trace_mpower, trace_logm. These names are copied from cvxquad. But it seems in Julia it's preferred to not have the underscores. What do you think?

I think the underscores are good here, it would be a bit unreadable otherwise. The [JuMP style guide](https://jump.dev/JuMP.jl/stable/developers/style/#Use-of-underscores-within-names) says to always use underscores as well. I don't think Convex is actually following any style guide currently though.

> I'd like to note that this seems to run much slower than the timings quoted in Efficient optimization of the quantum relative entropy. I guess SCS is slower than Matlab's solver. It's still workable though.

Ah ok. Yeah, hard to say if it's the solver or some issue in Convex.jl without direct benchmarks, but I think this is very much worth merging even if there are performance issues, because having working+tested code is a huge step forward already.

---

I have not thoroughly reviewed everything but what I saw looked good to me. Do you have any outstanding concerns @dstahlke?

Besides any further code review, I think we should probably wait on https://github.com/hfawzi/cvxquad/issues/7, or ask @hfawzi for permission to port the code in this case (since that would suffice I believe even if there is not a license outlining how people may use the code in general).

dstahlke:
I'm going to convert relative_entropy_epicone to be a constraint rather than a function, to match GeomMeanEpiCone.  So the syntax will be `T in RelativeEntropyEpicone(...)` rather than `T = relative_entropy_epicone(...)`.  I think this will help remind the user that T is only bounded from below.  And actually, it may be better to use \succeq and \preceq rather than `in` for all these cones to make sure the user is aware which ones bound above vs. below.  Any thoughts?  I think actually a lot of functions in Convex.jl are only bounded above/below, even though they are written `y = f(x)` when used.

Other than that, all that remains is some more coverage and license/permission from Hamza.  I've been pushing code in increments in order to get coverage updates.

dstahlke:
I think it's all good to go now.  Let me know if you need me to rebase it all into one commit.

ericphanson:
Awesome! Looks great. No rebasing needed, I think I'll squash-merge it.

I took it for a spin and confirmed some things I knew were true 😄 .

Max-entropy state is the majorization-minimizer:
```julia
julia> using Convex, SCS

julia> using MajorizationExtrema # https://github.com/ericphanson/MajorizationExtrema.jl

julia> σ = randdm(4)
4×4 Hermitian{ComplexF64, Matrix{ComplexF64}}:
   0.221816+0.0im        -0.0353134+0.0163144im   -0.0128095+0.0155548im    0.0135785+0.0357286im
 -0.0353134-0.0163144im      0.2402+0.0im          0.0189347+0.00939113im  -0.0390478-0.0315439im
 -0.0128095-0.0155548im   0.0189347-0.00939113im    0.212807+0.0im          0.0174911+0.0255289im
  0.0135785-0.0357286im  -0.0390478+0.0315439im    0.0174911-0.0255289im     0.325177+0.0im

julia> ρ = HermitianSemidefinite(4)
Variable
size: (4, 4)
sign: complex
vexity: affine
id: 131…583

julia> add_constraint!(ρ, tr(ρ) == 1)
2-element Vector{Constraint}:
 sdp constraint (affine)
└─ transpose (affine; real)
   └─ hcat (affine; real)
      ├─ transpose (affine; real)
      │  └─ …
      └─ transpose (affine; real)
         └─ …
 == constraint (affine)
├─ sum (affine; complex)
│  └─ diag (affine; complex)
│     └─ 4×4 complex variable (id: 131…583)
└─ 1

julia> problem = maximize(quantum_entropy(ρ), .5*nuclearnorm(ρ - σ) <= 0.1)
maximize
└─ quantum_entropy (concave; positive)
   └─ 4×4 complex variable (id: 131…583)
subject to
└─ <= constraint (convex)
   ├─ * (convex; positive)
   │  ├─ 0.5
   │  └─ nuclearnorm (convex; positive)
   │     └─ …
   └─ 0.1

julia> solve!(problem, SCS.Optimizer; silent_solver=true)

julia> tracedist(evaluate(ρ), majmin(σ, 0.1))
9.992598351190096e-7

```

Maximize conditional entropy near maximally-entangled state:
```julia
julia> conditional_entropy(ρ) = 2 - quantum_relative_entropy(ρ, kron(I(2)/2, partialtrace(ρ, 2, [2, 2]))) # assuming two qubits
conditional_entropy (generic function with 1 method)

julia> conditional_entropy(ρ)
+ (concave; real)
├─ 2
└─ - (concave; negative)
   └─ quantum_relative_entropy (convex; positive)
      ├─ 4×4 complex variable (id: 131…583)
      └─ transpose (affine; complex)
         └─ …

julia> problem = maximize(conditional_entropy(ρ), .5*nuclearnorm(ρ-σ) <= 0.1)
maximize
└─ + (concave; real)
   ├─ 2
   └─ - (concave; negative)
      └─ quantum_relative_entropy (convex; positive)
         ├─ …
         └─ …
subject to
└─ <= constraint (convex)
   ├─ * (convex; positive)
   │  ├─ 0.5
   │  └─ nuclearnorm (convex; positive)
   │     └─ …
   └─ 0.1

julia> solve!(problem, SCS.Optimizer; silent_solver=true)

julia> evaluate(ρ)
4×4 Matrix{ComplexF64}:
    0.466667+0.0im   2.44558e-10+0.0im  -9.81124e-10+0.0im      0.433333+0.0im
 3.88853e-10+0.0im      0.033332+0.0im  -1.95602e-12+0.0im   -1.94878e-9+0.0im
  5.23546e-9+0.0im  -4.77919e-13+0.0im     0.0333339+0.0im  -5.22583e-10+0.0im
    0.433333+0.0im    4.18306e-9+0.0im  -9.12328e-10+0.0im      0.466667+0.0im

julia> tracedist(evaluate(ρ), σ*0.9 + (0.1)*(I(4) - σ)/3)
1.259008819890921e-6
````

ericphanson (approved):
I haven't checked every detail but the tests are extensive and look correct, and the implementation looks good too.

dstahlke:
Thank you for the review, and for teaching me something new about entropy!

PR #420, opened 2021-02-15
lasso, ridge, elastic net example

PaulSoderlind:
This PR contains a csv file (in aux_files) and a jl file for examples of lasso, ridge and elastic net regressions.


ericphanson:
The build logs are reporting some issues; I think you need to add some of the packages to the Project.toml in the docs folder: https://github.com/jump-dev/Convex.jl/pull/420/checks?check_run_id=1903248459#step:5:212

(sorry it’s not the easiest to read)

PaulSoderlind:
I guess it's LaTeXStrings that's missing. However, I could easily skip the fancy $\gamma$ in the plot. Maybe better to not add another dependency, or what do think?

ericphanson:
I think it’s fine to add, it’s a pretty light dependency and nicer plots is always better :)

PaulSoderlind:
I added LaTeXStrings, but I am not sure what is meant by this (from the checks). Do I need to add `MathOptInterface` to the Doc\Toml too? 

(Sorry for bothering you with this, but I am far from a GitHub expert.)

```
Warning: failed to run `@example` block in src/examples/general_examples/lasso_regression.md:17-22
│ ```@example lasso_regression
│ using DelimitedFiles, LinearAlgebra, Statistics, Plots, LaTeXStrings, Convex, SCS
│ 
│ import MathOptInterface
│ const MOI = MathOptInterface
│ ```
│   c.value =
│    ArgumentError: Package MathOptInterface not found in current path:
│    - Run `import Pkg; Pkg.add("MathOptInterface")` to install the MathOptInterface package.
│    
└ @ Documenter.Expanders ~/.julia/packages/Documenter/lul8Y/src/Expanders.jl:563
```

ericphanson:
Yeah, you do. Convex uses it internally but it needs to be in the docs environment too

docs/examples_literate/general_examples/lasso_regression.jl
```diff
@@ -0,0 +1,154 @@
...
+
+# # Redo the Lasso Regression with Different Gamma Values
+#
+#
+# We now loop over $\gamma$ values.
+#
+# Remark: it would be quicker to put this loop inside the `LassoEN()` function so as to not recreate `L1`-`L4`.
+
+nγ = 101
+γM = range(0,600,length=nγ)             #different γ values
```
ericphanson, line 96: Julia 1.0 doesn't know about the `range` method, so we need to use the older syntax:
```suggestion
γM = range(0; stop=600, length=nγ)             #different γ values
```

docs/examples_literate/general_examples/lasso_regression.jl
```diff
@@ -0,0 +1,154 @@
...
+           label = permutedims(xNames),
+           size = (600,400) )
+display(p1)
+
+# # Ridge Regression
+#
+# We use the same function to do a ridge regression. Alternatively, do `b = inv(X'X + λ*I)*X'Y`.
+
+nλ = 101
+λM = range(0,3000,length=nλ)
```
ericphanson, line 117:
```suggestion
λM = range(0; stop=3000, length=nλ)
```

ericphanson:
Logs look good! (At least for your example, looks like we've got some other issues in there).

Thanks @PaulSoderlind!

PR #422, opened 2021-02-15
Update docs.yml

ericphanson:
In @PaulSoderlind's nice new example https://jump.dev/Convex.jl/dev/examples/general_examples/lasso_regression/ the plots aren't rendering correctly. I'm taking a guess that updating the Julia version might get us newer package versions that solve this (since AFAIK new versions of GR don't need latex since they bundle their own tex renderer).

We had been using Julia 1.0 to help ensure the examples run on the LTS. However 1.0 is very old and it might be holding us back here, so maybe it's worth bumping the version to 1.5.


ericphanson:
I'm not seeing any plots nor errors in the preview https://jump.dev/Convex.jl/previews/PR422/examples/general_examples/lasso_regression/. Huh, @PaulSoderlind any ideas? We might have to just drop LaTeXStrings after all

PaulSoderlind:
I believe that GR (the Plots default) incorporated LaTeX capabilities around version 49 or so, so Julia 1.0.5 might be too old.

Yes, I am happy to drop the very few latex strings in the plots (on lines 107, 128, and 147). 

How do we do this?

ericphanson:
I've pushed a commit to remove LaTeXStrings and went back down to Julia 1.0, since it didn't seem like julia 1.5 fixed things. I used unicode instead of latex, hopefully that will work OK (though I'm not totally sure..)

ericphanson:
Interesting, what seems to have done the trick is removing the `display` statements. I'm really not sure why that was a problem, but since this is working now I'll go ahead and merge it. And the unicode shows up OK too.

PaulSoderlind:
Thanks. That was good detective work.

PR #427, opened 2021-03-13
fix bad conversion

### LINKED ISSUE #426: BigFloats are only returned for objective functions, not variable values
If I try the example of optimization using BigFloats:

~~~
using Convex, SDPAFamily
y = Semidefinite(3)
p = maximize(eigmin(y), tr(y) <= sqrt(big(2)); numeric_type = BigFloat)
solve!(p, () -> SDPAFamily.Optimizer(presolve=true))
~~~

The values of the computed variables `y` are not returned as BigFloats

~~~
 typeof(y.value), typeof(p.optval)
(Array{Float64,2}, BigFloat)
~~~

The MOI model does contain the values as BigFloats, and it would be nice to get them:

~~~
using MathOptInterface; const MOI = MathOptInterface
typeof(MOI.get(p.model, MOI.VariablePrimal(), MOI.get(p.model, MOI.ListOfVariableIndices())))
Array{BigFloat,1}
~~~

ericphanson:
Fix #426 

I am not sure why we had the previous method, and I'm not totally sure this is the same (in particular, I'm not sure when `x.value` references the same array as the input to `set_value!(x, v)` with convert vs when it makes a copy). But it's definitely wrong to hardcode `Float64`s or `Complex{Float64}`s here. I'll think about this more tomorrow.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,10 @@
+# Changes in v0.14.3
+
+* fixed a bug where the values of variables were being converted to `Float64` even if the problem was solved in high precision. ([#427](https://github.com/jump-dev/Convex.jl/pull/427)).
+
 # Changes in v0.14.2
 
-* added lasso, ridge, and elastic net regression examples ([420](https://github.com/jump-dev/Convex.jl/pull/420)). Thanks to @PaulSoderlind!
+* added lasso, ridge, and elastic net regression examples ([#420](https://github.com/jump-dev/Convex.jl/pull/420)). Thanks to @PaulSoderlind!
 
 # Changes in v0.14.1
 
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Convex"
 uuid = "f65535da-76fb-5f13-bab9-19810c17039a"
-version = "0.14.3"
+version = "0.14.4"
 
 [deps]
 AbstractTrees = "1520ce14-60c1-5f80-bbc7-55ef81b5835c"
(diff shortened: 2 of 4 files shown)


ericphanson:
From some playing around, e.g.
```julia
julia> A = rand(4,4)
4×4 Matrix{Float64}:
 0.625895   0.556916   0.960171  0.47983
 0.0429329  0.0510638  0.635858  0.00855361
 0.4908     0.4517     0.693719  0.179855
 0.604191   0.500587   0.140534  0.214743

julia> B = convert(Matrix{Float64}, A)
4×4 Matrix{Float64}:
 0.625895   0.556916   0.960171  0.47983
 0.0429329  0.0510638  0.635858  0.00855361
 0.4908     0.4517     0.693719  0.179855
 0.604191   0.500587   0.140534  0.214743

julia> A === B
true

julia> A = rand(Float32, 4, 4)
4×4 Matrix{Float32}:
 0.254641  0.0674895  0.518716  0.567437
 0.183982  0.442368   0.560556  0.546095
 0.658679  0.157236   0.499382  0.334977
 0.728826  0.123285   0.346908  0.91826

julia> B = convert(Matrix{Float64}, A)
4×4 Matrix{Float64}:
 0.254641  0.0674895  0.518716  0.567437
 0.183982  0.442368   0.560556  0.546095
 0.658679  0.157236   0.499382  0.334977
 0.728826  0.123285   0.346908  0.91826

julia> A === B
false
```
It seems that `convert` copies if it needs to change the type, and but does not copy if it's already the type you are converting to. Thus, I think the most conservative fix would be to make a copy unless the input is of the types listed in the code here (i.e. an explicit `Matrix{Float64}` or such).

I'm not totally sure that's the most sensible behavior. I think it probably is a good idea for the value to be complex if the variable is complex (I suspect there could be strange bugs lurking otherwise), but otherwise I don't think it's essential to make a copy.

PR #428, opened 2021-03-14
Fix deprecated setting in Convex's tests for GLPK

ericphanson:


test/runtests.jl
```diff
@@ -56,7 +56,7 @@ end
 
     @testset "GLPK" begin
         run_tests(; exclude=[r"exp", r"sdp", r"socp"]) do p
-            solve!(p, () -> GLPK.Optimizer(msg_lev = GLPK.OFF))
+            solve!(p, () -> GLPK.Optimizer(msg_lev = GLPK.GLP_MSG_OFF))
```
odow, line 59: You probably need something like the following.
```Julia
function optimizer()
    model = GLPK.Optimizer()
    MOI.set(model, MOI.Silent(), true)
    return model
end

solve!(p, optimizer)
```

odow (approved):

ericphanson:
Ok, I used Convex's `silent_solver` keyword argument which basically does the same thing (well, it calls `MOI.set(model, MOI.Silent(), true)` on the populated model, not the optimizer, but I think it should work the same). Too bad keyword arguments for settings are deprecated, that was convenient.

odow:
You can also use `MOI.OptimizerWithAttributes(GLPK.Optimizer, MOI.Silent() => true)`. We removed kwargs because they don't separate the options from the constructor well-enough, and we're trying to push towards a standardized way of declaring options amongst solvers.

ericphanson:
Ok, thanks for the explanation! I probably need to update SDPAFamily at some point then.

PR #429, opened 2021-03-14
allow `sumlargest(x,k)`, `sumsmallest(x,k)`, and `sumlargesteigs(x,k)` for `k=0` 

### LINKED ISSUE #423: `sumsmallest`/`sumlargest` behavior for k=0
I ran into this when writing a program for robust optimization, which involves robustness with respect to removing the k-largest elements. I wanted to check that the result of solving the original non-robust formulation was equivalent to the result of solving the robust reformulation (with k=0), but when solving the program with constraints of the form:
```... + sumlargest(A, k) + ...```

I ran into this error:
```sumlargest and sumsmallest only support positive values of k```

Is there any reason that `sumlargest` and `sumsmallest` can't behave as 0 for k=0? There is some simplicity and convenience in this kind of behavior... 🙂

ericphanson:
fix #423

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,10 +1,14 @@
-# Changes in v0.14.4
+# Changes in v0.14.5
 
 * allow `sumlargest(x,k)`, `sumsmallest(x,k)`, and `sumlargesteigs(x,k)` for `k=0` (simply returns `Constant(0)`). ([#429](https://github.com/jump-dev/Convex.jl/pull/429)).
 
-# Changes in v0.14.3
+# Changes in v0.14.4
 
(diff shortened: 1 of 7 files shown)


ericphanson:
@odow would it be possible to check the "automerge" box in the settings? https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/automatically-merging-a-pull-request

I think that would let me do "merge when CI passes" here, which would be convenient.

odow:
image

ericphanson:
Awesome, thanks! I don't see the option on this PR, but presumably on the next one it'll work :D

PR #431, opened 2021-03-27
Use MOI.instantiate to create a new optimizer

odow:
I didn't test this. I just noticed from a post on dicourse: https://discourse.julialang.org/t/convex-jl-set-some-configurations-e-g-maximum-iteration-number-etc/57724


ericphanson:
Thanks, I didn’t know about`instantiate` (but I haven’t really been following MOI development much anymore). I guess it’s fine to remove the old error if there are other possibilities now (and it’s been awhile since the transition to MOI). Could you fix the failing test?

odow:
@ericphanson fixed.

PR #433, opened 2021-03-28
[docs] mention MOI.OptimizerWithAttributess

odow:


ericphanson:
Thanks! I wonder, should we export MOI?

docs/src/solvers.md
```diff
@@ -45,39 +45,35 @@ using GLPK
...
 ```julia
-using ECOS
-solve!(p, () -> ECOS.Optimizer(maxit=10000))
-using SCS
-solve!(p, () -> SCS.Optimizer(max_iters=10000))
+using Convex, SCS
+const MOI = Convex.MOI
+
+opt = MOI.OptimizerWithAttributes(SCS.Optimizer, MOI.Silent() => false)
+solve!(p, opt)
```
ericphanson, line 67: we could also use `solve!(p, opt; silent_solver=true)` here

PR #443, opened 2021-06-05
Fix installation process steps

tcveatch:
I had to do this to avoid the error message
  Pkg.update ERROR: UndefVarError: Pkg not defined
which was resolved by 
  julia> using Pkg
beforehand.


ericphanson:
Thanks!

Issue #447, opened 2021-07-18
Failing to evaluate matrix expressions 


hurak:
The function `evaluate` seems to fail when evaluating matrix expressions. Below is a minimum working example demonstrating an eigenvalue minimization for an affine matrix function of a vector argument: 

```
minimize   λ
subject to   A(x)⪯λ, where x is a vector.
```

Preparing the problem description in Convex.jl:
```julia
using Convex
using SCS

a₀ = [1.0 2.0; 2.0 3.0]
a₁ = [1.0 0.0; 0.0 -1.0]
a₂ = [4.0 5.0; 5.0 -6.0]

x = Variable(2)
λ = Variable()

A = a₀ + x[1]*a₁ + x[2]*a₂

problem = minimize(λ,A⪯λ)
```
Solving the problem:
```julia
julia> solve!(problem, SCS.Optimizer)
----------------------------------------------------------------------------
    SCS v2.1.3 - Splitting Conic Solver
    (c) Brendan O'Donoghue, Stanford University, 2012
----------------------------------------------------------------------------
Lin-sys: sparse-indirect, nnz in A = 10, CG tol ~ 1/iter^(2.00)
eps = 1.00e-05, alpha = 1.50, max_iters = 5000, normalize = 1, scale = 1.00
acceleration_lookback = 10, rho_x = 1.00e-03
Variables n = 4, constraints m = 5
Cones:  primal zero / dual free vars: 2
    sd vars: 3, sd blks: 1
Setup time: 5.91e-05s
----------------------------------------------------------------------------
 Iter | pri res | dua res | rel gap | pri obj | dua obj | kap/tau | time (s)
----------------------------------------------------------------------------
     0| 7.17e+18  8.71e+18  1.00e+00 -2.56e+19  3.48e+19  2.38e+19  3.75e-05 
    40| 2.10e-09  2.96e-09  2.08e-10  2.00e+00  2.00e+00  8.65e-18  2.06e-04 
----------------------------------------------------------------------------
Status: Solved
Timing: Solve time: 2.11e-04s
    Lin-sys: avg # CG iterations: 1.78, avg solve time: 3.58e-07s
    Cones: avg projection time: 5.61e-08s
    Acceleration: avg step time: 2.76e-06s
----------------------------------------------------------------------------
Error metrics:
dist(s, K) = 4.7016e-16, dist(y, K*) = 0.0000e+00, s'y/|s||y| = -5.4509e-01
[...]

PR #448, opened 2021-07-19
Fixing the issue #447 and updating the examples in docstrings for run_tests() and benchmark_suite()

hurak:
Fixing the issue #447 (simplifying a one-element vector or matrix to a scalar). In addition, updating the syntax for choosing the optimizers in the docstring examples for run_tests() and benchmark_suite().

### DIFF
--- a/src/atoms/affine/index.jl
+++ b/src/atoms/affine/index.jl
@@ -40,9 +40,9 @@ end
 
 function evaluate(x::IndexAtom)
     if x.inds === nothing
-        return getindex(evaluate(x.children[1]), x.rows, x.cols)
+        return output(getindex(evaluate(x.children[1]), x.rows, x.cols)) # output() needed to convert a 1-element vector to a scalar, see the link to the Issue below.
     else
-        return getindex(evaluate(x.children[1]), x.inds)
+        return output(getindex(evaluate(x.children[1]), x.inds)) # output() needed to convert a 1-element vector to a scalar. In response to https://github.com/jump-dev/Convex.jl/issues/447.
     end
 end
 
--- a/src/atoms/affine/index.jl
+++ b/src/atoms/affine/index.jl
@@ -39,11 +39,13 @@ function curvature(x::IndexAtom)
 end
 
 function evaluate(x::IndexAtom)
-    if x.inds === nothing
-        return output(getindex(evaluate(x.children[1]), x.rows, x.cols)) # output() needed to convert a 1-element vector to a scalar, see the link to the Issue below.
+    result = if x.inds === nothing
+        getindex(evaluate(x.children[1]), x.rows, x.cols)
     else
-        return output(getindex(evaluate(x.children[1]), x.inds)) # output() needed to convert a 1-element vector to a scalar. In response to https://github.com/jump-dev/Convex.jl/issues/447.
+        getindex(evaluate(x.children[1]), x.inds)
     end
+    # `output` needed to convert a 1-element array to a scalar (https://github.com/jump-dev/Convex.jl/issues/447)
+    return output(result)
 end
 
 function conic_form!(x::IndexAtom, unique_conic_forms::UniqueConicForms)
--- a/src/problem_depot/problem_depot.jl
+++ b/src/problem_depot/problem_depot.jl
@@ -21,9 +21,9 @@ eye(n) = Matrix{Float64}(I, n, n)
 
 A "depot" of Convex.jl problems, subdivided into categories.
 Each problem is stored as a function with the signature
-    
+
     f(handle_problem!, ::Val{test}, atol, rtol, ::Type{T}) where {T, test}
-    
+
 where `handle_problem!` specifies what to do with the `Problem` instance
 (e.g., `solve!` it with a chosen solver), an option `test` to choose
 whether or not to test the values (assuming it has been solved),
@@ -45,10 +45,10 @@ const PROBLEMS = Dict{String, Dict{String, Function}}()
 
 """
     foreach_problem(apply::Function, [class::String],
-        problems::Union{Nothing, Vector{String}, Vector{Regex}} = nothing; 
+        problems::Union{Nothing, Vector{String}, Vector{Regex}} = nothing;
         exclude::Vector{Regex} = Regex[])
 
-Provides a convience method for iterating over problems in [`PROBLEMS`](@ref). 
+Provides a convience method for iterating over problems in [`PROBLEMS`](@ref).
 For each problem in [`PROBLEMS`](@ref), apply the function `apply`, which
 takes two arguments: the name of the function associated to the problem,
 and the function associated to the problem itself.
@@ -58,8 +58,8 @@ problems (`class` should satsify `class ∈ keys(PROBLEMS)`), and pass third
 argument `problems` to only allow certain problems (specified by exact names or
 regex). Use the `exclude` keyword argument to exclude problems by regex.
 """
-function foreach_problem(   apply::Function, 
-                            problems::Union{Nothing, Vector{String}, Vector{Regex}} = nothing; 
+function foreach_problem(   apply::Function,
+                            problems::Union{Nothing, Vector{String}, Vector{Regex}} = nothing;
                             exclude::Vector{Regex} = Regex[])
     for class in keys(PROBLEMS)
         any(occursin.(exclude, Ref(class))) && continue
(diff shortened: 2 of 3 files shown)


src/problem_depot/problems/sdp.jl
```diff
@@ -1361,3 +1361,35 @@ end
...
+    A₀ = [1.0 2.0; 2.0 3.0]
+    A₁ = [1.0 0.0; 0.0 -1.0]
+    A₂ = [4.0 5.0; 5.0 -6.0]
+
+    x = Variable(2)
+    λ = Variable()
+
+    A = A₀ + A₁*x[1] + A₂*x[2]
+
+    p = minimize(λ,A⪯λ)
```
ericphanson, line 1386: This allows the problem to be used with different types (eg some solvers can handle BigFloats)
```suggestion
    p = minimize(λ,A⪯λ; numeric_type=T)
```

src/problem_depot/problem_depot.jl
```diff
@@ -151,12 +151,12 @@ MathOptInterface model, but not for the actual problem data.
 
 ```julia
 benchmark_suite(exclude=[r"mip"]) do p
-    solve!(p, SCSSolver(verbose=0))
+    solve!(p, () -> SCS.Optimizer(verbose=0))
```
ericphanson, line 154:
```suggestion
    solve!(p, SCS.Optimizer; silent_solver=true)
```

src/problem_depot/problem_depot.jl
```diff
@@ -105,12 +105,12 @@ MathOptInterface model, but not for the actual problem data.
 
 ```julia
 run_tests(exclude=[r"mip"]) do p
-    solve!(p, SCSSolver(verbose=0))
+    solve!(p, () -> SCS.Optimizer(verbose=0))
```
ericphanson, line 108: I would write these like
```suggestion
    solve!(p, SCS.Optimizer; silent_solver=true)
```
these days, since then it's easier to swap between solvers. This uses the `MOI.Silent()` attribute instead of solver-specific ones.

src/problem_depot/problems/sdp.jl
```diff
@@ -1361,3 +1361,35 @@ end
         end
     end
 end
+
+@add_problem sdp function sdp_min_maxeig_canon_lmi(handle_problem!, ::Val{test}, atol, rtol, ::Type{T}) where {T, test}
+    # Minimize the maximum eigenvalue of an affine matrix function of a vector argument. Formulated
+    # as a linear optimization subject to a canonical LMI (SDP) constraint:
+    #
+    # minimize      λ
+    # subject to    A(x)⪯λ
```
ericphanson, line 1370:
```suggestion
    # subject to    A(x) ⪯ λ*I
```

src/problem_depot/problems/sdp.jl
```diff
@@ -1361,3 +1361,35 @@ end
...
+    A₀ = [1.0 2.0; 2.0 3.0]
+    A₁ = [1.0 0.0; 0.0 -1.0]
+    A₂ = [4.0 5.0; 5.0 -6.0]
+
+    x = Variable(2)
+    λ = Variable()
+
+    A = A₀ + A₁*x[1] + A₂*x[2]
+
+    p = minimize(λ,A⪯λ; numeric_type=T)
```
ericphanson, line 1386: I think the problem here actually does something different: it minimizes λ such that `λ .- A` is positive semi-definite (note the `.-` instead of `-`). I hope this has not affected your or other's results; I see how it's very confusing.
```suggestion
    p = minimize(λ,A⪯λ*eye(2); numeric_type=T)
```
Maybe we should add this warning somewhere to the docs or examples? And/or deprecate the current behavior and error when there's both matrices and scalars in PSD constraints.

Here I'm using `eye(2)` which is defined in the `problem_depot.jl` file. Usually I'd use `using LinearAlgebra; I(2)` but that doesn't work on Julia 1.0.

src/atoms/affine/index.jl
```diff
@@ -40,9 +40,9 @@ end
 
 function evaluate(x::IndexAtom)
     if x.inds === nothing
-        return getindex(evaluate(x.children[1]), x.rows, x.cols)
+        return output(getindex(evaluate(x.children[1]), x.rows, x.cols)) # output() needed to convert a 1-element vector to a scalar, see the link to the Issue below.
     else
-        return getindex(evaluate(x.children[1]), x.inds)
+        return output(getindex(evaluate(x.children[1]), x.inds)) # output() needed to convert a 1-element vector to a scalar. In response to https://github.com/jump-dev/Convex.jl/issues/447.
     end
 end
```
ericphanson, line 47: I find it a little bit nicer looking written like this:
```julia
function evaluate(x::IndexAtom)
    result = if x.inds === nothing
        getindex(evaluate(x.children[1]), x.rows, x.cols)
    else
        getindex(evaluate(x.children[1]), x.inds)
    end
    # `output` needed to convert a 1-element array to a scalar (https://github.com/jump-dev/Convex.jl/issues/447)
    return output(result)
end
```

ericphanson (commented):
Thank you, this is great!

I had a few minor suggestions, and one comment about how Convex's semidefinite constraints treat `A ⪯ λ` for scalar `λ` in a confusing way.

src/problem_depot/problems/sdp.jl
hurak: Sure. Mathematically it is certainly more appropriate.

src/problem_depot/problems/sdp.jl
hurak: In fact, I originally quickly tried (after `using LinearAlgebra`)

```julia
julia> problem = minimize(λ,A⪯λ*I)
ERROR: MethodError: no method matching *(::Variable, ::UniformScaling{Bool})
Closest candidates are:
  *(::Any, ::Any, ::Any, ::Any...) at operators.jl:560
  *(::Union{MathOptInterface.ScalarAffineFunction{T}, MathOptInterface.ScalarQuadraticFunction{T}, MathOptInterface.VectorAffineFunction{T}, MathOptInterface.VectorQuadraticFunction{T}}, ::T) where T at /home/hurak/.julia/packages/MathOptInterface/YDdD3/src/Utilities/functions.jl:2995
  *(::Convex.AbstractExpr, ::Union{Number, AbstractArray}) at /home/hurak/.julia/dev/Convex/src/atoms/affine/multiply_divide.jl:103
  ...
Stacktrace:
 [1] top-level scope
   @ none:1
```
and seeing it fail (I think I still do not quite master the `UniformScaling` operator), rather than going with the ugly `Matrix(1.0I, 2, 2) ` (I was not even aware of the possibility of the usage `I(2)`), I just tried going with just a scalar on the r.h.s. And I did not realize that broadcasting will be applied here. 

And no, it did not affect my other results since the whole example is artificially crafted only to demonstrate the bug. That's why I did not discover that it does not compute what it should. I only blindly focused on the bug and did not think about the actual value of the result.

ericphanson: okay, glad to hear it wasn't a problem! Would be nice also if we could support UniformScaling's directly, maybe I'll file an issue for that.

hurak: Only now I realized that with the changed problem formulation, the correct solutions that I hardwired into the tests must be changed. Honestly, I do not know how to proceed formally now, after the PR has been submitted (I am still not quite fluent with this git/github collaborative development) and I do not want to screw it up. I will perhaps only put the correct lines here:
```julia
if test
    @test λ.value ≈ 2.400000051025101 atol=atol rtol=rtol
    @test x.value ≈ [3.0000000535867315, -0.4000000018585541] atol=atol rtol=rtol
    @test evaluate(A) ≈ [2.400000046152515 -9.292770553059881e-9; -9.292770553059881e-9 2.399999957564593] atol=atol rtol=rtol
end
```
They should go to the end of src/problem_depot/problems/sdp.jl.

ericphanson: Sure, so all you need to do is make another commit on your branch with those changes, and do `git push`. It will automatically update the PR here. (In general GitHub will keep the PR in-sync with your branch, so any changes you do there should be reflected here). Also, just so you know, clicking "resolve conversation" doesn't actually make any changes, it just hides the comments. (Clicking "commit suggestion" does make changes though).

ericphanson:
Thanks very much @hurak! I'll tag a release shortly with this fix.

hurak:
Pleasure. Thanks for guiding me through the process. I needed it. But I have learnt a bit and next time it will be smoother.

ericphanson:
No problem! I'm very happy to get more people involved with Convex.jl, I think more eyes on the code here will be good, and I don't have so much time for it these days (I used to use it for my PhD work but I don't use it at my job now, so it's harder to find time).

PR #450, opened 2021-07-25
fix #446

### LINKED ISSUE #446: quadform for positive semi-definite matrices
Hello Convex.jl community,

I am using Julia v1.4 and Convex v0.4.11 and apparently the function `quadform` currently does not support positive semi-definite nor negative semi-definite matrices. This is apparent from [quadform.jl](https://github.com/jump-dev/Convex.jl/blob/master/src/atoms/second_order_cone/quadform.jl) which uses  the `ispodef` routine found in LinearAlgebra.jl to test for positive or negative definiteness.

Unfortunately this restriction is not reflected in the documentation for Convex.jl, which states that to use `quadform` "if _x_ is not constant than _P_ must be symmetric and positive semi-definite". Perhaps the documentation on supported operations was copy/pasted from another language where this functionality is present.

Regardless it is an unfortunate omission, since AFAIK that minimization QPs with positive semi-definite terms in the objective are still convex. Are they still not DCP?

Best,
Joseph Sullivan

ericphanson:


NEWS.md
```diff
@@ -1,3 +1,7 @@
+# Changes in v0.14.13
+
+* fix `quadform` for positive semi-definite matrices (fixes a regression introduced in v0.14.11 that required strictly positive semi-definite inputs).
```
ericphanson, line 3:
```suggestion
* fix `quadform` for positive semi-definite matrices (fixes a regression introduced in v0.14.11 that required strictly positive semi-definite inputs) [#450](https://github.com/jump-dev/Convex.jl/pull/450).
```

src/atoms/second_order_cone/quadform.jl
```diff
@@ -2,16 +2,18 @@ function quadform(x::Value, A::AbstractExpr)
     return x' * A * x
 end
 
+is_psd(A::AbstractMatrix{T}) where {T} = isposdef(A + eps(float(T))*Matrix{T}(I, size(A)...))
```
ericphanson, line 5: not totally sure the best way to check this

PR #456, opened 2021-09-08
add `assume_psd` option to `quadform`

ericphanson:
Tries to address https://discourse.julialang.org/t/solving-a-convex-quadratic-program-using-gurobi-mosek/67214/9

@mtanneau does this seem like a decent approach?


mtanneau:
You'll skip the PSD-ness check by default, but I think the main issue is that, when you do call `is_psd`, you end up calling a **dense** cholesky, which will be terribly expensive.

I would suggest to use a sparse fallback in lieu of 
https://github.com/jump-dev/Convex.jl/blob/0c7d23dd2993a590f989a0bf1a890a393afa7f8b/src/atoms/second_order_cone/quadform.jl#L5

Doesn't have to be done in this PR, but I wanted to put it on the radar

ericphanson:
> You'll skip the PSD-ness check by default

To be clear, the default here is false, so we do the check by default.

> I would suggest to use a sparse fallback 

Ah makes sense. Do you have an implementation handy? I’m not familiar with sparse factorizations.

mtanneau:
> To be clear, the default here is false, so we do the check by default.

Oh, sorry, I missed that. Do you need to make the check by default? If you reformulate into SOC constraint then sure, but if it's an objective you can also let the solver deal with that.

mtanneau:
>Do you have an implementation handy? I’m not familiar with sparse factorizations.

SuiteSparse only supports `Float64`. I would suggest using `LDLFactorizations`, which supports all arithmetics.
I can PR into this PR :)

ericphanson:
> If you reformulate into SOC constraint then sure, but if it's an objective you can also let the solver deal with that.

Yeah, unfortunately Convex is only set up to apply every extended formulation it has and pass a SingeVariable objective to MOI. I tried to revamp things in #393 but I only had a week of time for that project and didn’t manage to make it happen.

> Do you know how CVX / cvxpy handle it?

No, good question. I know Convex has always done a check (previously by just calling `eigvals`).

> SuiteSparse only supports Float64. I would suggest using LDLFactorizations, which supports all arithmetics.
I can PR into this PR :)

Cool!

mtanneau:
I dug a little bit into cvxpy code:


* They do have a [convexity check](https://github.com/cvxpy/cvxpy/blob/f722fadcef0d1c877eb63438c27d50c5077f2ed3/cvxpy/atoms/quad_form.py#L61-L65)...
* ...

NB: adding pointer to [this PR](https://github.com/cvxpy/cvxpy/pull/1456) which might be useful

mtanneau:
Arf, LDLFactorizations does not support complex eltype :(

PR #457, opened 2021-09-11
Use sparse symmetric factorization when possible

mtanneau:
Ideally, we would be using a pure iterative method to estimate the smallest (most negative) eigenvalue of a hermitian matrix `A`... similar to what they use in cvxpy.

For pointer, I think we would want something like [Rayleigh quotient method](https://en.wikipedia.org/wiki/Rayleigh_quotient_iteration).
As far as I understand (though I might be wrong here), something similar is implemented in both the scipy method used by cvxpy and IterativeSolvers' [invpowm](https://iterativesolvers.julialinearalgebra.org/dev/eigenproblems/power_method) when "shift-invert" mode is enabled (the same operator shows up).

While these are iterative methods, the hidden fact is that every iteration requires a left-product with `inv(A - shift * I)`, which would require either a factorization, or another iterative method embedded in the first.

Thus, I decided to systematically use a factorization, namely, the code will attempt to factorize `A + shift*I`, and if it fails or finds a negative eigenvalue, `A` is declared not PSD.
The last hurdle is to use an efficient factorization backend, and I try to use sparse fallbacks as much as possible.
The final fallbacks are as follows:
* if `A` is sparse and real-valued, I use LDLFactorizations
* if `A` is sparse and `Complex{Float32}` or `Complex{Float64}`-valued, I use SuiteSparse
* everything else uses a dense fallback.

The tolerance for negative eigenvalues is the parameter `tol`, which is set to the square-root of the machine precision, after conversion to floating-point and real as appropriate.


src/atoms/second_order_cone/quadform.jl
```diff
@@ -2,7 +2,38 @@ function quadform(x::Value, A::AbstractExpr; assume_psd=false)
...
+    if T <: Real || T <: LinearAlgebra.BlasFloat
+        return is_psd(sparse(A); tol=tol)
+    else
+        return is_psd(Matrix(A); tol=tol)
+    end
+end
+is_psd(A::SparseMatrixCSC{Complex{T}}; tol=sqrt(eps(T))) where{T<:LinearAlgebra.BlasReal} = isposdef(A + tol*I)
+function is_psd(A::SparseMatrixCSC{T}; tol::T=sqrt(eps(T))) where{T<:Real}
+    # LDLFactorizations requires the input matrix to only have the upper triangle.
+    F = lu(A - shift * I)
```
ericphanson, line 24: What is `shift` here?

src/atoms/second_order_cone/quadform.jl
```diff
@@ -2,7 +2,38 @@ function quadform(x::Value, A::AbstractExpr; assume_psd=false)
...
+    A_ = Symmetric(sparse(UpperTriangular(A)) + tol*I)
+    try
+        F = ldl(A_)
+        d = F.D.diag
+        return minimum(d) >= 0
+    catch err
+        # If the matrix could not be factorized, then it is not PSD
+        isa(err, LDLFactorizations.SQDException) && return false
+        rethrow(err)  # Something else happened
```
ericphanson, line 33:
```suggestion
        rethrow()  # Something else happened
```
otherwise we truncate the stacktrace (https://docs.julialang.org/en/v1/base/base/#Base.rethrow)

src/atoms/second_order_cone/quadform.jl
```diff
@@ -2,7 +2,38 @@ function quadform(x::Value, A::AbstractExpr; assume_psd=false)
     return x' * A * x
 end
 
-is_psd(A::AbstractMatrix{T}) where {T} = isposdef(A + sqrt(eps(float(real(T))))*Matrix{T}(I, size(A)...))
+"""
+    is_psd(A; tol)
+
+Check whether `A` is positive semi-definite by computing a LDLᵀ factorization of `A + tol*I`
+"""
+function is_psd(A; tol=sqrt(eps(loat(real(eltype(A))))))
```
ericphanson, line 10:
```suggestion
function is_psd(A; tol=sqrt(eps(float(real(eltype(A))))))
```

ericphanson (commented):
Thanks @mtanneau!

I see LDLFactorizations is LGPL licensed. From some brief reading it seems like that should be fine, but since it’s the first non-MIT license I wonder if any of our users would have an issue with it?

I think Invenia uses or used to use Convex, maybe @nickrobinson251 or @iamed2 knows if the license would be an issue for them?

src/atoms/second_order_cone/quadform.jl
mtanneau: Thanks, that line should not have been there. I've removed it.

mtanneau:
> I see LDLFactorizations is LGPL licensed

We're only linking against it, so we're fine. The only restriction is that, if someone distributes something that includes the source code (I think a docker image or a compiled executable would qualify), then they have to make the code of LDLFactorizations available.
If your users don't get the library from you (e.g. they use Julia's package manager), then there's no issue because you are not re-distributing the source code of LDLFactorizations.

iamed2:
Invenia used to use Convex, but it also wouldn't have been a problem for us then.

ericphanson (approved):
Sounds good, thanks both!

PR #462, opened 2021-09-25
Improve numerical stability of `logsumexp`

### LINKED ISSUE #461: Consider (probably) numerically stable way of calculating `logsumexp`
Hi, developers :)

I found that `Convex.logsumexp` directly calculates this operation as [this](https://github.com/jump-dev/Convex.jl/blob/4068f80ec581ae77d8000d7f3fcb0387ce306577/src/atoms/exp_cone/logsumexp.jl#L41), while `Flux.logsumexp` chose a different way, which is explained as "numerically stable".
How about considering it?

~~Note that I left [an issue](https://github.com/FluxML/Flux.jl/issues/1718) in which I asked about "how it makes numerical stability". It would be helpful.~~ [This](https://gregorygundersen.com/blog/2020/02/09/log-sum-exp/) is an explanation of log-sum-exp trick.

JinraeKim:
Resolves #461.

Please refer to [this article](https://gregorygundersen.com/blog/2020/02/09/log-sum-exp/) how it makes `logsumexp` evaluation numerically stable.

### DIFF
--- a/src/atoms/exp_cone/logsumexp.jl
+++ b/src/atoms/exp_cone/logsumexp.jl
@@ -39,8 +39,10 @@ end
 
 function evaluate(x::LogSumExpAtom)
     _x = evaluate(x.children[1])
+    @show _x
     max_x = maximum(_x)
-    return max_x .+ log(sum(exp.(_x .- max_x)))
+    @show max_x
+    return max_x + log(sum(exp.(_x - max_x)))
 end
 
 logsumexp(x::AbstractExpr) = LogSumExpAtom(x)
--- a/src/atoms/exp_cone/logsumexp.jl
+++ b/src/atoms/exp_cone/logsumexp.jl
@@ -39,10 +39,8 @@ end
 
 function evaluate(x::LogSumExpAtom)
     _x = evaluate(x.children[1])
-    @show _x
     max_x = maximum(_x)
-    @show max_x
-    return max_x + log(sum(exp.(_x - max_x)))
+    return max_x + log(sum(exp.(_x .- max_x)))
 end
 
 logsumexp(x::AbstractExpr) = LogSumExpAtom(x)
--- a/src/atoms/exp_cone/logsumexp.jl
+++ b/src/atoms/exp_cone/logsumexp.jl
@@ -38,7 +38,9 @@ function curvature(x::LogSumExpAtom)
 end
 
 function evaluate(x::LogSumExpAtom)
-    return log(sum(exp.(evaluate(x.children[1]))))
+    _x = evaluate(x.children[1])
+    max_x = maximum(_x)
+    return max_x .+ log(sum(exp.(_x .- max_x)))
 end
 
 logsumexp(x::AbstractExpr) = LogSumExpAtom(x)
--- a/test/test_utilities.jl
+++ b/test/test_utilities.jl
@@ -512,4 +512,9 @@ end
         @test_throws ErrorException quadform(x, A; assume_psd=false)
         @test quadform(x, A; assume_psd=true) isa Convex.AbstractExpr
     end
+
+    @testset "`logsumexp` stability" begin
+        v = Convex.Constant([1000, 1000, 1000])
+        @test Convex.evaluate(Convex.logsumexp(v)) ≈ 1001.098612
+    end
 end


ericphanson:
Thanks! Could we add a test? Could do e.g.
```julia
using Convex
v = Constant([-1000, -1000, 1000])
@test evaluate(logsumexp(v)) ≈ 1000.0
```
in `test/test_utilities.jl` (or maybe `test/definitions.jl`), using an example from that post.

By the way, `evaluate` is only used at the end of solving the problem in order to evaluate the values of variables and expressions; it isn't used by the solver itself (which uses the conic form). Also, there's https://github.com/JuliaStats/LogExpFunctions.jl with fancier implementations, though the one you propose is simple and doesn't need any dependencies, and seems like it would be enough for this purpose.

JinraeKim:
@ericphanson 
I added some lines to `test/test_utilities.jl`.
Please refer to the following output of Julia REPL.
Actually, I'm not familiar with testing functions. Please check the new commit.

Note that I checked that with the previous `evaluate(logsumexp(...))` yields `Inf` as expected...!



```julia
Test Summary:                                  | Pass  Fail  Error  Total
Utilities                                      |  388     1      5    394
  `solve!` does not return anything            |                 1      1
  `silent_solver` works                        |                 1      1
  `solve!` can take an optimizer directly      |                 1      1
  `solve!` with MPB solver errors              |          1             1
  Complex objective function errors            |    1                   1
  `optval` is nothing before `solve!`          |    1            1      2
  Default problem type is `Float64`            |    1                   1
  `set_value!` doesn't convert to `Float64`    |    3                   3
  Show                                         |    7            1      8
  vartype and set_vartype                      |   21                  21
  Constructors                                 |  242                 242
  ConicObj                                     |    4                   4
  ConicObj                                     |    4                   4
  length and size                              |    8                   8
  lastindex and axes                           |    6                   6
  Parametric constants                         |   22                  22
  #341: Evaluate for constants                 |    9                   9
  Base.vect                                    |    1                   1
  Iteration                                    |    4                   4
  DCP warnings                                 |    3                   3
  `add_constraints!` (#380)                    |    4                   4
  `diagm` (#401)                               |    1                   1
  `is_psd` with type Float64                   |    6                   6
  `is_psd` with type Float32                   |    6                   6
  `is_psd` with type Int64                     |    6                   6
  `is_psd` with type ComplexF64                |    6                   6
  `is_psd` with type BigFloat                  |    6                   6
  `is_psd` with type Rational{BigInt}          |    6                   6
  `is_psd` with type Complex{Rational{BigInt}} |    6                   6
  `assume_psd` option for `quadform`           |    3                   3
  `logsumexp` stability                        |    1                   1
ERROR: Some tests did not pass: 388 passed, 1 failed, 5 errored, 0 broken.
Stacktrace:
 [1] finish(ts::Test.DefaultTestSet)
   @ Test /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/Test/src/Test.jl:913
 [2] macro expansion
   @ /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/Test/src/Test.jl:1161 [inlined]
 [3] top-level scope
   @ ~/.julia/dev/Convex/test/test_utilities.jl:24
in expression starting at /Users/jinrae/.julia/dev/Convex/test/test_utilities.jl:22

```

ericphanson:
Hm, the test you added looks good, and those test failures seem unrelated, which is a bit strange-- have you happened to change other things in your copy of Convex that's not in this PR?

The tests all passed on the previous commit: https://github.com/jump-dev/Convex.jl/runs/3707979866#step:6:335. And we'll be able to see in a few minutes when https://github.com/jump-dev/Convex.jl/pull/462/checks?check_run_id=3708083850 finishes.

JinraeKim:
> Hm, the test you added looks good (although I'm a bit surprised we don't get 1000.0), and those test failures seem unrelated, which is a bit strange-- have you happened to change other things in your copy of Convex that's not in this PR?
> 
> The tests all passed on the previous commit: https://github.com/jump-dev/Convex.jl/runs/3707979866#step:6:335. And we'll be able to see in a few minutes when https://github.com/jump-dev/Convex.jl/pull/462/checks?check_run_id=3708083850 finishes.

Oh, sorry. The reason why we didn't get 1000.0 is that I used the first example of [this](https://gregorygundersen.com/blog/2020/02/09/log-sum-exp/), while your example is with `[-1000, -1000, 1000]`.
I misread your example. Sorry (but the result of the first example is the same as the test's result).

The other errors are probably due to my lack of experiences about testing scripts :( Sorry for that.

ericphanson:
No problem at all!
That example looks good too. And tests passed here (except on nightly which has been failing for awhile) so I think this is good to go.

JinraeKim:
That's nice to hear 👍

ericphanson:
Btw the package TestEnv makes running tests easier and it will set up and environment with the test dependencies.

JinraeKim:
@ericphanson 
[This one](https://github.com/JuliaTesting/TestEnv.jl)?
I'll take a look at it. Thank you!

ericphanson:
Yeah, that one. You can do
```julia
using TestEnv
TestEnv.activate("Convex")
```
and then it will put you in an environment with SCS, etc defined. Then you can do
```julia
include("test/runtests.jl")
```
and it will run the tests.

You can also do just `] test` or `] test Convex` to use the Pkg stdlib to run the tests. That will start a new Julia process and run them. It's simple and reliable, but you if you run them more than once, you have to wait for compilation each time.

PR #467, opened 2021-11-13
Update to MOI v0.10

### LINKED ISSUE #466: problem with Convex and new SCS (0.8.1)
Julia 1.6.3, Win 10

example from SCS page:
```
using Convex, SCS
items  = [:Gold, :Silver, :Bronze]
values = [5.0, 3.0, 1.0]
weights = [2.0, 1.5, 0.3]

x = Variable(3)
p = maximize(x' * values, 0 <= x, x <= 1, x' * weights <= 3)
solve!(p, SCS.Optimizer)
```

gives 

```
ERROR: LoadError: MethodError: no method matching solve!(::Problem, ::Type{SCS.Optimizer})                                  
Closest candidates are:                                       
  solve!(::Problem; warmstart, check_vexity, verbose) at C:\Users\psoderlind\.julia\packages\Convex\6NNC8\src\solution.jl:16  solve!(::Problem, ::MathProgBase.SolverInterface.AbstractMathProgSolver; kwargs...) at ...
```

ericphanson:
Closes #454
Closes #466

Seems like it's not too bad! Or I haven't gotten to the tests that will fail yet.


ericphanson:
I think the GLPK failures (https://github.com/jump-dev/Convex.jl/runs/4200633319?check_suite_focus=true#step:6:1088) are GLPK issues, not Convex issues, since those tests pass with SCS; likewise, the nightly failure and SCS failure on the lieb ando problem isn't a Convex issue, since nothing there has changed. I'll update ConvexTests to try to surface these issues and file issues for the solvers.

I think this is good to go (assuming the docs build passes; at the time of writing it is still going, but the non-nightly tests have passed).

odow:
Is that really it? I'm surprised. I'll make a new release of GLPK.

ericphanson:
Yeah, the interface with MOI is fairly minimal, so I guess that has helped. That's since we do most things with our internal data structures and then emit a MBP-style conic problem, instead of using MOI bridges. Usually that's been a bad thing (since we can't support all the things that MOI could allow us to via bridges) but here I guess it's been convenient 😄.

PR #468, opened 2021-11-14
Support both MOI v0.9 and MOI v0.10

ericphanson:
Builds off of #467 

I think we can support both MOI v0.9 and MOI v0.10 pretty easily, since there are so few changes we need to make. It's a bit hacky, but I think it is worth it, since there might be a long tail of solvers that aren't updated. E.g. dropping MBP support in Convex was pretty annoying for me personally (even though I was the one who did it) since I ended up wanting to use Pajarito, and this requires much less compat shimming than that would have.

I've updated CI here to test with both versions.


odow (approved):
Nice.  If you want something that doesn't use `try-catch`, I think you should be able to use 
```julia
if !isdefined(MOI, :SingleVariable) || MOI.SingleVariable isa Function
    const SV_OR_VI, SV_OR_IDENTITY = MOI.VariableIndex, identity
else
    const SV_OR_VI, SV_OR_IDENTITY = MOI.SingleVariable, MOI.VariableIndex
end
```

ericphanson:
Thanks, that's nicer! (The second branch should have `MOI.SingleVariable, MOI.SingleVariable`, by the way). Made that change in f0eac7a; if CI passes I'll merge+tag.

PR #476, opened 2022-02-18
Update to MOI v1.0

odow:
CI is going to fail because it needs some solvers tagged, but leaving this PR as a placeholder.


odow:
I've marked the next version as `0.15.0-DEV`, and will merge once CI passes. The changes aren't controversial.

PR #484, opened 2022-03-02
Variety of fixes to the docs

### LINKED ISSUE #483: Fix the docs
It looks like the docs don't have tests to check they ran okay: https://jump.dev/Convex.jl/dev/quick_tutorial/.

This is a blocker for v0.15 release.

odow:
* Tidy make.jl
 * Enforce strict=true so failing docs not ignored
 * Use silent_solver where possible
 * Change repl blocks to example

Closes #483, but still running locally so there might be some failures.


odow:
@ericphanson any objections to me tagging a v0.15 release?

ericphanson:
Nope, fine by me! Thanks for all this great maintenance work.

PR #486, opened 2022-03-07
fix lasso_regression (example)

### LINKED ISSUE #485: LASSO/Elastic Net example
at https://jump.dev/Convex.jl/stable/examples/general_examples/lasso_regression/

needs an update after the latest Convex/SCS updates. (See the last figure for why.) I hope to give it a try in a week or so, but please feel welcome to beat me to it.

PaulSoderlind:
This makes the lasso_regression example work again (after the lastest Convex/SCS changes). I also simplified one or two things.

Closes #485


odow:
Thanks @PaulSoderlind it looks good now.

PR #489, opened 2022-03-25
Change Dict -> OrderedDict

### LINKED ISSUE #488: Different result each time problem is solved
This issue was originally opened as https://github.com/chriscoey/Hypatia.jl/issues/801#issuecomment-1041173544

The following non-random problem produces very different results each time the code is run. The problem does not appear to be within the solver Hypatia since if only the part after and including `prob = Convex.minimize(γ)` is run, the results are the same each time. 

```julia
using Hypatia, Convex, LinearAlgebra
(A, B1, B2, C1, C2, D11, D12, D21, D22) = ([0.0 10.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0; -10.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0; -4.0 -40.0 -0.01 0.0 0.0 0.0 0.0 0.0 0.0 0.0; 40.0 -4.0 0.0 -0.01 0.0 0.0 0.0 0.0 0.0 0.0; 0.0 0.0 0.0 0.0 -3535.46761115006 3535.46761115006 0.0 0.0 0.0 0.0; 0.0 0.0 0.0 0.0 -3535.46761115006 -3535.46761115006 0.0 0.0 0.0 0.0; 0.0 0.0 0.0 0.0 0.0 0.0 -3535.46761115006 3535.46761115006 0.0 0.0; 0.0 0.0 0.0 0.0 0.0 0.0 -3535.46761115006 -3535.46761115006 0.0 0.0; 512.0 5120.0 0.0 0.0 0.0 0.0 0.0 0.0 -2309.2856038177697 0.0; -5120.0 512.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 -2309.2856038177697], [0.0 0.0; 0.0 0.0; 4.0 0.0; 0.0 4.0; 0.0 0.0; 0.0 0.0; 0.0 0.0; 0.0 0.0; 0.0 0.0; 0.0 0.0], [1.0 0.0; 0.0 1.0; 0.0 0.0; 0.0 0.0; 0.0 0.0; 10725.385484097349 0.0; 0.0 0.0; 0.0 10725.385484097349; 0.0 0.0; 0.0 0.0], [-0.1 -1.0 2.49975 0.0 0.0 0.0 0.0 0.0 0.0 0.0; 1.0 -0.1 0.0 2.49975 0.0 0.0 0.0 0.0 0.0 0.0; 0.0 0.0 0.0 0.0 -46.28786141317 -6546.0921383757795 0.0 0.0 0.0 0.0; 0.0 0.0 0.0 0.0 0.0 0.0 -46.28786141317 -6546.0921383757795 0.0 0.0; 100.0 1000.0 0.0 0.0 0.0 0.0 0.0 0.0 -448.77718277317985 0.0; -1000.0 100.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 -448.77718277317985], [-1.0 -10.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0; 10.0 -1.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0], [0.1 0.0; 0.0 0.1; 0.0 0.0; 0.0 0.0; 0.0 0.0; 0.0 0.0], [0.0 0.0; 0.0 0.0; 10000.0 0.0; 0.0 10000.0; 0.0 0.0; 0.0 0.0], [1.0 0.0; 0.0 1.0], [0.0 0.0; 0.0 0.0])

ϵ = 1e-3

n = size(A,1)
ny1 = size(C1, 1)
nu1 = size(B1, 2)
ny2 = size(C2, 1)
nu2 = size(B2, 2)
γ  = Convex.Variable()
X1 = Convex.Variable(n,n)
Y1 = Convex.Variable(n,n)
An = Convex.Variable(n,n)
Bn = Convex.Variable(n, ny2)
Cn = Convex.Variable(nu2, n)
Dn = Convex.Variable(nu2, ny2)

M11 = A*Y1 + Y1*A' + B2*Cn + Cn'*B2'
M21 = A' + An + (B2*Dn*C2)'
M31 = (B1 + B2*Dn*D21)'
M41 = C1*Y1 + D12*Cn
M22 = X1*A + A'*X1 + Bn*C2 + C2'*Bn'
M32 = (X1*B1 + Bn*D21)'
M42 = C1 + D12*Dn*C2
M43 = D11 + D12*Dn*D21
M33 = -γ*Matrix(I(nu1))
M44 = -γ*Matrix(I(ny1))
M = [
    M11 M21' M31' M41'
    M21 M22  M32' M42'
    M31 M32  M33  M43'
    M41 M42  M43  M44
]
prob = Convex.minimize(γ)
prob.constraints += γ > 0
prob.constraints += M ⪯ -ϵ*I(size(M,1))
prob.constraints += [X1 I(n); I(n) Y1] ⪰ ϵ*I(2n)

opt = Hypatia.Optimizer(rescale=false)
Convex.solve!(prob, opt; silent_solver=false, verbose=true)
```
Tested on Convex v0.15.0

baggepinnen:
closes #488

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Convex"
 uuid = "f65535da-76fb-5f13-bab9-19810c17039a"
-version = "0.15.0"
+version = "0.15.1"
 
 [deps]
 AbstractTrees = "1520ce14-60c1-5f80-bbc7-55ef81b5835c"
--- a/docs/src/release_notes.md
+++ b/docs/src/release_notes.md
@@ -1,5 +1,8 @@
 # Release notes
 
+## v0.15.1 (March 28, 2022)
+* Use `OrderedDict` internally for reproducible results, issue: [#488](https://github.com/jump-dev/Convex.jl/issues/488), fix: [#489](https://github.com/jump-dev/Convex.jl/pull/489).
+
 ## v0.15.0 (March 2, 2022)
 
 ### Breaking changes
--- a/src/atoms/affine/stack.jl
+++ b/src/atoms/affine/stack.jl
@@ -45,7 +45,7 @@ function conic_form!(x::HcatAtom, unique_conic_forms::UniqueConicForms)
             push!(objectives, conic_form!(child, unique_conic_forms))
         end
         # build a dict from variable ids to sizes
-        variable_to_sizes = Dict{UInt64,Int}()
+        variable_to_sizes = OrderedDict{UInt64,Int}()
         for objective in objectives
             for id in keys(objective)
                 if !(id in keys(variable_to_sizes))
--- a/src/conic_form.jl
+++ b/src/conic_form.jl
@@ -107,7 +107,7 @@ const UniqueConstrMap = OrderedDict{Tuple{Symbol,UInt64},Int}
 const UniqueConstrList = Vector{ConicConstr}
 # map variables' hash to the variable itself 
 const IdToVariables = OrderedDict{UInt64,AbstractVariable}
-const ConicConstrToConstr = Dict{ConicConstr,Constraint}
+const ConicConstrToConstr = OrderedDict{ConicConstr,Constraint}
 # UniqueConicForms caches all the conic forms of expressions we've parsed so far
 struct UniqueConicForms
     exp_map::UniqueExpMap


baggepinnen:
@ericphanson would you mind triggering the CI to run on this PR? Thanks :)

ericphanson (commented):
LGTM, but if this does fix the issue, it would be great to have a test to ensure we don't regress

baggepinnen:
I can take the code from the issue #448 and solve it a number of times and check that the solutions are identical? I'm afraid I'm not familiar enough with the internals of Convex.jl to come up with a more targeted test.

ericphanson:
Do you know the optimal value? And how fast is it to solve? If it's relatively fast and you know the optimal value (not necessiarly which solution just the min) then it seems like a good candidate for the [ProblemDepot](https://jump.dev/Convex.jl/dev/problem_depot/); can add it here: https://github.com/jump-dev/Convex.jl/blob/master/src/problem_depot/problems/sdp.jl. This has the added benefits that solvers' who choose to test themselves against the ProblemDepot will have this problem in their tests, and it is used in [ConvexTests](https://github.com/ericphanson/ConvexTests.jl).

Though for solving it twice and getting the same solution... Yeah, I'm not sure we should add that to the ProblemDepot test, because some solvers may have their own internal non-determinism. So I guess this is OK as-is, not really sure how to test it either.

baggepinnen:
> Do you know the optimal value?

No :( 
Before this PR Hypatia occasionally produced a slightly better value than what is produced by Hypatia after this PR (but often much worse), but the PR makes it produce exactly the same result all times. I never get an optimal return status, it exists with either slow progress or almost optimal. 

It takes about 0.5 seconds to solve by Hypatia, but, e.g., SCS fails completely.

ericphanson:
I think the warning is from Convex but everything before that is from Hypatia.

Ok let’s merge this, since it’s an improvement. Do you mind bumping the patch version and adding a NEWS entry so we can make a release?

baggepinnen:
Done :)

chriscoey:
Nice!

Btw I can help with Hypatia options and numerics after my thesis is done at the end of April.

PR #494, opened 2022-05-31
remove deprecated AbstractTrees calls

ExpandingMan:
I've recently done a major overhaul of AbstractTrees.jl which will be part of the upcoming v0.4 release for that package.  Convex.jl is unique in that it is the only registered dependency of AbstractTrees.jl which uses the soon-to-be-deprecated functions `treekind` and `IndexedTree`.

It seems that Convex.jl contains a re-implementation of `print_tree` because of a problem that dates back to AbstractTrees.jl v0.2.  At first, I completely removed this re-implementation in favor of the current AbstractTrees implementation, but later realized that it has since diverged from `AbstractTrees.print_tree` even in v0.3.  In particular, while `maxdepth` has worked fine for a while, we do not have a `maxwidth`.  There are also some small differences in the exact printed forms.

Instead, I've removed the calls to `treekind` and `IndexedTree`.  This should be fine for all versions of AbstractTree.jl because these features were very poorly documented and almost never used (which has now been [explicitly verified](https://github.com/JuliaCollections/AbstractTrees.jl/issues/103)).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Convex"
 uuid = "f65535da-76fb-5f13-bab9-19810c17039a"
-version = "0.15.1"
+version = "0.15.2"
 
 [deps]
 AbstractTrees = "1520ce14-60c1-5f80-bbc7-55ef81b5835c"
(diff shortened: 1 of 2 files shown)


odow (approved):

odow:
You happy? Minimal changes is good. At some point, Convex [needs a new maintainer](https://github.com/jump-dev/Convex.jl#please-read-convexjl-needs-a-new-maintainer) to go through and tidy up a lot of the package.

ExpandingMan:
Yes, should be all good.

baggepinnen:
It seems Convex.jl needs a new release to include this change?
#495

ericphanson:
@baggepinnen ok, registering v0.15.2

Issue #310, opened 2019-08-13
Partially specified problems


ericphanson:
There was a post on discourse requesting partially specified problems: <https://discourse.julialang.org/t/convex-jl-trouble-with-partially-specified-function-inside-another-problem/26508/5>. There is also a good description in the CVX manual: http://web.cvxr.com/cvx/doc/advanced.html#new-functions-via-partially-specified-problems. I think this would be great, as it makes a very easy way to add new "atoms" (without actually having to make a struct, define methods, etc). I think it would work like this: the user (or someone implementing a new feature) defines a function which returns a problem, for example

```julia
function lamb_min(A::AbstractExpr)
    t = Variable()
    n = size(A, 1)
    n == size(A,2) || throw(ArgumentError())
    p = maximize(t, A - t*Matrix(1.0I, n, n) ⪰ 0)
    return p
end
```

Note: this is almost the actual implementation of the `lambdamin` atom: https://github.com/JuliaOpt/Convex.jl/blob/8a3e843d69c452ebeec80b24048a9a1e3efb84b9/src/atoms/sdp_cone/lambda_min_max.jl#L109-L110

This already works fine if I want to simply substitute some `AbstractExpr` into `lamb_min`; I could write

```julia
p = lambda_min(aff), constraints
solve!(p, solver)
```
for some affine expression `aff`. And if we wanted to be able to add additional constraints, we could pass those in as another argument to `lamb_min` (which is a bit clunky). But what if we want to substitute the result of `lamb_min` into another expression? E.g.

```julia
A = Variable(2,2)
p = maximize( lamb_min(A) + 1, [ A >= 0, A[1,1] == 2.0] )
```

Then we would like to treat `lamb_min(A)` as a `AbstractExpr` that evaluates to the solution to the inner problem. How do we do that?

1. Have `Problem` subtype `AbstractExpr`. To do so, we just need to add an `id_hash` and a `size`, and I believe add a `children` which yields the field `objective`. The `size` will always be `(1,1)`, so the latter two could be done by overriding `getproperty`.
2. Make sure `vexity(p::Problem)` means what it should for an `AbstractExpr`. Right now, if you minimize an affine function subject to affine constraints, `vexity(p)` will say `AffineVexity()` when it should be `ConvexVexity()` if it were to be treated as an `AbstractExpr`.
3. Make `conic_form!(p::Problem)` consistent with those for `AbstractExpr`. Right now it returns a hash in addition while other `conic_form!`'s do not.

[...]

PR #504, opened 2023-07-06
More MOIified implementation, again

ericphanson:
Updated https://github.com/jump-dev/Convex.jl/pull/393 to the latest master, and made a few fixes and cleanup. Still has a ways to go.

* replaces `UniqueConicForms` construction with similar-in-spirit `Context` object, which now holds the MOI problem so we can add constraints/variables directly there as we traverse the problem. This is the main improvement which lets us lower new cones directly to MOI and handle things like #388.
* makes `context` the first argument to `conic_forms!` to make it clear that is what is being mutated (not the atom)
* removes non-functional caching system
* closes https://github.com/jump-dev/Convex.jl/issues/78 by not using hashing to cache `conic_forms!` within a problem. We do so by simply not caching them. However I think if we wanted to, we could cache by `objectid` instead of hash, xref https://github.com/jump-dev/Convex.jl/pull/393#issuecomment-678726117.
* closes https://github.com/jump-dev/Convex.jl/issues/392 by deleting the file
* closes https://github.com/jump-dev/Convex.jl/issues/388
* does not yet solve https://github.com/jump-dev/Convex.jl/issues/384. I think this one is still somewhat hard (though easier now?) because we have no internal modelling of quadratic things at all. This partial rewrite makes it easier to add new constraints of the form "affine function in cone", and we can add atoms that use such constraints in extended formulations, but we still can't actually have any kind of standalone quadratic atom. Maybe one path there is to move these extended formulations into MOI itself (and DCP modelling so it can know when they are valid), so Convex is an even smaller wrapper, and handle everything at the MOI layer.
* populates duals in more cases. These need tests.


ericphanson:
Status:
* The old system has been replaced by the new one everywhere
* The new system is quite buggy still
* The majority of problems in the test suite can be solved, including most SDPs, SOCPs, exponential cone problems, duals, integer constraints, etc
* in terms of getting all tests passing, I think the biggest remaining issues are around complex variable handling, specifically where problems have a mix of complex variables and non-complex variables. E.g. I am missing a method for `complex_operate(vcat, T, ::ComplexTape, ::SparseAffineTape)`
* in terms of making the new system work reliably, I am concerned about the robustness of the `operate` system implemented here. I think there is too much room to hit unimplemented methods or incorrect fallbacks. I think it needs to be tightened up considerably to prevent unforseen issues.
* I am also concerned that the tests may not cover all problem formulations and there could be correctness issues left after getting all the tests to pass.

From a recent SCS run locally:
```julia
Test Summary:                                              | Pass  Fail  Error  Total     Time
SCS                                                        |  611     5     10    626  4m03.2s
  constant                                                 |   28                  28     4.4s
  affine                                                   |  140                 140    14.1s
  exp                                                      |   27                  27     4.8s
  socp                                                     |   99            1    100    14.5s
  lp                                                       |   56            2     58     4.5s
  sdp                                                      |  258     5      7    270  3m18.8s
```

odow:
> I think there is too much room to hit unimplemented methods or incorrect fallbacks. I think it needs to be tightened up considerably to prevent unforseen issues.

Agree. If you want to see my recent work on this:

 * https://github.com/jump-dev/MathOptInterface.jl/issues/2203
 * https://github.com/jump-dev/MathOptInterface.jl/pull/2208

ericphanson:
@odow would you mind summarizing any tips or general strategy for that work?

ericphanson:
I think I've made some good progress tightening up the `operate` stuff, although there's a few loose ends in the multiply atom to handle still. Otherwise it is pretty controlled and I am much less worried about unexpected situations. I also removed the fallbacks. Additionally, all tests are passing now, except for the `sdp_lieb_ando` ones which had issues on main. Now however we can't even formulate the problem (let alone solve it), which means there's still some bugs to shake out in this PR.

ericphanson:
Ok, I think this is in decent shape now; in particular the `operate` system is much more constrained and there should not be runtime MethodErrors / other errors possible in that system. However depending on what the users passes it could be possible to error elsewhere when trying to convert their type to a matrix or sparse matrix. 

Warmstarts are not re-implemented, but I'm not sure it's essential for this PR. 

There is probably some missing test coverage that should be fixed before we merge however.

I will mark it as ready-for-review now.

ericphanson:
I've been looking into perf since I believe this has regressed things. I also have some updates in #254. Running the benchmarks here, I get:
```
In [10]: include("benchmark/pprintjudge.jl")
Out[10]:   Benchmark Report for /Users/eph/Convex.jl
  ≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡

  Job Properties
  ================

    •  Time of benchmarks:
       • Target: 1 Aug 2023 - 09:52
       • Baseline: 1 Aug 2023 - 09:53

    •  Package commits:
       • Target: 4ecbe7
       • Baseline: 614b48

    •  Julia commits:
       • Target: e4ee48
       • Baseline: e4ee48

    •  Julia command flags:
       • Target: None
       • Baseline: None

    •  Environment variables:
       • Target: JULIA_NUM_THREADS => 1
       • Baseline: JULIA_NUM_THREADS => 1

  Results
  =========

  A ratio greater than 1.0 denotes a possible regression (marked with ❌), while a ratio less than 1.0 denotes a possible improvement
  (marked with ✅). Only significant results - results that indicate possible regressions or improvements - are shown below (thus, an
  empty table means that all benchmark results remained invariant between builds).

                                                                 ID  time ratio memory ratio
  ––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––– ––––––––––– ––––––––––––
              ["formulation", "affine", "affine_dot_multiply_atom"] 4.28 (5%) ❌  1.25 (1%) ❌
                      ["formulation", "affine", "affine_hcat_atom"] 1.83 (5%) ❌  0.93 (1%) ✅
                     ["formulation", "affine", "affine_trace_atom"] 5.22 (5%) ❌  1.23 (1%) ❌
  ["formulation", "constant", "constant_fix!_with_complex_numbers"] 3.10 (5%) ❌  1.05 (1%) ❌
                         ["formulation", "exp", "exp_entropy_atom"] 2.00 (5%) ❌  0.90 (1%) ✅
                 ["formulation", "exp", "exp_log_perspective_atom"] 2.26 (5%) ❌  0.75 (1%) ✅
                           ["formulation", "lp", "lp_maximum_atom"] 3.00 (5%) ❌  0.73 (1%) ✅
                    ["formulation", "mip", "mip_integer_variables"] 3.98 (5%) ❌  1.11 (1%) ❌
                      ["formulation", "sdp", "sdp_lambda_min_atom"] 1.80 (5%) ❌  0.55 (1%) ✅
                     ["formulation", "sdp", "sdp_sum_largest_eigs"] 1.38 (5%) ❌  0.53 (1%) ✅
         ["formulation", "sdp_and_exp", "sdp_and_exp_log_det_atom"] 0.80 (5%) ✅  0.24 (1%) ✅
                     ["formulation", "socp", "socp_quad_form_atom"] 2.30 (5%) ❌  0.95 (1%) ✅
                   ["formulation", "socp", "socp_sum_squares_atom"] 3.00 (5%) ❌  0.94 (1%) ✅
```
So we have regressed formulation for a bunch of problems, although with better memory consumption in many cases.

Right now I maintain a lazy "tape" of affine operations to compose. (This came from the original PR #393). I think it maybe is a mistake and will try without that.

ericphanson:
Probably just needs some review and a rebase. If you are able to give some review that would help move it forward. I don’t remember there being any real blockers or big TODOs left

docs/examples_literate/mixed_integer/aux_files/antidiag.jl
```diff
@@ -79,31 +79,27 @@ antidiag(x::AbstractExpr, k::Int = 0) = AntidiagAtom(x, k)
 # 3. We populate coeff with 1s at the correct indices
 # The canonical form will then be:
 # coeff * x - d = 0
-function conic_form!(
+function Convex._conic_form!(
```
blegat, line 82: @ericphanson This cannot be an internal function if it's something that can be implemented by extensions

blegat: What about `new_conic_form!` ?

blegat: I have renamed it in https://github.com/jump-dev/Convex.jl/pull/504/commits/a8b44b212d822552221526f4cbd011f30336b73a

odow:
I guess just a note: this is a very large change. Are we just going to YOLO a breaking release? Or can we take steps to split it into some smaller PRs?

blegat:
@odow @ericphanson Any objection to merge this and follow the plan of my above comment ? Since this PR has such a large diff, it's going to be a pain to review changes on top of it and I'd like to try a few things.

ericphanson:
I think that seems OK, especially if you've been looking through the diff as you've worked here. I think that can effectively be some amount of review. At work I'd suggest we pair review it, where we get on a call and go through it together, but realistically I don't think I would have time anytime soon for that.

Are there any areas you are concerned about?

blegat:
One thing I've been wondering is why allowing to pass an instantiated optimizer in `Context` and `solve!` that is allowed to be non-empty. In JuMP, we only allow an optimizer constructor and in MOI.CachingOptimizer, we check with `MOI.is_empty` to check that it is empty. Maybe we could restrict these to only getting an optimizer constructor ?

src/problems.jl
```diff
@@ -31,6 +30,8 @@ function Base.getproperty(p::Problem, s::Symbol)
         else
             return objective_value(p)
         end
+    elseif s === :size
+        return (1, 1)
```
blegat, line 34: I'm not so sure what this is for as well. It actually wasn't tested so I had to add a test

ericphanson:
> One thing I've been wondering is why allowing to pass an instantiated optimizer in Context and solve! that is allowed to be non-empty. In JuMP, we only allow an optimizer constructor and in MOI.CachingOptimizer, we check with MOI.is_empty to check that it is empty. Maybe we could restrict these to only getting an optimizer constructor ?

Hm, that would make sense. I don't think I understood (and probably still don't understand) the semantics of optimizers and just messed with it until it seemed to work. Changing it seems totally fine.

src/problems.jl
ericphanson: hm, if it was coming up in codecov as untested maybe it's not used at all and we can delete it. I don't really remember this, but if I had to guess it would be for something like partially specified problems (#310)

blegat: Let's try to remove it and see if anything breaks:
https://github.com/jump-dev/Convex.jl/pull/504/commits/5b401f0d9615f173a34d93c2ba2f3080f210c003

blegat:
I have done the changes suggested in https://github.com/jump-dev/Convex.jl/pull/504#discussion_r1425519518 and https://github.com/jump-dev/Convex.jl/pull/504#issuecomment-1851960770 and it looks good to me know, I'll merge next week if there is no objection

odow:
This is too big to review. But given Convex.jl has seen very little activity recently, I'm okay releasing this as a breaking change if it is an improvement that fixes open issues.

@mlubin should take a brief look before merging.

odow:
I've added this PR to the agenda for next week's developer call.

mlubin:
It looks like `warmstart=true` is removed and there's a `FIXME` there. Do we plan to merge and release this without the ability to restart from the previous primal/dual solution? Is there a workaround for users?

odow:
There's an open issue: https://github.com/jump-dev/Convex.jl/issues/518

blegat:
> Do we plan to merge and release this without the ability to restart from the previous primal/dual solution? Is there a workaround for users?

No, the plan to wait a while before releasing and at least fixing https://github.com/jump-dev/Convex.jl/issues/518

mlubin (approved):
Ok to merge. (Comment can be addressed now or later.)

odow:
Developer call agrees to merge as a breaking change. @blegat plans a few other changes once this lands before we tag the next release.

PR #505, opened 2023-07-22
Write to file

maxkapur:
This PR adds a `write_to_file(problem, filename)` API. It's basically just a copy-paste job of what @ericphanson demonstrated in the enhancement request: https://github.com/jump-dev/Convex.jl/issues/395

Example:

```julia
using Convex, SCS

let
    x = Variable(8)
    problem = minimize(sumsquares(randn(5, 8) * x - randn(5)), [0 <= x, x <= 1])

    solve!(problem, SCS.Optimizer)
    write_to_file(problem, "file.sdpa")
end
```

`file.sdpa`:

```text
11
7
-1 -8 -8 -1 -1 5 2
0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 1.0
0 1 1 1 -1.0
...
```

The function throws an informative error if the user has not yet called `solve!`.

I included a **docstring** and some **unit tests** (`test/write_to_file.jl`).

Note that there is no corresponding `read_from_file()` function, because (if I am not mistaken) boiling a convex problem down into the MOI format via Convex.jl's DCP rules is a nonreversible operation.

### DIFF
--- a/src/problems.jl
+++ b/src/problems.jl
@@ -212,16 +212,16 @@ problem is loaded into a MathOptInterface model.
 The file format is inferred from the filename extension. Supported file
 types depend on the model type.
 """
-function write_to_file(problem::Problem{T}, filename::String) where T
+function write_to_file(problem::Problem{T}, filename::String) where {T}
     isnothing(problem.model) && throw(
         ArgumentError(
             """
             Problem has not been loaded into a MathOptInterface model; 
             call `solve!(problem, optimizer)` before writing problem to file.
-            """
-        )   
+            """,
+        ),
     )
-    dest = MOI.FileFormats.Model(filename=filename)
+    dest = MOI.FileFormats.Model(filename = filename)
     src = problem.model
     MOI.copy_to(MOI.Bridges.full_bridge_optimizer(dest, T), src)
     return MOI.write_to_file(dest, filename) # nothing
--- a/src/problems.jl
+++ b/src/problems.jl
@@ -212,7 +212,7 @@ problem is loaded into a MathOptInterface model.
 The file format is inferred from the filename extension. Supported file
 types depend on the model type.
 """
-function write_to_file(problem::Problem, filename::String)
+function write_to_file(problem::Problem{T}, filename::String) where T
     isnothing(problem.model) && throw(
         ArgumentError(
             """
@@ -223,6 +223,6 @@ function write_to_file(problem::Problem, filename::String)
     )
     dest = MOI.FileFormats.Model(filename=filename)
     src = problem.model
-    MOI.copy_to(MOI.Bridges.full_bridge_optimizer(dest, Float64), src)
+    MOI.copy_to(MOI.Bridges.full_bridge_optimizer(dest, T), src)
     return MOI.write_to_file(dest, filename) # nothing
 end
--- a/src/problems.jl
+++ b/src/problems.jl
@@ -209,18 +209,19 @@ Write the current problem to the file at `filename`. Requires solving
 the problem at least once using [`solve!`](@ref) to ensure that the
 problem is loaded into a MathOptInterface model.
 
-Supported file types depend on the model type.
+The file format is inferred from the filename extension. Supported file
+types depend on the model type.
 """
 function write_to_file(problem::Problem, filename::String)
     isnothing(problem.model) && throw(
         ArgumentError(
             """
-            Problem has not been loaded into a MathOptInterface model; call 
-            `solve!(problem, optimizer)` before writing problem to file.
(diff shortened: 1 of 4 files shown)


src/problems.jl
```diff
@@ -201,3 +201,28 @@ function cache_conic_form!(
...
+    write_to_file(problem::Problem, filename::String)
+
+Write the current problem to the file at `filename`. Requires solving
+the problem at least once using [`solve!`](@ref) to ensure that the
+problem is loaded into a MathOptInterface model.
+
+The file format is inferred from the filename extension. Supported file
+types depend on the model type.
+"""
+function write_to_file(problem::Problem, filename::String)
```
ericphanson, line 215:
```suggestion
function write_to_file(problem::Problem{T}, filename::String) where T
```

src/problems.jl
```diff
@@ -201,3 +201,28 @@ function cache_conic_form!(
...
+        ArgumentError(
+            """
+            Problem has not been loaded into a MathOptInterface model; 
+            call `solve!(problem, optimizer)` before writing problem to file.
+            """
+        )   
+    )
+    dest = MOI.FileFormats.Model(filename=filename)
+    src = problem.model
+    MOI.copy_to(MOI.Bridges.full_bridge_optimizer(dest, Float64), src)
```
ericphanson, line 226:
```suggestion
    MOI.copy_to(MOI.Bridges.full_bridge_optimizer(dest, T), src)
```

ericphanson (commented):
Thanks, this looks good to me! One comment is we can get the element type form the problem, which could let us save out problems at higher precision

maxkapur:
> Thanks, this looks good to me! One comment is we can get the element type form the problem, which could let us save out problems at higher precision

Done! I also fixed the code format. Not sure why the Nightly CI job failed, though—the error came during precompilation and doesn't happen on my local machine. Hoping it was just a fluke. https://github.com/jump-dev/Convex.jl/actions/runs/5633302721/job/15272471727

ericphanson:
That error isn’t related to your PR, but we will need to fix it. The issue is Convex pirates `hcat` (and I think `hvcat` and `vcat`) by overwriting a method which I suppose now errors. (FYI: I think this can be fixed by switching from the pirate approach to an eval-loop for defining methods for “At least one argument of this type” as in https://www.oxinabox.net/2023/06/17/resident-eval.html).

But I think we shouldn’t do anything about it in this PR as it is a separate issue.

For this PR, I think it would be good to add a test for non-Float64 numeric type. This can be done by passing `numeric_type=BigFloat` for example in the `solve!` call.

maxkapur:
OK, I have added test instances with `Float32` and `BigFloat` eltypes to the test. Unfortunately, I can't get Convex.jl to solve problems with non-`Float64` eltypes, so these tests are behind a `@test_broken` flag for now.

Here's an example of the kind of error I'm encountering:

```julia
using Convex, Hypatia

function main()
    m, n = 5, 10
    x = Variable(n)
    A = randn(Float32, m, n)
    b = randn(Float32, m)
    problem::Problem{Float32} = minimize(
        sumsquares(A * x - b),
        [0 <= x, x <= 1],
        numeric_type = Float32,
    )

    return solve!(problem, Hypatia.Optimizer)
end

main()
```

yields

```
ERROR: LoadError: MathOptInterface.UnsupportedConstraint{MathOptInterface.VectorAffineFunction{Float32}, MathOptInterface.SecondOrderCone}: `MathOptInterface.VectorAffineFunction{Float32}`-in-`MathOptInterface.SecondOrderCone` constraint is not supported by the model.
```

In this example, I have used Hypatia.jl as a solver (thinking that a pure-Julia solver will be more likely to support the eltype abstraction), but the test code uses SCS (and generates a similar error) to avoid the need to add a new test dependency.

A similar error occurs for the `BigFloat` eltype (in both SCS and Hypatia).

I noticed that all of the other examples in the test repo are using `Float64`, and I can't see any mention of other eltypes in the documentation. Is this worth filing a separate issue, or is the work you are doing on the more MOIified interface likely to fix this up anyway?

src/problems.jl
```diff
@@ -201,3 +201,28 @@ function cache_conic_form!(
...
+        ArgumentError(
+            """
+            Problem has not been loaded into a MathOptInterface model; 
+            call `solve!(problem, optimizer)` before writing problem to file.
+            """,
+        ),
+    )
+    dest = MOI.FileFormats.Model(filename = filename)
+    src = problem.model
+    MOI.copy_to(MOI.Bridges.full_bridge_optimizer(dest, T), src)
```
odow, line 226: The file formats support only `Float64`.

odow:
Closing in favor of #531

PR #511, opened 2023-10-20
Add MOI optimizer

blegat:
The goal of this PR is to provide
```julia
using JuMP
model = Model(() -> Convex.Optimizer(SCS.Optimizer()))
```
where the nonlinear constraints and objective will be reformulated into conic constraints for SCS.


blegat:
No, we should do this, and we will need a layer of bridges after the Convex layer to reformulate further. Since https://github.com/jump-dev/Convex.jl/pull/504 broke the MOI wrapper of this PR, I'm first updating it there and then I'll add this

PR #513, opened 2023-10-24
remove piracy in `hcat`, `vcat`, `hvcat`

ericphanson:
closes https://github.com/jump-dev/Convex.jl/issues/512

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Convex"
 uuid = "f65535da-76fb-5f13-bab9-19810c17039a"
-version = "0.15.3"
+version = "0.16.0"
 
 [deps]
 AbstractTrees = "1520ce14-60c1-5f80-bbc7-55ef81b5835c"
--- a/docs/src/release_notes.md
+++ b/docs/src/release_notes.md
@@ -1,5 +1,10 @@
 # Release notes
 
+## v0.16.0 (October 23, 2023)
+
+* `hcat`, `vcat`, and `hvcat` with Convex.jl objects (variables, expressions, and so forth) only support up to 5 arguments.
+* Relatedly, type piracy for these functions has been removed, making Convex.jl loadable on Julia v.10.
+
 ## v0.15.3 (February 11, 2023)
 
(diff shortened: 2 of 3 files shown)


odow:
I don't know that I like this approach.

Wouldn't it be better to force `Convex.hcat` and `Convex.vcat`?

Alternatively, this might work:
```julia
Base.hcat(args::AbstractExpr...) = HcatAtom(args...)

function Base.hcat(args::AbstractExprOrValue...)
    if all(Base.Fix2(isa, Value), args)
        return Base.cat(args..., dims = Val(2))
    end
    return HcatAtom(map(arg -> convert(AbstractExpr, arg), args)...)
end
```

ericphanson:
I agree, this is unsatisfying. However: `Convex.hcat` doesn't allow use of the `[A B]` implicit hcat syntax which is often used in these problems. And I think
```julia
function Base.hcat(args::AbstractExprOrValue...)
    if all(Base.Fix2(isa, Value), args)
        return Base.cat(args..., dims = Val(2))
    end
    return HcatAtom(map(arg -> convert(AbstractExpr, arg), args)...)
end
```
is the same piracy as we are doing now, no? It's the method signature that matters for that, not what happens inside

KristofferC:
Just FYI, these type of combinatorial method definition can lead to bad load times for the package (https://github.com/JuliaDiff/ReverseDiff.jl/issues/226).

ericphanson:
Yeah, I noticed that when I initially chose `n=32` which never finished precompiling 😂. Do you have any ideas for how this can be improved?

I think the "right" fix is to be able to express "`Varags{Union{Value,AbstractExpr}}` but at least 1 `AbstractExpr`" in the type system, but I am guessing that isn't so easy to do.

odow:
> is the same piracy as we are doing now, no?

Sort of, but to a lesser extent. Since `AbstractExprOrValue` is basically `Union{Number,AbstractVector,AbstractExpr}`, it should take lower precedence than the `Union{Number,AbstractVector}` method in Base.

PR #515, opened 2023-10-24
Less piracy of `hcat` and `vcat`

### LINKED ISSUE #512: Fixing type piracy is becoming more urgent
The type piracy in 

https://github.com/jump-dev/Convex.jl/blob/c6a21df2979dc1eaf60674f7f4b1324a4f30f8b8/src/atoms/affine/stack.jl#L111-L120

causes the package to now error on the 1.10 backport branch https://s3.amazonaws.com/julialang-reports/nanosoldier/pkgeval/by_hash/4e1759c_vs_0ba6ec2/EnsembleKalmanProcesses.primary.log

```
Failed to precompile Convex [f65535da-76fb-5f13-bab9-19810c17039a] to "/home/pkgeval/.julia/compiled/v1.10/Convex/jl_CmPZEd".
WARNING: Method definition hcat(Union{Number, AbstractArray{T, N} where N where T}...) in module Base at abstractarray.jl:1995 overwritten in module Convex at /home/pkgeval/.julia/packages/Convex/tSTAW/src/atoms/affine/stack.jl:120.
ERROR: LoadError: Method overwriting is not permitted during Module precompile.
Stacktrace:
 [1] top-level scope
   @ ~/.julia/packages/Convex/tSTAW/src/atoms/affine/stack.jl:120
```

ericphanson:
Fixes #512 
Closes #513 
Closes #514 

Non-breaking alternative to #513. Since this isn't breaking, we probably should do it even if we want something like that.

### DIFF
--- a/src/atoms/affine/stack.jl
+++ b/src/atoms/affine/stack.jl
@@ -112,21 +112,25 @@ end
 
 # * `Value` is not owned by Convex.jl
 # * splatting creates zero-argument functions, which again are not owned by Convex.jl
+Base.hcat(args::AbstractExpr...) = HcatAtom(args...)
 
-hcat(args::AbstractExpr...) = HcatAtom(args...)
-function hcat(args::AbstractExprOrValue...)
+function Base.hcat(args::AbstractExprOrValue...)
+    if all(Base.Fix2(isa, Value), args)
+        return Base.cat(args..., dims = Val(2))
+    end
     return HcatAtom(map(arg -> convert(AbstractExpr, arg), args)...)
 end
-hcat(args::Value...) = Base.cat(args..., dims = Val(2))
 
 # TODO: implement vertical concatenation in a more efficient way
 vcat(args::AbstractExpr...) = transpose(HcatAtom(map(transpose, args)...))
 function vcat(args::AbstractExprOrValue...)
+    if all(Base.Fix2(isa, Value), args)
+        return Base.cat(args..., dims = Val(1))
+    end
     return transpose(
         HcatAtom(map(arg -> transpose(convert(AbstractExpr, arg)), args)...),
     )
 end
-vcat(args::Value...) = Base.cat(args..., dims = Val(1)) # Note: this makes general vcat slower for anyone using Convex...
 
 function hvcat(rows::Tuple{Vararg{Int}}, args::AbstractExprOrValue...)
     nbr = length(rows)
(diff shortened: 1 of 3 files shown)


src/atoms/affine/stack.jl
```diff
@@ -112,21 +112,25 @@ end
...
+    end
     return HcatAtom(map(arg -> convert(AbstractExpr, arg), args)...)
 end
-hcat(args::Value...) = Base.cat(args..., dims = Val(2))
 
 # TODO: implement vertical concatenation in a more efficient way
 vcat(args::AbstractExpr...) = transpose(HcatAtom(map(transpose, args)...))
 function vcat(args::AbstractExprOrValue...)
+    if all(Base.Fix2(isa, Value), args)
+        return Base.cat(args..., dims = Val(1))
```
vtjnash, line 128: It I understood correctly, it seems like this code path shouldn't be reachable, as `vcat(::Value....)` is defined in Base and is a subtype (morespecific) than `vcat(::AbstractExprOrValue...)`?

odow: I guess it guards against that method changing in Base. But yes, it's probably not reachable in the current setup.

ericphanson:
if there are no objections I will merge/tag this soon, to unbreak Convex on nightly

src/atoms/affine/stack.jl
```diff
@@ -112,21 +112,25 @@ end
 
 # * `Value` is not owned by Convex.jl
 # * splatting creates zero-argument functions, which again are not owned by Convex.jl
+Base.hcat(args::AbstractExpr...) = HcatAtom(args...)
 
-hcat(args::AbstractExpr...) = HcatAtom(args...)
-function hcat(args::AbstractExprOrValue...)
+function Base.hcat(args::AbstractExprOrValue...)
+    if all(Base.Fix2(isa, Value), args)
```
odow, line 118: Should we try removing these `if` and see if the tests break?

ericphanson: To me it feels safer to have it, re- your comment about something changing in Base. But I guess having it could mask issues if we’d rather it start erroring should something change 🤔

odow: I guess either approach is okay.

odow (approved):
LGTM

PR #517, opened 2023-11-25
Use the word `trace` in describing `tr`

ericphanson:
xref https://github.com/jump-dev/Convex.jl/issues/319#issuecomment-1826346337


ericphanson:
[preview docs](https://jump.dev/Convex.jl/previews/PR517/operations/#Linear-Program-Representable-Functions)
Screenshot 2023-11-25 at 16 28 12

PR #522, opened 2023-12-29
add test for issue 510

ericphanson:
closes #510

I believe this was fixed by #504, though I'm not sure how

PR #523, opened 2023-12-29
Error when solving non-DCP problems

ericphanson:
Noticed
```julia
WARNING: Method definition emit_dcp_warnings() in module Convex at /Users/eph/Convex.jl/src/Convex.jl:71 overwritten in module Main at /Users/eph/Convex.jl/test/test_utilities.jl:519.
WARNING: Method definition emit_dcp_warnings() in module Main at /Users/eph/Convex.jl/test/test_utilities.jl:519 overwritten at /Users/eph/Convex.jl/test/test_utilities.jl:521.
```
while running the tests for #522. Using method invalidation for this is silly as there's no reason to recompile code; checking a global ref is very fast compared to everything else Convex is doing. Also, this fix is non-breaking as the previous advice to invalidate the method will continue to work here.


odow (commented):
I have an alternative suggestion: that instead of a warning, we emit a _hard_ error. This is what CVX(py) does.

I can see so many people ignoring the warnings. Convex is for DCP. There are plenty of other options if people don't want to use DCP.

ericphanson:
That's a good idea, I've made those changes here. I've still left an opt-out in case there are rare cases in which it can be useful. I chose not to mention the opt-out in the error message since I don't want folks turning to it without understanding all the implications.

src/dcp.jl
```diff
@@ -20,10 +20,19 @@ struct AffineVexity <: Vexity end
 struct ConvexVexity <: Vexity end
 struct ConcaveVexity <: Vexity end
 
+struct DCPViolationError <: Exception end
+
+function Base.showerror(io::IO, ::DCPViolationError)
+    return print(
+        io,
+        "DCPViolationError: Expression not DCP compliant. Trying to solve non-DCP compliant problems can lead to unexpected behavior.",
```
odow, line 28: So. Based on https://ask.cvxr.com, we need this error to be veeeeeery explicit. So many people ask questions. I'l make a change

docs/src/advanced.md
```diff
@@ -5,7 +5,7 @@ DCP warnings
 ------------
```
odow, line 5: The line above this should be `DCP errors`

src/Convex.jl
```diff
@@ -57,18 +59,24 @@ export add_constraints!, maximize, minimize, Problem, satisfy, solve!
...
+Controls whether or not errors are emitted for when an expression fails to be
+of disciplined convex form. When called without an argument, returns the current setting
+for whether or not DCP errors are emitted. To turn errors off, call
 
-Controls whether or not warnings are emitted for when an expression fails to be
-of disciplined convex form. To turn warnings off, override the method via
+    Convex.allow_dcp_violations(true)
 
-    Convex.emit_dcp_warnings() = false
+to (globally) allow non-DCP problems to be solved. Note that if a problem is not DCP, the solution obtained by Convex.jl can be wildly incorrect (even if the solution status is `OPTIMAL`!).
```
odow, line 74: Is there a practical use-case for allowing this? cc @mlubin

ericphanson: I remember someone using it for alternating minimization. It could be something where they were going to fix a variable but hadn’t yet and the error occurs too early (when you construct the expression not when you solve the problem). The early errors should give more useful  stacktraces though.

src/dcp.jl
odow, line 28:
```suggestion
        "DCPViolationError: Expression not DCP compliant. This either means " *
        "that your expression is not convex, or that we could not prove it " *
        "was convex using the rules of disciplined convex programming. For a " *
        "list of supported operations, see https://jump.dev/Convex.jl/stable/operations/. " *
        "For help writing your problem as a disciplined convex program, " *
        "please post a reproducible example on https://jump.dev/forum.",
```

test/test_utilities.jl
```diff
@@ -513,13 +513,20 @@ end
...
+            # Reset unconditionally
+            Convex.allow_dcp_violations(false)
+        end
+        @test_throws DCPViolationError Convex.NotDcp()
 
-        @eval Convex.emit_dcp_warnings() = false
-        @test_logs Convex.NotDcp()
-        @eval Convex.emit_dcp_warnings() = true
-        @test_logs (:warn, r"not DCP compliant") Convex.NotDcp()
+        str = sprint(Base.showerror, DCPViolationError)
```
odow, line 528:
```suggestion
        str = sprint(Base.showerror, DCPViolationError())
```

src/Convex.jl
ericphanson: actually something weird with the current place the errors are thrown is that it is only when you call `vexity`. This happens during problem solving so we at least always throw then. But it doesn't happen when you construct the expression, unless you `show` it, which does check the vexity (to print it). So I think this isn't ideal, since you end up with different-seeming behavior in the REPL (as it shows things).

odow: What about just erroring on `optimize!`?

ericphanson: ok, I've made that changed, and removed the opt-out

docs/examples_literate/supplemental_material/Convex.jl_intro_ISMP2015.jl
```diff
@@ -21,7 +21,7 @@
 #-
 
 ## Make the Convex.jl module available
-using Convex, SparseArrays, LinearAlgebra
+using Convex, SparseArrays, LinearAlgebra, Test
```
ericphanson, line 24:
```suggestion
using Convex, SparseArrays, LinearAlgebra
```

src/dcp.jl
```diff
@@ -20,15 +20,22 @@ struct AffineVexity <: Vexity end
...
-        end
-        return new()
-    end
+struct DCPViolationError <: Exception end
+
+function Base.showerror(io::IO, ::DCPViolationError)
+    return print(
+        io,
+        "DCPViolationError: Expression not DCP compliant. This either means " *
+        "that your expression is not convex, or that we could not prove it " *
```
odow, line 29:
```suggestion
        "that your problem is not convex, or that we could not prove it " *
```

odow (approved):

src/solution.jl
```diff
@@ -52,6 +52,10 @@ function solve!(p::Problem, optimizer_factory; silent_solver = false)
         MOI.set(model, MOI.Silent(), true)
     end
 
+    if vexity(p) == NotDcp()
```
ericphanson, line 55: I think this isn’t quite right actually, we need to check for eg concave minimization problems

odow: Or at least, this is right, but there might be other cases.

PR #524, opened 2023-12-29
cleanup & bugfix `dot`

### LINKED ISSUE #508: `dot` buggy in complex case?
https://github.com/jump-dev/Convex.jl/blob/c6a21df2979dc1eaf60674f7f4b1324a4f30f8b8/src/atoms/affine/dot.jl#L13-L17

This doesn't look correct, the first argument should be complex-conjugated

ericphanson:
closes #508

I meant to just add the `conj` but noticed the `dot(*)` methods were technically piracy as well. I'm not sure we handle broadcast fusion well though, so possibly having the function barrier is helpful. But if so we should probably name it something else, or tell folks to call `broadcast(*, x, y)` directly.

As for `vecdot` it has been removed upstream since Julia 1.0.


odow (approved):
I don't know the code well enough, but this seems reasonable.

PR #519, opened 2023-12-28
Add JuMP example in README

blegat:

PR #525, opened 2023-12-29
Add JuMP example to README.md

odow:
Cherry-picked from #519

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,76 @@ problem.status # :Optimal, :Infeasible, :Unbounded etc.
 problem.optval
 ```
 
+## Using with JuMP
+
+The `master` branch of this package (not yet released) contains an experimental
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -65,6 +65,76 @@ problem.status # :Optimal, :Infeasible, :Unbounded etc.
...
 
+## Using with JuMP
+
+The `master` branch of this package (not yet released) contains an experimental
+JuMP solver. This solver reformulates a nonlinear JuMP model into a conic
+program using DCP. Note that it currently only supports scalar nonlinear
+programs. Also note that `add_bridges = false` must be used.
+
+```julia
+julia> model = Model(() -> Convex.Optimizer(Clarabel.Optimizer); add_bridges = false)
```
odow, line 76: I think we should think about how to fix this before merging.

PR #530, opened 2024-01-05
Add support for objective to MOI wrapper

blegat:
Cherry-picked from #520

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -321,7 +321,7 @@ function MOI.get(
     ci::MOI.ConstraintIndex{MOI.ScalarNonlinearFunction,S},
 ) where {S<:MOI.AbstractScalarSet}
     ret = MOI.get(model.context.model, attr, model.constraint_map[ci.value])
-    return _flip_dual(ret, S)
+    return _flip_dual(ret[], S)
 end
 
 function MOI.get(model::Optimizer, I::Type{<:MOI.Index}, name::String)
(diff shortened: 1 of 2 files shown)


src/MOI_wrapper.jl
```diff
@@ -291,6 +309,21 @@ function MOI.get(
...
 
+# See `constraints/constraints.jl`
+# `LessThan` constraints are reformulated as `rhs - lhs` unlike MOI while
+# `GreaterThan` constraints are reformulated as `lhs - rhs` like in MOI
+_flip_dual(x, ::Type{S}) where {S<:MOI.LessThan} = -x
+_flip_dual(x, ::Type{S}) where {S<:MOI.AbstractScalarSet} = x
+
+function MOI.get(
+    model::Optimizer,
+    attr::MOI.AbstractConstraintAttribute,
```
odow, line 320: Shouldn't this be `MOI.ConstraintDual` or something? Why are we negating every attribute?

src/MOI_wrapper.jl
```diff
@@ -291,6 +309,21 @@ function MOI.get(
...
+    ci::MOI.ConstraintIndex{MOI.ScalarNonlinearFunction,S},
+) where {S<:MOI.AbstractScalarSet}
+    ret = MOI.get(model.context.model, attr, model.constraint_map[ci.value])
+    return _flip_dual(ret, S)
```
odow, line 324:
```suggestion
    return _flip_dual(ret[], S)
```

src/MOI_wrapper.jl
blegat: Yes, and ConstraintPrimal

odow (approved):

PR #532, opened 2024-01-08
Refactor the tests into functional form

odow:
The tests were a varied mix of `@testset`, with global variables etc.

This PR cleans everything up.


test/test_problem_depot.jl
```diff
@@ -0,0 +1,109 @@
+module TestProblemDepot
+
+using Convex
+using Test
+
+import Clarabel
+# import ECOS
+import GLPK
```
odow, line 8:
```suggestion
import GLPK
import MathOptInterface as MOI
```

PR #534, opened 2024-01-09
Fix conversion of MOI.ScalarNonlinearFunction to Convex.AbstractExpr

### LINKED ISSUE #533: Do not use eval
https://github.com/jump-dev/Convex.jl/blob/78c0183102ed044a1c3d29b0b774700c3a46841c/src/MOI_wrapper.jl#L141

The correct approach is to use a `Nonlinear.OperatorRegistry`.

odow:
Closes #533 

I think for now that it's safest to explicitly opt-in to the functions that Convex supports.


test/MOI_wrapper.jl
```diff
@@ -42,6 +42,49 @@ function test_runtests()
...
+        1.0*x*x=>1.0*y*y,
+        1.0*x*x+2.0*x+3.0=>1.0*y*y+2.0*y+3.0,
+        MOI.ScalarNonlinearFunction(:+, Any[x])=>y,
+        MOI.ScalarNonlinearFunction(:+, Any[x, 2])=>+(y, 2),
+        MOI.ScalarNonlinearFunction(:+, Any[x, x, x])=>sum([y, y, y]),
+        MOI.ScalarNonlinearFunction(:-, Any[x])=>-y,
+        MOI.ScalarNonlinearFunction(:-, Any[x, 2])=>-(y, 2),
+        MOI.ScalarNonlinearFunction(:*, Any[x, 2])=>*(y, 2),
+        MOI.ScalarNonlinearFunction(:/, Any[x, 2])=>y/2,
+        MOI.ScalarNonlinearFunction(:^, Any[x, 2])=>square(y),
```
odow, line 63:
```suggestion
        MOI.ScalarNonlinearFunction(:^, Any[x, 2])=>Convex.square(y),
```

blegat (approved):
Indeed, explicit opt-in is cleaner, everything is in the same repo so if Convex.jl expressions supports something new, we can add it here at the same time

PR #537, opened 2024-01-09
Refactor imports to explicitly overload methods

odow:
There were lots of imports scattered around the code that make it hard to know what functions were defined by Convex, and what were overloaded. One example is `Base.sign`, which was imported in `src/expressions.jl`. 

The exports, e.g., from LinearAlgebra, are retained.<s>, so the only breaking part about this is `Convex.sign` to `Base.sign`, and we can't fix this by exporting because it's a `Base` method.</s>

### DIFF
--- a/src/Convex.jl
+++ b/src/Convex.jl
@@ -64,8 +64,10 @@ for k in (
     :norm2,
     :tr,
 )
-    @eval $k = LinearAlgebra.$k
-    @eval export $k
+    @eval begin
+        using LinearAlgebra: $k
+        export $k
+    end
 end
 
 # Constraints
--- a/src/Convex.jl
+++ b/src/Convex.jl
@@ -1,42 +1,61 @@
-__precompile__()
-
 module Convex
-using OrderedCollections: OrderedDict
-using LinearAlgebra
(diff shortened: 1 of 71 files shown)


src/Convex.jl
```diff
@@ -1,62 +1,115 @@
...
+    :norm2,
+    :tr,
+)
+    @eval begin
+        using LinearAlgebra: $k
+        export $k
+    end
+end
+using AbstractTrees: children
+using Base: sign
```
blegat, line 112: Why is this line needed ? You explicitly do `Base.sign`.

odow: Because docs and tests assume that there is `Convex.sign`. Same for `Convex.children`. I didn't add a comment because this is tested.

blegat: Ah ok, so we need it for backward compatibility

odow: Yeah. Before tagging the next release, there are a bunch of deprecation things we could remove. This would be one candidate.

odow: But that's for a different PR.

PR #538, opened 2024-01-10
Tidy src/atoms/affine

odow:
Just some small formatting changes, mainly so I have an excuse to read through and understand the code.

### DIFF
--- a/src/atoms/affine/add_subtract.jl
+++ b/src/atoms/affine/add_subtract.jl
@@ -1,53 +1,30 @@
-#############################################################################
-# add_subtract.jl
-# Handles unary negation, addition and subtraction of variables, constants
-# and expressions.
-# All expressions and atoms are subtpyes of AbstractExpr.
-# Please read expressions.jl first.
-#############################################################################
-
-### Unary Negation
-
 mutable struct NegateAtom <: AbstractExpr
     children::Tuple{AbstractExpr}
     size::Tuple{Int,Int}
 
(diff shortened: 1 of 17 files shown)


src/atoms/affine/real_imag.jl
```diff
@@ -1,86 +1,64 @@
...
-        return new(children, x.size)
-    end
+    ImaginaryAtom(x::AbstractExpr) = new((x,), x.size)
 end
 
 head(io::IO, ::ImaginaryAtom) = print(io, "imag")
 
 function Base.sign(x::ImaginaryAtom)
+    # FIXME(odow): what is this?
     sign(x.children[1]) == ComplexSign()
```
odow, line 45: Follow up later: I don't think this method is correct.

src/atoms/affine/add_subtract.jl
```diff
@@ -1,53 +1,30 @@
-#############################################################################
-# add_subtract.jl
-# Handles unary negation, addition and subtraction of variables, constants
-# and expressions.
-# All expressions and atoms are subtpyes of AbstractExpr.
```
blegat, line 5: Is this not useful or has it be moved somewhere ?

odow: I didn't think it was useful. It's pretty obvious that `NegateAtom` handles negation. And we really need a docstring for each atom with examples, rather than this vague comment in the file. But that's a separate PR...

PR #549, opened 2024-01-11
Rename src/atom files for clarity

odow:
So I got quite confused reading through the atom code recently, because there is a mix of atom definitions, and then files like
https://github.com/jump-dev/Convex.jl/blob/2bb674e48a16a01c438cce92a44f40d87990f95d/src/atoms/affine/trace.jl#L1
which implement reformulations.

There were also files that contained multiple atoms.


blegat (approved):

odow:
Thoughts on moving the reformulations to a `src/reformulations` directory? I think it'd be useful to separate the atoms which require various methods etc, from the operators which build other atoms.

ericphanson:
One thing to note is that some of these used to be atoms but got rewritten as formulations, but left in the same file to minimize the diff. Eg https://github.com/jump-dev/Convex.jl/pull/284. But I think it’s fine to ask that such changes move the code to a different file.

PR #550, opened 2024-01-11
Tidy atoms/sdp_cone

odow:
Okay, so the `sdp_cone` directory had a mix of atoms, reformulations, and cones?

There are also a bunch of atoms that do not end with "Atom". We should probably rename them while we're making a breaking release?


ericphanson:
Changing them is non-breaking as they are implementation details. Feel free to rename though!

chriscoey:
nit....can we rename sdp cone to psd cone?

odow:
Done. As you can probably tell, I'm starting at the outside and working my way into the core of Convex.jl. If we have good tests and documentation for all the user-facing atoms, then we should be able to make more changes to the core with confidence.

odow:
> nit....can we rename sdp cone to psd cone?

Hahahahaha. Let's leave that for a separate PR.

PR #551, opened 2024-01-12
remove internal function `latex_formulation`

ericphanson:
I think I initially was using this code for debugging at some point in #504, but it was untested and bitrotted. Here I've added a test, I guess we could also export it and add it to the release notes. Alternatively we could delete it.


odow:
I think I'd prefer to delete. One useful utility might be `backend(p::Problem, optimizer)` which returns the MOI model. Then people are free to do what they like with it.

odow:
I think, where possible, we should try to minimize the API surface of Convex.jl.

odow (approved):

PR #553, opened 2024-01-12
Refactor and rename src/constraints

odow:
Since these are implementation details (they are not exported), I'm being a bit ruthless with renaming for clarity.

One question for @ericphanson: it's not clear to me why some atoms call `MOI_add_constraint`, and others call a `<: Constraint`. Is it because we can only return duals via `<: Constraint`?


ericphanson:
Yeah, we only have some constraints modeled as objects. Those the user can construct with operator-overloading syntax, and they support `show` and duals (for the most part). However the MOI model only needs an MOI constraint so if we don’t have an object defined then we sometimes just add the constraint directly instead.

PR #555, opened 2024-01-12
Deprecate strict inequalities

### LINKED ISSUE #552: Convex treats strict inequalities and non-strict
See 

https://github.com/jump-dev/Convex.jl/blob/32b487dbb0cf33ca758ca4a2c3a7ec008f42ce39/src/constraints/constraints.jl#L214-L218

```Julia
julia> using Convex

julia> x = Variable()
Variable
size: (1, 1)
sign: real
vexity: affine
id: 179…950

julia> x > 0
≥ constraint (affine)
├─ real variable (id: 179…950)
└─ [0;;]

julia> x >= 0
≥ constraint (affine)
├─ real variable (id: 179…950)
└─ [0;;]
```

This seems like a foot-gun that we should remove.

odow:
Closes #552

### DIFF
--- a/src/deprecations.jl
+++ b/src/deprecations.jl
@@ -1,12 +1,29 @@
 # TODO(odow): Remove code in a future release
-function Base.:<(::AbstractExpr, ::AbstractExpr)
-    return error("Strict inequality `<` has been removed. Use `<=` instead.")
+
+function Base.:<(x::AbstractExpr, y::AbstractExpr)
+    @warn(
+        "Strict inequality operator `<` has never enforced the strict " *
+        "inequality, and instead adds the non-strict inequality `<=`. This " *
+        "operator will be removed in the next release.",
+        maxlog = 1,
+    )
+    return x <= y
 end
+
 Base.:<(lhs::AbstractExpr, rhs::Value) = <(lhs, constant(rhs))
+
 Base.:<(lhs::Value, rhs::AbstractExpr) = <(constant(lhs), rhs)
 
-function Base.:>(::AbstractExpr, ::AbstractExpr)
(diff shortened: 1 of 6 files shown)


test/test_utilities.jl
```diff
@@ -936,6 +936,23 @@ function test_write_to_file()
     return
 end
 
+function test_strict_inequality_deprecation()
+    x = VariableIndex()
```
odow, line 940:
```suggestion
    x = Variable()
```

src/deprecations.jl
```diff
@@ -0,0 +1,12 @@
+# TODO(odow): Remove code in a future release
+function Base.:<(::AbstractExpr, ::AbstractExpr)
+    return error("Strict inequality `<` has been removed. Use `<=` instead.")
+end
+Base.:<(lhs::AbstractExpr, rhs::Value) = <(lhs, constant(rhs))
+Base.:<(lhs::Value, rhs::AbstractExpr) = <(constant(lhs), rhs)
+
+function Base.:>(::AbstractExpr, ::AbstractExpr)
+    return error("Strict inequality `>` has been removed. Use `>` instead.")
```
odow, line 9:
```suggestion
    return error("Strict inequality `>` has been removed. Use `>=` instead.")
```

src/deprecations.jl
```diff
@@ -0,0 +1,12 @@
+# TODO(odow): Remove code in a future release
+function Base.:<(::AbstractExpr, ::AbstractExpr)
+    return error("Strict inequality `<` has been removed. Use `<=` instead.")
```
blegat, line 3: Maybe we can warn instead of error and then in the next release we remove

Issue #542, opened 2024-01-10
Unit tests for atoms


odow:
I have been going through the atom code, #538, #539, #540, #541, and I noticed that the coverage for these files is relatively low. We don't seem to have any specific unit tests, everything is just tested end-to-end in `problem_depot`.

The coverage was also hiding a number of obvious bugs https://github.com/jump-dev/Convex.jl/pull/541#discussion_r1446831539, https://github.com/jump-dev/Convex.jl/pull/538#discussion_r1446831944.

We should think about how to test each atom, like we do the bridges in MOI.

odow:
I've made a start in https://github.com/jump-dev/Convex.jl/pull/546

odow:
#556 is coming along nicely.

odow:
Closing because there's still a few PSD atoms to go, but the majority are done.

PR #558, opened 2024-01-15
Add tests for src/atoms/exp

odow:
Part of #542


src/atoms/exp_cone/EntropyAtom.jl
```diff
@@ -21,7 +23,7 @@ curvature(::EntropyAtom) = ConcaveVexity()
 
 function evaluate(x::EntropyAtom)
     c = evaluate(x.children[1])
-    return -c .* log.(c)
+    return -sum(c .* log.(c))
```
odow, line 26: Careful tests continue to find bugs :smile:

PR #561, opened 2024-01-15
Add tests for atoms/second_order_cone

odow:
Part of #542


src/variable.jl
```diff
@@ -102,9 +102,10 @@ sign!(x::AbstractVariable, s::Sign) = x.sign = s
 Returns the current value of `x` if assigned; errors otherwise.
 """
 function evaluate(x::AbstractVariable)
-    return _value(x) === nothing ?
-           error("Value of the variable is yet to be calculated") :
-           output(_value(x))
+    if _value(x) === nothing
+        error("Value of the variable is yet to be calculated")
+    end
+    return output(copy(_value(x)))
```
odow, line 108: I think a copy is needed here? It'd be weird if modifying the result of `evaluate` modified the `.value` of a variable. (Which is what used to happen in the `huber` example.)

PR #574, opened 2024-01-18
[docs] refactor and re-organize the documentation

odow:


docs/src/introduction/faq.md
```diff
@@ -0,0 +1,69 @@
...
+
+## How does Convex.jl differ from JuMP?
+
+Convex.jl and JuMP are both modelling languages for mathematical programming
+embedded in Julia, and both interface with solvers via MathOptInterface, so many
+of the same solvers are available in both.
+
+Convex.jl converts problems to a standard conic form. This approach requires
+(and certifies) that the problem is convex and DCP compliant, and guarantees
+global optimality of the resulting solution.
```
ericphanson, line 18: "guarantees" might be a little strong. For example, the second example of @araujoms's nice blog post https://mateusaraujo.info/2023/02/08/sdps-are-not-cheat-codes/ by can be reproduced in Convex easily:

```julia
julia> using Convex, SCS

julia> X = HermitianSemidefinite(3, 3)
Variable
size: (3, 3)
sign: complex
vexity: affine
id: 176…498

julia> p = minimize(real(X[2,2]), X ⪰ 0, X[3,3] == 0, X[2,2] + 2*X[1,3] == 1)
minimize
└─ real (affine; real)
   └─ index (affine; complex)
      └─ 3×3 complex variable (id: 176…498)
subject to
├─ sdp constraint (affine)
│  └─ reshape (affine; real)
│     └─ * (affine; real)
│        ├─ …
│        └─ …
├─ == constraint (affine)
│  ├─ index (affine; complex)
│  │  └─ 3×3 complex variable (id: 176…498)
│  └─ [0;;]
└─ == constraint (affine)
   ├─ + (affine; complex)
   │  ├─ index (affine; complex)
   │  │  └─ …
   │  └─ * (affine; complex)
   │     ├─ …
   │     └─ …
   └─ [1;;]

julia> solve!(p, SCS.Optimizer; silent_solver=true)

julia> p.optval
0.14584807577454767

julia> p.status
OPTIMAL::TerminationStatusCode = 1
```
where as the blog post says, the correct solution is 1. BTW, Clarabel is also pretty far off:
```julia
julia> solve!(p, Clarabel.Optimizer; silent_solver=true)

julia> p.status
OPTIMAL::TerminationStatusCode = 1

julia> p.optval
0.8844497345252251
```
so it's not just SCS.

Maybe:
```suggestion
(and certifies) that the problem is convex and DCP compliant, and guarantees
global optimality of the resulting solution for many classes of problem.
```
or something like that.

docs/src/introduction/faq.md
```diff
@@ -0,0 +1,69 @@
+# FAQ
+
+## Where can I get help?
+
+For usage questions, please contact us via the
+[Julia Discourse](https://discourse.julialang.org/c/domain/opt). If you're
+running into bugs or have feature requests, please use the [GitHub Issue
+Tracker](https://github.com/jump-dev/Convex.jl/issues).
+
+## How does Convex.jl differ from JuMP?
```
ericphanson, line 10: I think it would make sense to mention the reformulations here. There are some very sophisticated ones ported from https://github.com/hfawzi/cvxquad that can be quite valuable for quantum information theory researchers ([which even cvxpy doesn't have yet](https://ask.cvxr.com/t/can-cvxquad-be-applied-in-python/12277/2)), and having reformulations in general can make some problems simpler to write out

odow: I've tried to make very minimal changes to the content of the docs for now. I've added this to https://github.com/jump-dev/Convex.jl/issues/344#issuecomment-1899369803

docs/src/introduction/faq.md
```diff
@@ -0,0 +1,70 @@
...
+
+Convex.jl and JuMP are both modelling languages for mathematical programming
+embedded in Julia, and both interface with solvers via MathOptInterface, so many
+of the same solvers are available in both.
+
+Convex.jl converts problems to a standard conic form. This approach requires
+(and certifies) that the problem is convex and DCP compliant, and guarantees
+global optimality of the resulting solution (if the solver succeeds. For some
+models, the solver may experience numerical difficulty).
```
ericphanson, line 19: this is definitely better, but note both solutions said they were "OPTIMAL". I'm not really sure it's just a numeric issue as much as an algorithmic issue; e.g. https://optimization-online.org/wp-content/uploads/2018/07/6695.pdf shows SDPA-GMP failing on some SDP's with a duality gap. Although I've seen SDPA-GMP fail easily on many problems without sufficient presolve so maybe that is a factor there too.

docs/src/introduction/faq.md
ericphanson: ah somehow I thought this section was new! makes sense

docs/src/introduction/faq.md
araujoms: Thanks for the citing my blog! I second this comment: I had in fact a similar issue with MOSEK: the problem was not strictly feasible, and as expected MOSEK failed to solve it, but then falsely claimed that it was "OPTIMAL". I submitted a [bug report](https://groups.google.com/g/mosek/c/mkYUzjmJY5U/m/vAFs_IUPCgAJ) to MOSEK but only managed to get insulted by the CEO.

araujoms: As for the proof, there's a simpler argument: a matrix `A` is positive semidefinite iff `A = B'*B` for some matrix `B`. Now if `A` has a diagonal element equal to zero, then `v'*A*v = 0` for some basis vector `v`. But  `v'*A*v = v'*B'*B*v = norm(B*v)^2`, so `B*v = 0`, which implies `B'*B*v = A*v = 0`, so the entire column must be zero.

ericphanson: ah, that is a nice argument. It may be worth adding that to your blog post, or at least noting that the entire column is zero (and therefore X13 = 0) rather than jumping to X13=0, since providing the intermediate step can help readers figure it out (or figure out what to look up).

araujoms: Fair enough, I added it to the blog post (as a footnote).

odow: I don't want to derail this PR with a discussion, so I've added a TODO to https://github.com/jump-dev/Convex.jl/issues/344. Once this PR is merged I'm happy to review other changes.

odow: You could also argue that "if the solver succeeds" is true, and that solvers reporting `OPTIMAL` are a bug. That's their problem, not ours.

PR #576, opened 2024-01-19
[docs] move examples to docs/src and don't create notebooks

odow:
I've never had anyone ask for notebooks of the JuMP docs.


odow:
Thoughts @ericphanson?

blegat:
> I've never had anyone ask for notebooks of the JuMP docs.

I used to build them locally when I needed them but now I prefer running `.jl` files with VS code than jupyter notebooks so I don't need the notebooks anymore. This is maybe a trend that's common which explains why people tend to not request notebooks

ericphanson:
I’m fine with dropping them, I think keeping this as maintainable as possible is more important. I do think a few people used them, but one can also always build your own notebook by copying examples manually into jupyter, which isn’t a bad way to learn imo.

ericphanson (approved):

blegat (approved):

PR #578, opened 2024-01-19
Deprecate in(x, ::Symbol) to create PSD constraint

odow:


test/test_utilities.jl
```diff
@@ -958,6 +958,14 @@ function test_deprecation_norm()
     return
 end
 
+function test_deprecation_isposdef()
+    x = Variable(2, 2)
+    @test_logs (:warn,) LinearAlgebra.isposdef(x)
+    c = LinearAlgebra.isposdef(x)
+    @test c isa Constraint.PositiveSemidefiniteConstraint
```
odow, line 965:
```suggestion
    @test c isa Convex.PositiveSemidefiniteConstraint
```

test/test_utilities.jl
```diff
@@ -958,6 +958,14 @@ function test_deprecation_norm()
     return
 end
 
+function test_deprecation_isposdef()
+    x = Variable(2, 2)
+    @test_logs (:warn,) LinearAlgebra.isposdef(x)
+    c = LinearAlgebra.isposdef(x)
+    @test c isa Convex.PositiveSemidefiniteConstraint
```
odow, line 965:
```suggestion
    @test c isa Convex.PositiveSemidefiniteConeConstraint
```

src/deprecations.jl
```diff
@@ -37,3 +37,12 @@ Base.:>(lhs::Value, rhs::AbstractExpr) = >(constant(lhs), rhs)
 @deprecate get_vectorized_size(x::AbstractExpr) length(x)
 
 @deprecate operatornorm(x::AbstractExpr) LinearAlgebra.opnorm(x)
+
+function LinearAlgebra.isposdef(x::AbstractExpr)
+    @warn(
+        "Using isposdef(x) to construct a semidefinite constraint is " *
+        "deprecated. Use `x in :semidefinite` or `x ⪰ 0` instead.",
```
blegat, line 44:
```suggestion
        "deprecated. Use `x in :PSD` or `x ⪰ 0` instead.",
```

src/constraints/PositiveSemidefiniteConeConstraint.jl
```diff
@@ -57,10 +57,6 @@ function populate_dual!(
     return
 end
 
-# TODO: Remove isposdef, change tests to use in. Update documentation and
-# notebooks
-LinearAlgebra.isposdef(x::AbstractExpr) = in(x, :SDP)
-
```
blegat, line 63: Actually, I prefer `isposdef` than `x in :semidefinite`, maybe we can deprecate the `in :SDP` instead ?

odow: I don't really have an opinion. I'm just following the comment :smile:

`isposdef` seems more 'convex'.

blegat: Ah yes indeed, I didn't see that comment. Indeed, the `Symbol` approach isn't particularly convenient since the typos in the symbol aren't checked at compile time (this is why we switched to `Enum` in JuMP v0.19). Also, this is actually the only symbol supported, it's not like other symbols are supported so it's kind of weird.

blegat: This seems to date back to @karanveerm's PR in 2015 https://github.com/jump-dev/Convex.jl/pull/50. I think we can keep `isposdef` for now

odow: I'm happy to swap to deprecating the `in` syntax. we don't need three ways to specify a PSD matrix.

test/test_utilities.jl
```diff
@@ -958,6 +958,13 @@ function test_deprecation_norm()
     return
 end
 
+function test_deprecation_in_symbol()
+    x = Variable(2, 2)
+    @test_logs (:warn,) in(x, :SDP)
```
odow, line 963:
```suggestion
    @test_logs (:warn,) (x in :SDP)
```

docs/src/manual/complex-domain_optimization.md
```diff
@@ -95,9 +95,11 @@ using Convex, SCS
 A = [ 0.47213595 0.11469794+0.48586827im; 0.11469794-0.48586827im  0.52786405]
 B = ComplexVariable(2, 2)
 ρ = kron(A, B)
-constraints = [partialtrace(ρ, 1, [2; 2]) == [1 0; 0 0]
-               tr(ρ) == 1
-               ρ in :SDP]
+constraints = [
+    partialtrace(ρ, 1, [2; 2]) == [1 0; 0 0],
+    tr(ρ) == 1
```
odow, line 100:
```suggestion
    tr(ρ) == 1,
```

blegat (approved):

PR #582, opened 2024-01-22
Tidy src/solution.jl

odow:
@ericphanson I have some questions about this one. I'll comment in-line.

### DIFF
--- a/src/solution.jl
+++ b/src/solution.jl
@@ -1,13 +1,10 @@
 function add_variables!(model, var::AbstractVariable)
-    var.id_hash == objectid(:constant) &&
-        error("Internal error: constant used as variable")
-    return if sign(var) == ComplexSign()
+    if sign(var) == ComplexSign()
         L = MOI.add_variables(model, length(var))
         R = MOI.add_variables(model, length(var))
         return (L, R)
-    else
-        return MOI.add_variables(model, length(var))
     end
+    return MOI.add_variables(model, length(var))
 end
 
 scalar_fn(x::Number) = x # for `satisfy` problems? Not sure...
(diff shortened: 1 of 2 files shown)


src/solution.jl
```diff
@@ -1,13 +1,10 @@
 function add_variables!(model, var::AbstractVariable)
-    var.id_hash == objectid(:constant) &&
-        error("Internal error: constant used as variable")
```
odow, line 3: No idea why this is here. Safe to remove? The comparison seems arbitrary and unrelated to the actual `id_hash` of a constant.

src/solution.jl
```diff
@@ -16,113 +13,74 @@ scalar_fn(x::SparseTape) = scalar_fn(to_vaf(x))
...
-                )
-            else
-                vectorized_value =
-                    MOI.get(model, MOI.VariablePrimal(), var_indices)
-                set_value!(
-                    var,
-                    unpackvec(
-                        vectorized_value,
-                        size(var),
-                        iscomplex(sign(var)),
```
odow, line 83: Okay here's a tricky one. This branch occurs if `var_indices` is not a tuple, which means it must be a `Vector{MOI.VariableIndex}`. If so, in what situation is `sign(var)` complex? Because the `add_variables!` method adds a tuple if the sign is complex.

Is the `AbstractVariable` API intended to be extensible by third-party packages? Is it used? Are there any examples that we should be checking for breakage?

src/solution.jl
```diff
@@ -16,113 +13,74 @@ scalar_fn(x::SparseTape) = scalar_fn(to_vaf(x))
...
         end
     end
-
-    return nothing
+    return
 end
 
-function populate_dual!(model::MOI.ModelLike, constr::Constraint, ::Nothing)
-    constr.dual = nothing
-    return nothing
```
odow, line 112: Ditto here: for this method to be called, the element in `.constr_to_moi_inds` would need to be `nothing`, but this can't happen with the current code base. Is it needed for third-parties?

src/solution.jl
ericphanson: Uhh, I think if it wasn’t complex we’d get an error anyway inside unpackvec since the sizes wouldn’t work. So seems safe to say it must be real.

we do support custom AbstractVariables, but afaict no one has ever done that. I do have tests to ensure the api is working and I think we do solve a problem there. Maybe only in the real case though

src/solution.jl
ericphanson: I think this is from when there were duals we didn’t populate. Maybe still true for some constraints?

src/solution.jl
ericphanson: I’ll try to take a closer look at these tomorrow

src/solution.jl
odow: Ahhhhh. Okay. That's cause I'm still missing a few tests for `src/constraints/RelativeEntropyEpiCone` and friends. And the current tests only add these indirectly via `QuantumEntropyAtom` etc, so there are no tests for populating duals.

Thanks for the insight. I'm slowly learning the code.

odow: The good news is that we're up to 98% coverage :smile:

src/solution.jl
ericphanson: Ok, I had another look, and I think the diff here is fine if the tests pass, but I don't 100% remember why it was written like this. We do test [3 kinds of custom AbstractVariables](https://github.com/jump-dev/Convex.jl/blob/946d18d131a3404c984725fbd05cae1bf38d0bc4/test/test_utilities.jl#L744-L917) so I think that should cover us pretty well.

BTW I had bigger plans for the AbstractVariable stuff to build higher level functionality / DSLs on top of Convex such as for quantum information theory, so that's where that was going, but I think the real useful feature is adding constraints to a `Variable` itself (so you have a function like `make_probability_vector(d)` that doesn't need the model as an input and the result of which can be reused between models) and the full AbstractVariable thing isn't necessarily much more useful than that.

odow: Yes, I've been mulling over https://github.com/jump-dev/Convex.jl/issues/310

PR #583, opened 2024-01-23
Tidy src/variable.jl

odow:
@ericphanson thoughts on `:Bin` vs `BinVar`? I assume your long-term plan was to deprecate the symbols? They're pretty brittle and the docs don't mention the Symbol approach.

### DIFF
--- a/src/problem_depot/problems/mip.jl
+++ b/src/problem_depot/problems/mip.jl
@@ -121,7 +121,7 @@ end
         @test p.optval ≈ 2 atol = atol rtol = rtol
     end
 
-    x = Variable(2, :BinVar)
+    x = Variable(2, BinVar)
     p = minimize(sum(x), x[1] >= 0.5, x >= 0; numeric_type = T)
 
     if test
--- a/src/problem_depot/problems/mip.jl
+++ b/src/problem_depot/problems/mip.jl
@@ -46,7 +46,7 @@ end
         @test p.optval ≈ 5 atol = atol rtol = rtol
     end
 
-    x = Variable(2, :Int)
+    x = Variable(2, IntVar)
     p = minimize(sum(x), x >= 4.3; numeric_type = T)
 
     if test
@@ -69,7 +69,7 @@ end
         @test p.optval ≈ 12 atol = atol rtol = rtol
     end
 
-    x = Variable(2, :Int)
+    x = Variable(2, IntVar)
     p = minimize(norm(x, 1), x[1] >= 4.3; numeric_type = T)
 
     if test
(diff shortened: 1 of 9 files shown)


ericphanson:
Yeah, that was the plan. After using it intermittently though I don’t really like the *Var suffixes; they are hard to remember and look a bit weird. I’m definitely open to alternative approaches!

odow:
What about `MOI.ZeroOne` and `MOI.Integer`, or the `Base` types `Int` and `Bool`?

ericphanson:
Either of those seems better! The MOI ones might help folks connect the dots with MOI more, while the Base ones might look less scary. Both could lead to the issue of folks trying other objects that aren’t supported, but we could give good errors.

I see JuMP supports `Int`, `Bin`, and `binary=true`, so that could also make sense here.

ericphanson:
Btw now that JuMP supports generic number types, it seems ok to lower to jump rather than MOI- maybe that could allow better interop, like mixing Convex and JuMP variables in a model. I bring this up here since I suspect `Bin` is from JuMP, not MOI.

odow:
I think we want to keep lowering to MOI instead of JuMP. We have some ideas for DCP in JuMP. One is to lower JuMP's nonlinear expressions to Convex (requires https://github.com/jump-dev/MathOptInterface.jl/issues/2402). 

The other is to add DCP support to JuMP directly. Working through this code base is giving me some ideas. But this is a long-term thing. Nothing in the short-term.

src/problem_depot/problems/mip.jl
```diff
@@ -121,7 +121,7 @@ end
         @test p.optval ≈ 2 atol = atol rtol = rtol
     end
 
-    x = Variable(2, :Bin)
+    x = Variable(2, :BinVar)
```
odow, line 124:
```suggestion
    x = Variable(2, BinVar)
```

test/test_utilities.jl
```diff
@@ -1205,6 +1205,48 @@ function test_scalar_fn_objective_square()
...
+    x = ComplexVariable()
+    set_value!(x, 1.0)
+    @test x.value == 1.0 + 0.0im
+    y = ComplexVariable(2)
+    set_value!(y, [1.0, 2.0])
+    @test y.value == [1.0 + 0.0im, 2.0 + 0.0im]
+    return
+end
+
+
```
odow, line 1249:
```suggestion
```

blegat:
> Btw now that JuMP supports generic number types, it seems ok to lower to jump rather than MOI- maybe that could allow better interop, like mixing Convex and JuMP variables in a model

You can already mix any constraints supported by JuMP with DCP constraints when using `Convex.Optimizer`, what would using Convex at the JuMP level achieve ? Mixing the Convex expressions and the JuMP expressions is tricky, it was the approach of ParameterJuMP but we're now going the other direction with ParametricOptInterface. I feel staying at the MOI level where you only need to care about the MOI API and not arbitrary interaction with the user is easier :)

blegat (approved):

PR #584, opened 2024-01-29
Add a second cache for bridged variables

### LINKED ISSUE #564: Double bridging with MOI layers
Convex.Optimizer add bridges to its inner optimizer because bridges might be needed to transformed the conic problem it generates. Moreover, `Convex.Optimizer` returns `true` for `supports_incremental_interface` since otherwise `default_copy_to` would error:
https://github.com/jump-dev/MathOptInterface.jl/blob/ede6d1cb7a3e233de5cd39a04ca888da4acd2163/src/Utilities/copy.jl#L479-L482
Because of this, after `MOI.instantiate`, we get `MOI.LazyBridgeOptimizer(Convex.Optimizer(MOI.LazyBridgeOptimizer(Clarabel.Optimizer)))` and this creates issues because bridges use negative indices to distinguish its indices but if we have two layers of bridges without a caching optimizer in between that's not working.
A workaround would be for `Convex.Optimizer` to return `false` for `default_copy_to` but the error in `default_copy_to` prevents this.
I think the issue is kind of like https://github.com/jump-dev/MathOptInterface.jl/issues/2361 where these trait should really just be "does `instantiate` need to add a cache layer if it adds a bridge layer ?" and if we remove these error then it can be just that.

This is a recurring issue with MOI layers. For instance, DiffOpt also adds an internal bridge layers and uses `default_copy_to` but it does it by directly copying to the internal optimizer thus avoiding the need to define `supports_incremental_interface` even if it should be true. I expect that ParametricOptInterface should have the same issue.

odow:
Closes #564

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -12,7 +12,7 @@ struct Optimizer{T,M} <: MOI.AbstractOptimizer
         )
     end
     function Optimizer{T}(optimizer_constructor) where {T}
-        return Optimizer(Context{T}(optimizer_constructor))
+        return Optimizer(Context{T}(optimizer_constructor; add_cache = true))
     end
 end
 
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -12,6 +12,7 @@ struct Optimizer{T,M} <: MOI.AbstractOptimizer
         )
     end
     function Optimizer{T}(optimizer_constructor) where {T}
+        # See https://github.com/jump-dev/Convex.jl/issues/564 for details
         return Optimizer(Context{T}(optimizer_constructor; add_cache = true))
     end
 end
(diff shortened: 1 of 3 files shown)


src/MOI_wrapper.jl
```diff
@@ -12,7 +12,7 @@ struct Optimizer{T,M} <: MOI.AbstractOptimizer
         )
     end
     function Optimizer{T}(optimizer_constructor) where {T}
-        return Optimizer(Context{T}(optimizer_constructor))
+        return Optimizer(Context{T}(optimizer_constructor; add_cache = true))
```
blegat, line 15: Can you add a reference to the MOI issue ? I'm not sure I will remember if I read it a year from now

odow, line 15:
```suggestion
        # See https://github.com/jump-dev/Convex.jl/issues/564 for details
        return Optimizer(Context{T}(optimizer_constructor; add_cache = true))
```

PR #586, opened 2024-04-02
Fix quadform with fixed variables that are modified between solves

### LINKED ISSUE #585: Sequentially solve a quadratic programming problem
Related to https://github.com/jump-dev/Convex.jl/issues/398

I'm trying to solve a quadratic problem sequentially but only the first problem gets solved. 

If I run the above code from [398](https://github.com/jump-dev/Convex.jl/issues/398)

    n=5
    b = randn(n) 
    x = Variable(n)
    H=Semidefinite(n)
    Hval=randn(n,n)
    Hval.=Hval'*Hval+10*diagm(ones(n)) # symmetric positive definite
    fix!(H,Hval)
    problem = minimize(x'b +quadform(x,H), [x >= 0])
    solve!(problem, SCS.Optimizer)

The solution is 

     julia> x.value
     5×1 Matrix{Float64}:
     -1.1415298336189401e-5
     0.051775607935912626
     -2.338104123291265e-5
     -1.764813180548554e-5
     -1.8960736530572698e-5

Then I create a new H matrix and fix! it and solve again 

    Hval2=randn(n,n)
    Hval2.=Hval2'*Hval2+10*diagm(ones(n)) # symmetric positive definite
    fix!(H,Hval2)
    solve!(problem, SCS.Optimizer)

But I get the exact same solution (below) as the first problem. Seems like the Hval2 value was ignored. I noticed this on versions 0.15.4 and the master branch (v0.16). Have I missed something or is this a bug? 

     julia> x.value
    5×1 Matrix{Float64}:
    -1.1415298336189401e-5
    0.051775607935912626
    -2.338104123291265e-5
    -1.764813180548554e-5
    -1.8960736530572698e-5

odow:
Closes #585 

Introduced by https://github.com/jump-dev/Convex.jl/pull/399

@ericphanson does this make sense?

### DIFF
--- a/src/problem_depot/problems/socp.jl
+++ b/src/problem_depot/problems/socp.jl
@@ -376,7 +376,7 @@ end
     Hval = randn(n, n)
     Hval .= Hval' * Hval + 10 * diagm(0 => ones(n)) # symmetric positive definite
     fix!(H, Hval)
-    p = minimize(x'b + quadform(x, H), [x >= 0]; numeric_type = T)
+    p = minimize(x'b + quadform(x, evaluate(H)), [x >= 0]; numeric_type = T)
     handle_problem!(p)
 
     p2 = minimize(x'b + quadform(x, Hval), [x >= 0]; numeric_type = T)
--- a/src/reformulations/quadform.jl
+++ b/src/reformulations/quadform.jl
@@ -49,6 +49,18 @@ function _is_psd(
     end
 end
 
+"""
+    quadform(x::AbstractExpr, A::AbstractExpr; assume_psd=false)
+
+Represents `x' * A * x` where either:
+
+ * `x` is a vector-valued variable and `A` is a positive semidefinite or
+   negative semidefinite matrix (and in particular Hermitian or real symmetric).
+   If `assume_psd=true`, then `A` will be assumed to be positive semidefinite.
+   Otherwise, `Convex._is_psd` will be used to check if `A` is positive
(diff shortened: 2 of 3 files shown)


ericphanson:
thank you for digging into this! 

Ooph. I think I had forgotten about how these expressions happen at construction time, while only `conic_form` gets re-done at solve time each time. This makes me wonder if that mistake is made elsewhere. For a bit of time I had converted some atoms to non-atom expressions for simplicity, but now I wonder if that was very misguided, since atoms are the "symbolic" problem-construction-stage representation that get materialized in `conic_form`, and only the latter happens again on each solve, and therefore using atoms to defer work makes `fix!` work correctly. So I am somewhat concerned there could be other issues. But most cases just lower to another expression of atoms without doing any actual computation, and those expressions are ultimately in terms of lazy atoms. I guess the tell here is that we call `evaluate`, meaning we are looking at the current value rather than the symbolic value.

I think one fix here is to convert `quad_form` to an atom (probably following [CVX](https://www.cvxpy.org/_modules/cvxpy/atoms/quad_form.html)); this way we could defer the work that depends on the numerical value until `conic_form` time (which gets redone each solve). Another is to detect it and error as here.

We could also consider removing `fix!` and `free!`. They are a source of bugs and do not really provide performance improvements, though they are an ergonomic way to interact with a model.

odow:
Yeah I think this was the only reformulation that used `evaluate` from what I could tell

PR #588, opened 2024-04-10
Add support for setting MOI.VariablePrimalStart

### LINKED ISSUE #233: Provide initial guess of variables without warmstart
Is it possible to provide an initial guess of the unknowns when using Convex.jl to solve a convex optimization problem, without using a warmstart from a previous solve? For example, in the sample code I want to use a function like JuMP's setvalue to provide an initial guess of the unknown variable x.

using Convex
m = 4;  n = 5; A = randn(m, n); b = randn(m, 1);
x = Variable(n);
setvalue(x,ones(n,1)) # setvalue does not work for Convex.jl. Is there a work-around?
problem = minimize(sumsquares(A * x - b), [x >= 0])
solve!(problem)

### LINKED ISSUE #518: Add warmstart back
Warmstarting was removed in https://github.com/jump-dev/Convex.jl/pull/504, it should be added back before releasing
In particular, we should revert the following which removes warmstarting in the docs:
https://github.com/jump-dev/Convex.jl/pull/504/commits/10ed4448642bb8f86b864d7f88e373bd202dd390
https://github.com/jump-dev/Convex.jl/pull/504/commits/a50cff64b8c6a83948db219f99da4d4cfa57b662

odow:
Closes #518 
Closes #233

I'll comment something in-line

### DIFF
--- a/docs/src/manual/advanced.md
+++ b/docs/src/manual/advanced.md
@@ -34,6 +34,31 @@ p.constraints[1].dual
 constraint.dual
 ```
 
+## Warmstarting
+
+If you're solving the same problem many times with different values of a
+parameter, Convex.jl can initialize many solvers with the solution to the
+previous problem, which sometimes speeds up the solution time. This is called a
+**warm start**.
+
+To use this feature, pass the optional argument `warmstart=true` to the `solve!`
+method.
+
+```julia
+using Convex, SCS
+n = 1_000
+y = rand(n)
+x = Variable(n)
+lambda = Variable(Positive())
+fix!(lambda, 100)
+problem = minimize(sumsquares(y - x) + lambda * sumsquares(x - 10))
+@time solve!(problem, SCS.Optimizer)
+# Now warmstart. If the solver takes advantage of warmstarts, this run will be
+# faster
+fix!(lambda, 105)
+@time solve!(problem, SCS.Optimizer; warmstart = true)
+```
+
 ## Fixing and freeing variables
 
 Convex.jl allows you to fix a variable `x` to a value by calling the `fix!`
--- a/test/test_utilities.jl
+++ b/test/test_utilities.jl
@@ -1259,7 +1259,8 @@ function test_variable_primal_start()
     problem = satisfy([x1 >= 0, x2 <= 0, x3 <= 1, x4 == 5, x5 == 0])
     context = Convex.Context(
         problem,
-        () -> MOI.Utilities.UniversalFallback(MOI.Utilities.Model{Float64}()),
+        () ->
+            MOI.Utilities.UniversalFallback(MOI.Utilities.Model{Float64}()),
     )
     Convex._add_variable_primal_start(context)
     for (key, query) in (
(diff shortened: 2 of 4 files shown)


docs/src/manual/advanced.md
```diff
@@ -34,6 +34,31 @@ p.constraints[1].dual
...
+y = rand(n)
+x = Variable(n)
+lambda = Variable(Positive())
+fix!(lambda, 100)
+problem = minimize(sumsquares(y - x) + lambda * sumsquares(x - 10))
+@time solve!(problem, SCS.Optimizer)
+# Now warmstart. If the solver takes advantage of warmstarts, this run will be
+# faster
+fix!(lambda, 105)
+@time solve!(problem, SCS.Optimizer; warmstart = true)
```
odow, line 59: This example doesn't really make sense because SCS requires a complete start, and Convex may introduce arbitrary numbers of intermediate variables.

blegat: In MOI bridges, when we introduce intermediate variables, we set the starting values for them as well. Maybe that's another motivation to move these parts to use the custom MOI cones and let the bridges do the reformulation.

odow: I realized that the loop will account for these values after a solve. You just can't set their value _before_ a solve.

blegat: I think the SCS MOI wrapper should throw a warning if the starting values are only partially set.
Is this a regression from the previous release of Convex.jl ?

odow: Ah. SCS uses `0` if default:
https://github.com/jump-dev/SCS.jl/blob/e362f9551ba9f977487263cb5162d5801bf5cb3b/src/MOI_wrapper/MOI_wrapper.jl#L345

Your confusion is because Convex conflates `VariablePrimalStart` with `VariablePrimal` If `warmstart = true`, it looks up `.value`, which is also where the primal solution is stored.

blegat: So what about your first message in this thread ? Does the example make sense or not ?

odow: In retrospect it does make sense.

test/test_utilities.jl
```diff
@@ -1246,6 +1246,43 @@ function test_set_value_complex()
...
+    x3 = Variable(3, 3)
+    x3.value = 3.0
+    x4 = ComplexVariable()
+    x4.value = 1 + 2im
+    x5 = ComplexVariable(2, 2)
+    x5.value = [1 2; 3im 4]
+    problem = satisfy([x1 >= 0, x2 <= 0, x3 <= 1, x4 == 5, x5 == 0])
+    context = Convex.Context(
+        problem,
+        () -> MOI.Utilities.UniversalFallback(MOI.Utilities.Model{Float64}()),
```
odow, line 1262:
```suggestion
        () ->
            MOI.Utilities.UniversalFallback(MOI.Utilities.Model{Float64}()),
```

blegat (approved):

PR #592, opened 2024-04-15
Rm mention of `latex_formulation` from changelog

ericphanson:
This code was never in a release, it was added for debugging during 0.16 development and removed in v0.16 development too


odow:
Closing in favor of https://github.com/jump-dev/Convex.jl/pull/589.

PR #593, opened 2024-04-15
[breaking] Fix a<=b to return a-b in Nonpositives instead of b-a in Nonnegatives

odow:
x-ref https://github.com/jump-dev/Convex.jl/pull/590

### DIFF
--- a/docs/src/changelog.md
+++ b/docs/src/changelog.md
@@ -22,6 +22,11 @@ changes.
       elementwise values. This does not affect the result of `relative_entropy`.
     * The function `constant` should be used instead of the type `Constant`
       (which now refers to exclusively real constants).
+    * The constraint `a <= b` now produces `a - b in Nonpositives()` instead of
+      `b - a in Nonnegatives()`. The primal solutions are equivalent, but **the
+      dual variable associated with such constraints is now reversed in sign**.
+      (Following the convention in MathOptInterface, the dual of `a <= b` is
+      always, negative, regardless of optimization sense.) (#593)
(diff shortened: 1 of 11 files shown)


docs/src/changelog.md
```diff
@@ -22,6 +22,11 @@ changes.
       elementwise values. This does not affect the result of `relative_entropy`.
     * The function `constant` should be used instead of the type `Constant`
       (which now refers to exclusively real constants).
+    * The constraint `a <= b` now produces `a - b in Nonpositives()` instead of
+      `b - a in Nonnegatives()`. The primal solutions are equivalent, but **the
+      dual variable associated with such constraints is now reversed in sign**.
+      (Following the convention in MathOptInterface, the dual of `a <= b` is
+      always, negative, regardless of optimization sense.) (#593)
```
odow, line 29:
```suggestion
      always negative, regardless of optimization sense.) (#593)
```

src/atoms/exp_cone/LogSumExpAtom.jl
```diff
@@ -32,7 +32,7 @@ function new_conic_form!(context::Context, e::LogSumExpAtom)
     # log(sum(exp(x))) <= t  <=>  sum(exp(x)) <= exp(t) <=> sum(exp(x - t)) <= 1
     t = Variable()
     z = sum(exp(e.children[1] - t * ones(size(e.children[1]))))
-    add_constraint!(context, z <= 1)
+    add_constraint!(context, 1 >= z)
```
odow, line 35: Changing these just means we don't have to update the tests, and they use the more popular `Nonnegatives` instead of `Nonpositives`.

blegat (approved):

PR #595, opened 2024-04-18
[breaking] deprecate field access to .optval and .status

odow:
So we should decide whether to do: https://github.com/jump-dev/Convex.jl/issues/346


Project.toml
```diff
@@ -9,6 +9,7 @@ LDLFactorizations = "40e66cde-538c-5869-a4ad-c39174c6795b"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MathOptInterface = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
 OrderedCollections = "bac558e1-5e72-5ebc-8fee-abe8a469f55d"
+SCS = "c946c3f1-0d1f-5ce8-9dea-7daa1f7e2d13"
```
odow, line 12:
```suggestion
```

odow:
Ah. `objective_value` and `termination_status` weren't actually exported. So there'd likely be a conflict if someone has `using JuMP, Convex`.

odow:
I tend to think this is overly breaking for little improvement. I wonder if we should close this as won't-fix.

PR #598, opened 2024-04-18
Add default set_with_size

blegat:
I'm not sure we should allow matrices with more than one columns to be constrained to SOC. We can have for a user coming up for a good motivation and leave it as error at least for now.

### DIFF
--- a/src/constraints/GenericConstraint.jl
+++ b/src/constraints/GenericConstraint.jl
@@ -12,6 +12,14 @@ function GenericConstraint{S}(child::AbstractExpr) where {S<:MOI.AbstractSet}
     return GenericConstraint(child, set_with_size(S, size(child)))
 end
 
+function set_with_size(::Type{S}, sz::Tuple{Int,Int}) where {S<:MOI.AbstractVectorSet}
+    if sz[2] != 1
+        error("Cannot constrain a matrix of size `$sz` to be long to the cone" *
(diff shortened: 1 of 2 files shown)


src/constraints/GenericConstraint.jl
```diff
@@ -12,6 +12,19 @@ function GenericConstraint{S}(child::AbstractExpr) where {S<:MOI.AbstractSet}
     return GenericConstraint(child, set_with_size(S, size(child)))
 end
 
+function set_with_size(
+    ::Type{S},
+    sz::Tuple{Int,Int},
+) where {S<:MOI.AbstractVectorSet}
+    if sz[2] != 1
```
odow, line 19: This needs a test

test/test_constraints.jl
```diff
@@ -239,6 +239,20 @@ function test_GenericConstraint_SecondOrderCone()
...
 end
 
+function test_GenericConstraint_SecondOrderCone_set_with_size()
+    x = Variable(2, 2)
+    S = MOI.SecondOrderCone
+    sz = size(x)
+    @test_throws(
+        ErrorException(
+            "Cannot constrain a matrix of size `$sz` to be long to the cone " *
+            "`$S``, there should be only one column.",
```
odow, line 249:
```suggestion
            "`$S`, there should be only one column.",
```

PR #599, opened 2024-04-18
Replace is_feasible by distance_to_set

blegat:
I think a default fallback returning `true` is incorrect. Since the constraint is not added when it is constant, we should check that it is feasible otherwise the infeasibility is silently ignored and the resulting status is incorrect.

### DIFF
--- a/src/constraints/GenericConstraint.jl
+++ b/src/constraints/GenericConstraint.jl
@@ -20,7 +20,7 @@ vexity(c::GenericConstraint) = vexity(vexity(c.child), c.set)
 
 function _add_constraint!(context::Context, c::GenericConstraint)
     if vexity(c.child) == ConstVexity()
-        dist = MOI.Utilities.distance_to_set(evaluate(c.child), set)
+        dist = MOI.Utilities.distance_to_set(evaluate(c.child), c.set)
         if dist > CONSTANT_CONSTRAINT_TOL[]
             context.detected_infeasible_during_formulation[] = true
         end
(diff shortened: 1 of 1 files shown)


odow:
`distance_to_set` isn't implemented for PositiveSemidefiniteConeSquare`

src/constraints/GenericConstraint.jl
```diff
@@ -14,17 +14,23 @@ end
...
+# A fallback. Define a new method if `MOI.Utilities.distance_to_set`
+# is not defined.
+function is_feasible(x, set, tol)
+    return MOI.Utilities.distance_to_set(x, set) <= tol
+end
+
+function is_feasible(x::Number, set::MOI.AbstractVectorSet, tol)
+    return is_feasible([x], set, tol)
+end
+
```
odow, line 28:
```suggestion

```

src/constraints/GenericConstraint.jl
```diff
@@ -14,17 +14,22 @@ end
...
 
+# A fallback. Define a new method if `MOI.Utilities.distance_to_set`
+# is not defined.
+function is_feasible(x, set, tol)
+    return MOI.Utilities.distance_to_set(x, set) <= tol
+end
+
+function is_feasible(x::Number, set::MOI.AbstractVectorSet, tol)
+    return is_feasible([x], set, tol)
+end
```
odow, line 27:
```suggestion
end

```

PR #600, opened 2024-04-19
Simplify exponential cone constraint

blegat:
I don't know why the exponential constraint between matrices is simply the exponential cone entrywise. It's that actually used somewhere is are the matrices always 1 x 1 ? This PR removes the ability to use a matrix version, only the dimension-3 exponential cone is allowed.


src/constraints/GenericConstraint.jl
```diff
@@ -41,6 +41,20 @@ end
 
 vexity(c::GenericConstraint) = vexity(vexity(c.child), c.set)
 
+function vexity(
+    vex::Vexity,
+    ::Union{
+        MOI.PositiveSemidefiniteConeSquare,
+        MOI.SecondOrderCone,
+        MOI.ExponentialCone,
```
odow, line 49: I would much prefer that we kept separate methods for this, even if they are duplicated.

src/constraints/GenericConstraint.jl
```diff
@@ -241,14 +248,9 @@ end
 ⪯(x::AbstractExpr, y::Value) = ⪰(y, x)
 
 # ==============================================================================
-#     SecondOrderCone
+#     SecondOrderCone, ExponentialCone
```
odow, line 251: Let's just duplicate the code. It keeps things much clearer in the long term.

odow:
I don't think this works, because the various `LogAtom` and `ExpAtom`s apply to matrices.

odow:
Closing in favor of #602

PR #604, opened 2024-04-23
Refactor GeomMeanEpiCone into GenericConstraint

blegat:
This is an attempt to refactor `GeomMeanEpiCone` into `GenericConstraint`. In the long term (meaning definitely post v0.16), we can replace the custom `_add_constraint` with a bridge so that the cone can be used in JuMP as well.
If we do the same for the two other remaining constraints, we can then get rid of `abstract type Constraint` and rename `GenericConstraint` into `Constraint`.
~~This design is blocked by https://github.com/jump-dev/Convex.jl/issues/603~~


src/constraints/GeomMeanEpiCone.jl
```diff
@@ -133,12 +65,13 @@ end
...
+    constraint::GenericConstraint{GeomMeanEpiCone},
 )
-    A = constraint.cone.A
-    B = constraint.cone.B
-    t = constraint.cone.t
-    T = constraint.T
+    n = constraint.set.side_dimension
+    T = constraint.child[1:n^2]
+    A = constraint.child[n^2 .+ (1:n^2)]
+    B = constraint.child[2n^2 .+ (1:n^2)]
```
ericphanson, line 73: indexing in Convex is non-trivial (it's another sparse affine operation); I'd worry that this could regress problem formulation (and possibly even solve times). In Convex's current setup, I'd wonder if adding a non-concretely typed field to GenericConstraint to carry extra data would be better, and that maybe a dynamic dispatch is better than indexing for Convex.

blegat: Maybe we could allow `child` to be a `Tuple` and interpret it as a `vcat`

ericphanson: yeah, if it's a `Tuple` with elements of `AbstractExpr`, then indexing is just normal julia indexing, so there should be no issues there.

odow:
I don't particularly like this design. The "Convex.jl" approach would probably be to add a `TupleAtom`?

blegat:
It's the concatenation that should belong to the cone so we should use a `vcat` atom (which is seems gets rewritten with `hcat` and `reshape`s). If `getindex` on a `vcat` atom, would check if the index fit exactly in one of the argument of `vcat` and then in that case unwrap the `vcat` atom then it would fix https://github.com/jump-dev/Convex.jl/issues/603

blegat:
Now that https://github.com/jump-dev/Convex.jl/issues/603 is fixed, I reverted the `Tuple` approach

src/constraints/GeomMeanEpiCone.jl
```diff
@@ -5,146 +5,83 @@
...
     context::Context,
-    constraint::GeomMeanEpiConeConstraint,
+    constraint::GenericConstraint{GeomMeanEpiConeSquare},
 )
-    A = constraint.cone.A
-    B = constraint.cone.B
-    t = constraint.cone.t
-    T = constraint.T
+    n = constraint.set.side_dimension
+    T = reshape(constraint.child[1:n^2], n, n)
```
blegat, line 81: It may seems silly to vectorize to fit into the `GenericConstraint` and then have to `reshape` here but in the long term, this `_add_constraint!` will be replaced by bridges which will work with the vectorized version anyway.

blegat:
I had to refactor a bit to work around this bug of JuliaFormatter: https://github.com/domluna/JuliaFormatter.jl/issues/833

src/constraints/GeomMeanEpiCone.jl
```diff
@@ -5,146 +5,87 @@
...
-        if size(T) != cone.size
-            throw(DimensionMismatch("T must be size $(cone.size)"))
-        end
-        return new(T, cone)
-    end
+head(io::IO, ::GeomMeanEpiConeSquare) = print(io, "∈(GeomMeanEpiConeSquare)")
 
-    function GeomMeanEpiConeConstraint(T::Value, cone::GeomMeanEpiCone)
-        return GeomMeanEpiConeConstraint(constant(T), cone)
+function Base.in(func::Tuple, set::GeomMeanEpiConeSquare)
```
blegat, line 48: We should probably get rid of this as it will become type piracy once `GeomMeanEpiConeSquare` is moved to MOI

src/constraints/GeomMeanEpiCone.jl
ericphanson: reshape should be cheap also, it only updates the `size` metadata, it doesn't touch the data (since everything is also vectorized inside `conic_form!`s)

src/constraints/GeomMeanEpiCone.jl
ericphanson: maybe we can have syntax like `(A,B,C) in Constraint(MOI.GeomMeanEpiConeSquare())` instead, where `Constraint` would be owned by Convex

src/constraints/GeomMeanEpiCone.jl
```diff
@@ -5,146 +5,87 @@
...
+## Reference
+
+Ported from CVXQUAD which is based on the paper: "Lieb's concavity theorem,
+matrix geometric means and semidefinite optimization" by Hamza Fawzi and James
+Saunderson (arXiv:1512.03401)
 """
-mutable struct GeomMeanEpiCone
-    A::AbstractExpr
-    B::AbstractExpr
+struct GeomMeanEpiConeSquare <: MOI.AbstractVectorSet
```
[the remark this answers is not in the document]
blegat: For consistency for`GeometricMeanCone`, yes, we should expand. Epi vs Epigraph I'm ok with both. In Hypatia, I think Epi is used

src/constraints/GeomMeanEpiCone.jl
blegat: Yes, this can be a follow up PR

src/constraints/GeomMeanEpiCone.jl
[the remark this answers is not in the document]
odow: Renamed to `GeometricMeanEpiConeSquare`

src/constraints/GeomMeanEpiCone.jl
odow: Let's just make this `GenericConstraint(::Tuple, ::GeometricMeanEpiConeSquare)` for now?

odow: Or `Base.in(f::AbstractExpr, set::MOI.AbstractSet) = GenericConstraint(f, set)`? Why do we need the tuple? Why not `vcat(T, A, B)`?

test/test_constraints.jl
```diff
@@ -297,6 +297,45 @@ function test_GenericConstraint_RelativeEntropyConeConstraint()
     return
 end
 
+### constraints/GenericConstraint_RelativeEntropyCone
```
blegat, line 300:
```suggestion
### constraints/GenericConstraint_GeometricMeanEpiConeSquare
```

src/constraints/GeomMeanEpiCone.jl
blegat: It was just be able to throw the DimensionMismatch errors

blegat: We can also just have some internal utility for now

blegat: Like we did for `_add_vectorized_exp_cone`

odow: How about now? I've just gone for explicit `Convex.GenericConstraint`.

blegat:
Looks good to me, any objection to merge ?

ericphanson (approved):

PR #605, opened 2024-04-25
Add RootDetAtom

### LINKED ISSUE #214: Support for det(P)^(1/n)
In trying to solve a problem with objective function
`maximize det(P)`
there are two common transformations that turn the objective into a convex objective.
The first is to convert the problem to
`maximize logdet(P)`
which is currently implemented and can be solved using e.g. SCS.
Another option is to instead solve
`maximize det(P)^(1/n)`
This is equivalent to the geometric mean of the eigenvalues of `P`. From what I've read, this can be modeled with semidefinite and second order cones (I don't know the details of this). This allows a lot more solvers to be used to solve this optimization problem (instead of relying on a first order method like SCS)

Would it be possible to get support for the nth root of the determinant added?

odow:
Closes #214

### DIFF
--- a/docs/src/manual/operations.md
+++ b/docs/src/manual/operations.md
@@ -113,7 +113,7 @@ solver (including SCS and Mosek).
 | `opnorm(x, 2)` (`operatornorm(x)`)| max of singular values of $x$      | convex       | not monotonic |                        |
 | `eigmax(x)`                       | max eigenvalue of $x$              | convex       | not monotonic |                        |
 | `eigmin(x)`                       | min eigenvalue of $x$              | concave      | not monotonic |                        |
-| `rootdet(X)`                      | root-determinant of the matrix X   | concave      | not monotonic |                        |
+| `rootdet(X)`                      | n-th root-determinant of the $n$-by-$n$ matrix X, that is $det(X)^{1/n}$ | concave      | not monotonic |                        |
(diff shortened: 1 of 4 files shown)


docs/src/manual/operations.md
```diff
@@ -113,6 +113,7 @@ solver (including SCS and Mosek).
 | `opnorm(x, 2)` (`operatornorm(x)`)| max of singular values of $x$      | convex       | not monotonic |                        |
 | `eigmax(x)`                       | max eigenvalue of $x$              | convex       | not monotonic |                        |
 | `eigmin(x)`                       | min eigenvalue of $x$              | concave      | not monotonic |                        |
+| `rootdet(X)`                      | root-determinant of the matrix X   | concave      | not monotonic |                        |
```
blegat, line 116: Maybe mention somewhere that by root we mean the `n`-th root ?

odow: CVX calls this `det_rootn`. Is that better?

blegat: `rootdet` is more consistent with `LinearAlgebra.logdet`. We could also just support `det(Q)^(1/n)`, that will be needed anyway for the MOI wrapper to support `JuMP.@objective(model, Max, det(Q)^(1/n))`

blegat: We decided to go for `rootdet` in the JuMP-dev call, we can add `rootdet` as operator to JuMP later if needed

Issue #603, opened 2024-04-23
vexity(::IndexAtom) is inexact


blegat:
The `vexity` of the `IndexAtom` is incorrect:
```julia
julia> x = Variable()
Variable
size: (1, 1)
sign: real
vexity: affine
id: 176…085

julia> v = vcat(x, square(x))
reshape (convex; real)
└─ * (convex; real)
   ├─ 2×2 Matrix{Bool}
   └─ reshape (convex; real)
      └─ hcat (convex; real)
         ├─ …
         └─ …

julia> vexity(v)
Convex.ConvexVexity()

julia> vexity(v[1])
Convex.ConvexVexity()

julia> v = vcat(-square(x), square(x))
reshape (Convex.NotDcp; real)
└─ * (Convex.NotDcp; real)
   ├─ 2×2 Matrix{Bool}
   └─ reshape (Convex.NotDcp; real)
      └─ hcat (Convex.NotDcp; real)
         ├─ …
         └─ …

julia> vexity(v)
Convex.NotDcp()

julia> vexity(v[1])
Convex.NotDcp()
```

ericphanson:
The current definition doesn’t depend on the index, it just defers to the object being indexed.  I think it hasn’t come up before since usually all elements in the vector have the same vexity.

odow:
This isn't incorrect so much as inexact: could be improved, but it won't give wrong answers.

blegat:
We discussed this in the JuMP-dev call, the conclusion was that we should try adding a `vcat` atom and specialize `getindex` for the `vcat` atom to try and see if the indices fit in one of the argument in the `vcat` in which case we remove the `vcat` atom and propagage `getindex`

PR #607, opened 2024-04-29
Add VcatAtom

odow:
x-ref #603

### DIFF
--- a/src/atoms/affine/HcatAtom.jl
+++ b/src/atoms/affine/HcatAtom.jl
@@ -54,7 +54,7 @@ function Base.hvcat(
     output_rows = Vector{HcatAtom}(undef, length(rows))
     offset = 0
     for (i, n) in enumerate(rows)
-        output_rows[i] = HcatAtom(args[offset .+ (1:n)]...)
+        output_rows[i] = HcatAtom(args[offset.+(1:n)]...)
         offset += n
     end
     return vcat(output_rows...)
--- a/src/atoms/affine/HcatAtom.jl
+++ b/src/atoms/affine/HcatAtom.jl
@@ -29,7 +29,7 @@ monotonicity(x::HcatAtom) = ntuple(_ -> Nondecreasing(), length(x.children))
 
 curvature(::HcatAtom) = ConstVexity()
 
-evaluate(x::HcatAtom) = reduce(hcat, map(evaluate, x.children))
+evaluate(x::HcatAtom) = reduce(hcat, collect(map(evaluate, x.children)))
 
 function new_conic_form!(context::Context{T}, x::HcatAtom) where {T}
     args = map(c -> conic_form!(context, c), AbstractTrees.children(x))
--- a/src/atoms/affine/HcatAtom.jl
+++ b/src/atoms/affine/HcatAtom.jl
@@ -41,7 +41,9 @@ end
 Base.hcat(args::AbstractExpr...) = HcatAtom(args...)
 
 function Base.hcat(args::Union{AbstractExpr,Value}...)
-    @assert !all(Base.Fix2(isa, Value), args)
+    if all(Base.Fix2(isa, Value), args)
+        return Base.cat(args..., dims = Val(2))
+    end
     return HcatAtom(args...)
 end
 
--- a/src/atoms/affine/HcatAtom.jl
+++ b/src/atoms/affine/HcatAtom.jl
@@ -29,70 +29,31 @@ monotonicity(x::HcatAtom) = ntuple(_ -> Nondecreasing(), length(x.children))
 
 curvature(::HcatAtom) = ConstVexity()
 
-evaluate(x::HcatAtom) = hcat(map(evaluate, x.children)...)
+evaluate(x::HcatAtom) = reduce(hcat, map(evaluate, x.children))
 
 function new_conic_form!(context::Context{T}, x::HcatAtom) where {T}
-    objectives = map(c -> conic_form!(context, c), AbstractTrees.children(x))
-    # Suppose the child objectives for two children e1 (2 x 1) and e2 (2 x 2)
-    # look something like
-    #  e1: x => 1 2 3
-    #           4 5 6
-    #      y => 2 4
-    #           7 8
-    #  e2: x => 1 1 1
-    #           2 2 2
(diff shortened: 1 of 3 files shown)


src/atoms/affine/HcatAtom.jl
```diff
@@ -73,26 +47,15 @@ function Base.hcat(args::Union{AbstractExpr,Value}...)
...
-    nbr = length(rows)
-    rs = Vector{Any}(undef, nbr)
-    a = 1
-    for i in 1:nbr
-        rs[i] = HcatAtom(args[a:a-1+rows[i]]...)
-        a += rows[i]
+    output_rows = Vector{HcatAtom}(undef, length(rows))
+    offset = 0
+    for (i, n) in enumerate(rows)
+        output_rows[i] = HcatAtom(args[offset .+ (1:n)]...)
```
odow, line 57:
```suggestion
        output_rows[i] = HcatAtom(args[offset.+(1:n)]...)
```

src/atoms/affine/VcatAtom.jl
```diff
@@ -0,0 +1,61 @@
...
+    # which is not what we are after. We need to first transpose each child to
+    # get:
+    #   x^T, y^T = [1 2; 3 4], [5 6; 7 8])
+    # then hcat them to get:
+    #   hcat(x^T, y^T) = [1 2 5 6; 3 4 7 8]
+    # then transpose this to get:
+    #   hcat(x^T, y^T)^T = [1 3; 2 4; 5 7; 6 8]
+    # so our final connic form produces the desired
+    #   [1, 2, 5, 6, 3, 4, 7, 8]
+    return conic_form!(context, transpose(reduce(hcat, transpose.(x.children))))
```
odow, line 51: This part is not nice, but I don't know an alternative.

src/atoms/affine/VcatAtom.jl
```diff
@@ -0,0 +1,61 @@
...
+    # get:
+    #   vcat([1, 2, 3, 4], [5, 6, 7, 8]) = [1, 2, 3, 4, 5, 6, 7, 8]
+    # which is not what we are after. We need to first transpose each child to
+    # get:
+    #   x^T, y^T = [1 2; 3 4], [5 6; 7 8])
+    # then hcat them to get:
+    #   hcat(x^T, y^T) = [1 2 5 6; 3 4 7 8]
+    # then transpose this to get:
+    #   hcat(x^T, y^T)^T = [1 3; 2 4; 5 7; 6 8]
+    # so our final connic form produces the desired
```
blegat, line 49:
```suggestion
    # so our final conic form produces the desired
```

src/atoms/affine/VcatAtom.jl
```diff
@@ -0,0 +1,61 @@
...
+    #   [1, 2, 5, 6, 3, 4, 7, 8]
```
blegat, line 50: How does the outer transpose work ? If we get the conic form of the `hcat` then we still need to apply the transpose to that vectorized form right ?

src/atoms/affine/HcatAtom.jl
```diff
@@ -29,41 +29,15 @@ monotonicity(x::HcatAtom) = ntuple(_ -> Nondecreasing(), length(x.children))
 
 curvature(::HcatAtom) = ConstVexity()
 
-evaluate(x::HcatAtom) = hcat(map(evaluate, x.children)...)
+evaluate(x::HcatAtom) = reduce(hcat, map(evaluate, x.children))
```
ericphanson, line 32: To avoid the quadratic time naive version:
```suggestion
evaluate(x::HcatAtom) = reduce(hcat, collect(map(evaluate, x.children)))
```

src/atoms/affine/VcatAtom.jl
ericphanson: One option is to manually build the operator we want to do: https://github.com/jump-dev/Convex.jl/blob/0bd470bafbaecdd165a71708a6bf745ebe9b550b/src/reformulations/partialtranspose.jl#L44

ericphanson: actually, I think that `permutedims_matrix` function already does what we need:
```julia
julia> x = [1 3; 2 4]
2×2 Matrix{Int64}:
 1  3
 2  4

julia> y = [5 7; 6 8]
2×2 Matrix{Int64}:
 5  7
 6  8

julia> M = permutedims_matrix((size(x, 1),size(x,2),size(y,1)), (1,3,2))
8×8 SparseMatrixCSC{Bool, Int64} with 8 stored entries:
 1  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅
 ⋅  1  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅
 ⋅  ⋅  ⋅  ⋅  1  ⋅  ⋅  ⋅
 ⋅  ⋅  ⋅  ⋅  ⋅  1  ⋅  ⋅
 ⋅  ⋅  1  ⋅  ⋅  ⋅  ⋅  ⋅
 ⋅  ⋅  ⋅  1  ⋅  ⋅  ⋅  ⋅
 ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  1  ⋅
 ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  1

julia> reshape(M * vcat(vec(x),vec(y)), size(x,1) + size(y,1), size(x,2))
4×2 Matrix{Int64}:
 1  3
 2  4
 5  7
 6  8
```

src/atoms/affine/VcatAtom.jl
odow: What do you mean by how does it work? I've added extra tests, and everything seems to pass.

src/atoms/affine/VcatAtom.jl
odow: I've kept as-is for now. It wasn't obvious how to generalize this to the n-ary case, and what we currently have works.

blegat (approved):

src/atoms/affine/VcatAtom.jl
blegat: We can try this refactoring in a separate PR

PR #610, opened 2024-05-02
Document atoms

odow:
Just exploring a few options:

https://jump.dev/Convex.jl/previews/PR610/reference/atoms/#Supported-operations

I still have these to go:

- [x] BroadcastMultiplyAtom
- [x] DiagMatrixAtom
- [x] HcatAtom
- [ ] IndexAtom
- [x] MatrixFracAtom
- [x] MultiplyAtom
- [x] QolElemAtom
- [x] QuadOverLinAtom
- [x] QuantumEntropyAtom
- [x] QuantumRelativeEntropyAtom
- [x] RationnalNormAtom
- [x] ReshapeAtom
- [x] TraceLogmAtom
- [x] TraceMpowerAtom
- [x] VcatAtom


ericphanson:
the current docs are kind of centered around the constructor functions rather than the structs, since usually users should use the former. I think though it definitely is nice to document the actual primitives that are available. I think maybe it could be most useful to keep the functions as the first thing the users see/interact with, and add docstrings to them, and then have the atoms as more advanced docs.

odow:
> the current docs are kind of centered around the constructor functions rather than the structs, since usually users should use the former

Yeah... I've realized this. I just wondered about adding docstrings for `Base.abs`, etc. But perhaps it's okay.

odow:
There's also: https://jump.dev/Convex.jl/previews/PR610/manual/operations/

But it doesn't have every atom or reformulation (e.g., `abs2` is missing).

I also wonder about how important the DCP rules are. I don't think many people are computing the rules in their head from the docs? They probably just try it to see if it works.

odow:
What are thoughts on https://jump.dev/Convex.jl/previews/PR610/atoms/

Too verbose? What parts are useful?

ericphanson:
I like it! IMO not too verbose. In fact I think the reformulations could be more verbose. I think the examples are quite useful, and I like the mini table too about convexity properties. I find them useful at least, and I think when folks are debugging stuff it comes in handy.

odow:
> In fact I think the reformulations could be more verbose

Yes, if we're not focusing on atoms, then it makes sense to have these identical. 

I even wonder if it makes sense to move all of the `Base.abs` etc methods to a single `operators.jl` file. The atoms (as in, the structs) are standalone, and how exactly we implement the operators is independent. We might, for example, decide to rewrite some of the existing "reformulations" to become atoms, but that wouldn't change the public interface.

ericphanson:
true, and we have done that in the past (though often the other way, deleting slow or buggy atoms in favor of reformulations)

blegat (approved):
I really like this so I think we should merge before releasing v0.16. I've added it in the milestone

odow:
I don't know if this needs to block the release

blegat:
It does not, you can remove it from the milestone if this PR is not ready to merge

odow:
@blegat thoughts on keeping the operators in a separate file or keep with atom structs?

blegat:
> @blegat thoughts on keeping the operators in a separate file or keep with atom structs?

I like it in a separate file

odow:
So what do we think about this?

docs/make.jl
```diff
@@ -112,6 +112,7 @@ Documenter.makedocs(
         "Manual" => [
             "manual/types.md",
             "manual/operations.md",
+            "manual/atoms.md",
```
odow, line 115:
```suggestion
```

docs/make.jl
```diff
@@ -122,7 +122,7 @@ Documenter.makedocs(
             "developer/contributing.md",
             "developer/credits.md",
         ],
-        "reference.md",
+        "API Referennce" => ["reference/api.md", "reference/atoms.md"],
```
odow, line 125:
```suggestion
        "API Reference" => ["reference/api.md", "reference/atoms.md"],
```

odow:
I've moved this to the API reference, so I think I'm going to merge, unless there are other comments.

blegat (approved):

PR #613, opened 2024-05-04
optimize `ExpAtom`

### LINKED ISSUE #254: Massive RAM for moderate problem

ericphanson:
closes #254 

Before: on the original problem, today I OOM'd on my 16 GB macbook pro. On [previous attempts](https://github.com/jump-dev/Convex.jl/issues/254#issuecomment-1636944289), I got
```julia
1303.736478 seconds (11.94 M allocations: 191.588 GiB, 92.49% gc time)
```

On this branch:
```julia
  5.474096 seconds (1.13 M allocations: 216.129 MiB, 20.82% gc time)
```

The issue is formulating 33378 IndexAtoms and so forth. In general in Convex we must never do scalar operations if we hope to formulate things performantly (unless we move everything to a scalar representation like MOI).

Here, I couldn't see how to avoid the scalar operations, so I lower to MOI to do them on that level, which is efficient.

I filed #614 to track this issue more generally.

### DIFF
--- a/src/atoms/ExpAtom.jl
+++ b/src/atoms/ExpAtom.jl
@@ -45,10 +45,10 @@ function new_conic_form!(context::Context{T}, e::ExpAtom) where {T}
     # Instead, we will drop to the MOI level to implement this in terms of scalar operations.
     # First, we will get `x` as an MOI.VectorAffineFunction
     x_tape = conic_form!(context, x)
-    # since `ExpAtom` is restricted to `sign(x)` being real, `x_tape` is either `Vector{T}` or `SparseTape{T}`.
-    # The `Vector{T}` case happens when `x` is a constant (or a function of a constant).
+    # since `ExpAtom` is restricted to `sign(x)` being real, `x_tape` is either `SPARSE_VECTOR{T}` or `SparseTape{T}`.
+    # The `SPARSE_VECTOR{T}` case happens when `x` is a constant (or a function of a constant).
     # In this case, we can just take `exp` directly.
-    if x_tape isa Vector
+    if x_tape isa SPARSE_VECTOR
         return exp.(x_tape)
     end
     vaf = to_vaf(x_tape)
--- a/src/atoms/ExpAtom.jl
+++ b/src/atoms/ExpAtom.jl
@@ -36,8 +36,8 @@ function new_conic_form!(context::Context{T}, e::ExpAtom) where {T}
     z = Variable(m, n)
     # Naive implementation:
     # for i in 1:m, j in 1:n
-    # f = vcat(x[i, j], 1, z[i, j])
-        # add_constraint!(context, GenericConstraint{MOI.ExponentialCone}(f))
+    #     f = vcat(x[i, j], 1, z[i, j])
+    #     add_constraint!(context, GenericConstraint{MOI.ExponentialCone}(f))
     # end
     # return conic_form!(context, z)
     # This is slow, since we are indexing on the Convex side, and convex is based around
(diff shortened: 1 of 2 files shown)


src/atoms/ExpAtom.jl
```diff
@@ -34,9 +34,37 @@ function new_conic_form!(context::Context{T}, e::ExpAtom) where {T}
...
+    # for i in 1:m, j in 1:n
+    #     f = vcat(x[i, j], 1, z[i, j])
+    #     add_constraint!(context, GenericConstraint{MOI.ExponentialCone}(f))
+    # end
+    # return conic_form!(context, z)
+    # This is slow, since we are indexing on the Convex side, and convex is based around
+    # vector/matrix operations. We don't want to produce n*m IndexAtoms!
+    # Instead, we will drop to the MOI level to implement this in terms of scalar operations.
+    # First, we will get `x` as an MOI.VectorAffineFunction
+    x_tape = conic_form!(context, x)
```
ericphanson, line 47: our tests seem to say it's fine, but I don't actually 100% remember if this will always be a `SparseTape{T}` like I expect here... I wonder if in some edge cases it can be something else, like a `Vector{T}`. I need to double-check.

src/problem_depot/problems/exp.jl
```diff
@@ -22,6 +22,19 @@
         @test evaluate(exp(y)) ≈ 1 atol = atol rtol = rtol
     end
 
+    # Test for constant `exp` (#613)
+    y = Variable()
+    x = constant([1,2,3])
```
ericphanson, line 27:
```suggestion
    x = constant([1, 2, 3])
```

src/problem_depot/problems/exp.jl
```diff
@@ -22,6 +22,19 @@
...
+    p = minimize(sum(exp(x)) + y, y >= 0; numeric_type = T)
+    if test
+        @test problem_vexity(p) == ConvexVexity()
+    end
+    handle_problem!(p)
+    if test
+        @test p.optval ≈ sum(exp.([1,2,3])) atol = atol rtol = rtol
```
ericphanson, line 34:
```suggestion
        @test p.optval ≈ sum(exp.([1, 2, 3])) atol = atol rtol = rtol
```

src/atoms/ExpAtom.jl
```diff
@@ -34,9 +34,43 @@ function new_conic_form!(context::Context{T}, e::ExpAtom) where {T}
...
+    # since `ExpAtom` is restricted to `sign(x)` being real, `x_tape` is either `Vector{T}` or `SparseTape{T}`.
+    # The `Vector{T}` case happens when `x` is a constant (or a function of a constant).
+    # In this case, we can just take `exp` directly.
+    if x_tape isa Vector
```
ericphanson, line 51:
```suggestion
    # since `ExpAtom` is restricted to `sign(x)` being real, `x_tape` is either `SPARSE_VECTOR{T}` or `SparseTape{T}`.
    # The `SPARSE_VECTOR{T}` case happens when `x` is a constant (or a function of a constant).
    # In this case, we can just take `exp` directly.
    if x_tape isa SPARSE_VECTOR
```

odow:
Closing in favor of #618

PR #617, opened 2024-05-04
Test conic forms for constant values

ericphanson:
Addresses https://github.com/jump-dev/Convex.jl/issues/616 in the real case only. In the complex case unfortunately I found many issues, too many to address in this PR. Some would be addressed similarly to the change for `new_conic_form!(context::Context{T}, e::RelativeEntropyAtom)`, where `GenericConstraint` nicely handles the constant case, but direct `MOI_add_constraint` does not.

(The issues here are not really correctness bugs, but rather errors one would get if for some reason they had constant values rather than variables in some atoms, such as if you `fix!`'d a variable).

### DIFF
--- a/test/test_atoms.jl
+++ b/test/test_atoms.jl
@@ -10,7 +10,6 @@ using Test
 
 # Do not use `using LinearAlgebra` to check symbols are re-exported by Convex.
 import LinearAlgebra
-import Clarabel
 import MathOptInterface as MOI
 
 function runtests()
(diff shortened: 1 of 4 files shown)


test/test_atoms.jl
```diff
@@ -10,7 +10,7 @@ using Test
 
 # Do not use `using LinearAlgebra` to check symbols are re-exported by Convex.
 import LinearAlgebra
-
+import Clarabel
```
odow, line 13: I'd rather _not_ have solvers in these tests. They should all just be logic tests. Solvers should go in `problem_depot`.

I'll make some changes to this PR.

src/atoms/RelativeEntropyAtom.jl
```diff
@@ -40,14 +40,14 @@ end
...
-    MOI_add_constraint(context.model, f, MOI.RelativeEntropyCone(2d + 1))
-    return u
+    t = Variable()
+    triple = vcat(t, e.children[2], e.children[1])
+    # We cache the conic forms in a particular order so the tests are happy
+    conic_form!(context, e.children[1])
+    conic_form!(context, e.children[2])
+    conic_form!(context, t)
+    add_constraint!(context, GenericConstraint{MOI.RelativeEntropyCone}(triple))
+    return conic_form!(context, t)
```
odow, line 50: I moved all of this to #623

odow:
Okay. I've simplified the test handling, but now it requires #623

odow:
Urgh. I borked that. Let's just wait and I'll rebase on `master`.

test/test_atoms.jl
```diff
@@ -74,6 +74,46 @@ function _test_atom(build_fn, target_string::String; value_type = Float64)
...
+function _to_constant(e::Convex.AbstractExpr)
+    e.children = map(_to_constant, e.children)
+    return e
+end
+
+function _test_constant_atom(build_fn; value_type)
+    context = Convex.Context{value_type}(MOI.Utilities.Model{value_type})
+    atom = _to_constant(build_fn(context))
+    if Convex.iscomplex(atom) || any(Convex.iscomplex, atom.children)
+        return
```
odow, line 105: I'll fix this in a separate PR.

odow (approved):

PR #620, opened 2024-05-05
Improve performance of QolElemAtom

odow:
x-ref #614

### DIFF
--- a/src/VectorAffineFunctionAsMatrix.jl
+++ b/src/VectorAffineFunctionAsMatrix.jl
@@ -21,6 +21,8 @@ function MOI.output_dimension(v::VectorAffineFunctionAsMatrix)
     return size(v.matrix, 1)
 end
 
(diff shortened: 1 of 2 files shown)


src/VectorAffineFunctionAsMatrix.jl
```diff
@@ -21,6 +21,8 @@ function MOI.output_dimension(v::VectorAffineFunctionAsMatrix)
     return size(v.matrix, 1)
 end
 
+to_vaf(tape::SPARSE_VECTOR) = tape
```
ericphanson, line 24: I've wanted this also, but I skipped it since `vaf` was supposed to mean VectorAffineFunction. Maybe we can do this + rename it?

odow: I'm going to leave this hack for now, and rename it in a separate PR. `to_vaf` is really `_to_vector_moi`

PR #624, opened 2024-05-05
[breaking] scalar row indexing produces column vector

### LINKED ISSUE #509: Scalar indexing of row of matrix creates row vector, causing incorrect `dot`
```julia
julia> x = Variable(2,2)
Variable
size: (2, 2)
sign: real
vexity: affine
id: 936…769

julia> set_value!(x, ones(2,2))
2×2 Matrix{Float64}:
 1.0  1.0
 1.0  1.0

julia> evaluate(x[2, :])
1×2 Matrix{Float64}:
 1.0  1.0

julia> evaluate(x)[2, :]
2-element Vector{Float64}:
 1.0
 1.0
```

In particular this causes `dot` to be incorrect: https://discourse.julialang.org/t/convex-jl-constraints-not-enforced/102439/2.

ericphanson:
fixes #509 

There _could_ be other issues like this out there, but at least this fixes that particular one, and I can't quite think of others off the top of my head.

### DIFF
--- a/docs/src/examples/optimization_with_complex_variables/power_flow_optimization.jl
+++ b/docs/src/examples/optimization_with_complex_variables/power_flow_optimization.jl
@@ -18,7 +18,7 @@ W = ComplexVariable(n, n);
 objective = real(sum(diag(W)));
 c1 = Constraint[];
 for i in 2:n
-    push!(c1, Y[i, :]' * W[i, :] == inj[i])
+    push!(c1, dot(Y[i, :], W[i, :]) == inj[i])
 end
 c2 = isposdef(W)
 c3 = real(W[1, 1]) == 1.06^2;
--- a/docs/src/examples/optimization_with_complex_variables/power_flow_optimization.jl
+++ b/docs/src/examples/optimization_with_complex_variables/power_flow_optimization.jl
@@ -18,7 +18,7 @@ W = ComplexVariable(n, n);
 objective = real(sum(diag(W)));
 c1 = Constraint[];
 for i in 2:n
-    push!(c1, sum(W[i, :] .* (Y[i, :]')) == inj[i])
+    push!(c1, sum(W[i, :] .* conj(Y[i, :])) == inj[i])
 end
 c2 = isposdef(W)
 c3 = real(W[1, 1]) == 1.06^2;
--- a/docs/src/examples/optimization_with_complex_variables/power_flow_optimization.jl
+++ b/docs/src/examples/optimization_with_complex_variables/power_flow_optimization.jl
@@ -18,7 +18,7 @@ W = ComplexVariable(n, n);
 objective = real(sum(diag(W)));
 c1 = Constraint[];
 for i in 2:n
-    push!(c1, sum(W[i, :] .* conj(Y[i, :])) == inj[i])
+    push!(c1, Y[i, :]' * W[i, :] == inj[i])
 end
 c2 = isposdef(W)
 c3 = real(W[1, 1]) == 1.06^2;
--- a/src/atoms/IndexAtom.jl
+++ b/src/atoms/IndexAtom.jl
@@ -11,11 +11,7 @@ mutable struct IndexAtom <: AbstractExpr
     inds::Union{AbstractArray,Nothing}
 end
 
-function IndexAtom(
-    x::AbstractExpr,
-    rows::AbstractArray,
-    cols::AbstractArray,
-)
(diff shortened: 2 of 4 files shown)


src/atoms/IndexAtom.jl
```diff
@@ -9,18 +9,14 @@ mutable struct IndexAtom <: AbstractExpr
...
+function IndexAtom(x::AbstractExpr, rows::AbstractArray, cols::AbstractArray)
+    return IndexAtom((x,), (length(rows), length(cols)), rows, cols, nothing)
+end
 
-    function IndexAtom(x::AbstractExpr, inds::AbstractArray)
-        return new((x,), (length(inds), 1), nothing, nothing, inds)
-    end
+function IndexAtom(x::AbstractExpr, inds::AbstractArray)
+    return IndexAtom((x,), (length(inds), 1), nothing, nothing, inds)
 end
```
ericphanson, line 20: I made these outer constructors to not suppress the creation of the default inner constructor, so I could use it in the bugfix

odow:
The doc failure seems suspicious.

odow:
This isn't quite correct. Here's a reproducer from the docs:
```Julia
julia> using Convex

julia> import MathOptInterface as MOI

julia> x = Variable(2, 2)
Variable
size: (2, 2)
sign: real
vexity: affine
id: 885…241

julia> context = Convex.Context{Float64}(MOI.Utilities.Model{Float64});

julia> Y = [1 2; 3 4]
2×2 Matrix{Int64}:
 1  2
 3  4

julia> Convex.to_vaf(Convex.conic_form!(context, x[1, :] .* Y[1, :]'))
┌                              ┐
│0.0 + 1.0 MOI.VariableIndex(1)│
│0.0 + 1.0 MOI.VariableIndex(3)│
│0.0 + 2.0 MOI.VariableIndex(1)│
│0.0 + 2.0 MOI.VariableIndex(3)│
└                              ┘

julia> Convex.to_vaf(Convex.conic_form!(context, x[1:1, :] .* Y[1, :]'))
┌                              ┐
│0.0 + 1.0 MOI.VariableIndex(1)│
│0.0 + 2.0 MOI.VariableIndex(3)│
└                              ┘
```

These two calls should be the same.

odow:
Okay, perhaps the docs are wrong, because:
```Julia
julia> a = [1, 2]
2-element Vector{Int64}:
 1
 2

julia> a .* a'
2×2 Matrix{Int64}:
 1  2
 2  4
```

docs/src/examples/optimization_with_complex_variables/power_flow_optimization.jl
```diff
@@ -18,7 +18,7 @@ W = ComplexVariable(n, n);
 objective = real(sum(diag(W)));
 c1 = Constraint[];
 for i in 2:n
-    push!(c1, sum(W[i, :] .* (Y[i, :]')) == inj[i])
+    push!(c1, sum(W[i, :] .* conj(Y[i, :])) == inj[i])
```
odow, line 21: Ooooofff. This is pretty subtle...

odow (approved):

odow (approved):
I think this is worth it. The next version might break existing code, but it's arguably a bug fix because the current code departs from regular Julia. There's a strong chance that any breakage is likely a bug to begin with.

docs/src/examples/optimization_with_complex_variables/power_flow_optimization.jl
ericphanson: Good catch! Btw maybe this could be `dot`? It used to be bugged in the complex case so maybe the original author didn’t use it

src/problem_depot/problems/affine.jl
```diff
@@ -280,6 +280,24 @@ end
         @test p.optval ≈ 147 atol = atol rtol = rtol
     end
 
+    # *Scalar* indexing creates a column vector
```
ericphanson, line 283: This test is probably misplaced since it doesn’t use a solver

docs/src/examples/optimization_with_complex_variables/power_flow_optimization.jl
odow: Updated to use `dot`. Yeah, I assume the previous code was just trial and error to find what worked.

src/problem_depot/problems/affine.jl
odow: I will move

PR #632, opened 2024-05-06
Add default fallbacks for head and vexity with MOI sets

### LINKED ISSUE #630: Should GenericConstraint constrain the sets?
It seems each one needs a definition of vexity and head, so plugging an arbitrary set in will give confusing errors to users. Maybe we can check this and give a nicer error? Maybe an allowlist or to be more extensible, we could use hasmethod to try to detect if the set is supported?

odow:
Closes #630

One possible solution to discuss.

 * The `head` method is reasonable.
 * The `vexity` method is up for debate. There are arguments for and against.

### DIFF
--- a/src/constraints/GenericConstraint.jl
+++ b/src/constraints/GenericConstraint.jl
@@ -50,15 +50,38 @@ end
 
 vexity(c::GenericConstraint) = vexity(vexity(c.child), c.set)
 
-# The default vexity for MOI sets is assumed to required affine functions and be
-# convex. This is for the most common case of the conic sets. There might be
-# some incorrect answers, like if the user adds something like the `MOI.SOS1` or
-# `MOI.Complements`, but this shouldn't really be relevant for Convex.jl?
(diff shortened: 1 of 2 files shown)


ericphanson:
The default vexities worry me :/

src/constraints/GenericConstraint.jl
```diff
@@ -46,6 +50,17 @@ end
 
 vexity(c::GenericConstraint) = vexity(vexity(c.child), c.set)
 
+# The default vexity for MOI sets is assumed to required affine functions and be
+# convex. This is for the most common case of the connic sets. There might be
```
blegat, line 54:
```suggestion
# convex. This is for the most common case of the conic sets. There might be
```

src/constraints/GenericConstraint.jl
```diff
@@ -46,6 +50,17 @@ end
...
+# some incorrect answers, like if the user adds something like the `MOI.SOS1` or
```
blegat, line 55: In that case, I think it's better to error than do something incorrect. I would simply have an error as default

ericphanson (approved):

PR #633, opened 2024-05-07
Report compilation stats when silent_solver=false

odow:
x-ref https://github.com/jump-dev/Convex.jl/pull/631/files#r1591749619

### DIFF
--- a/src/solution.jl
+++ b/src/solution.jl
@@ -85,7 +85,11 @@ function solve!(
     if problem_vexity(p) in (ConcaveVexity(), NotDcp())
         throw(DCPViolationError())
     end
-    context = Context(p, optimizer_factory)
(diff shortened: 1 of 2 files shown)


ericphanson:
This seems good to me. Currently `silent_solver` just sets `MOI.Silent`; if the user set `MOI.Silent` themselves on the optimizer, we would still print here. I think maybe that's fine? The other thing is that `silent_solver` emphasizes the `solver` part, and this isn't exactly the solver. I wonder if we want to deprecate `silent_solver` to `silent` and print this whenever `silent==false`.

odow:
We could deprecate `silent_solver` in a different PR.

I think this info might actually spur people to report performance issues, particularly if it takes a small time to solve but a long time to compile.

PR #636, opened 2024-05-08
Add OptimizationSenseAtom

### LINKED ISSUE #310: Partially specified problems

odow:
Closes #310

### DIFF
--- a/src/atoms/OptimizationSenseAtom.jl
+++ b/src/atoms/OptimizationSenseAtom.jl
@@ -1,0 +1,39 @@
+# Copyright (c) 2014: Madeleine Udell and contributors
+#
+# Use of this source code is governed by a BSD-style license that can be found
+# in the LICENSE file or at https://opensource.org/license/bsd-2-clause
+
+mutable struct OptimizationSenseAtom <: AbstractExpr
+    children::Tuple{AbstractExpr}
+    sense::MOI.OptimizationSense
+    size::Tuple{Int,Int}
+
+    function OptimizationSenseAtom(
+        x::AbstractExpr,
+        sense::MOI.OptimizationSense,
+    )
+        @assert sense != MOI.FEASIBILITY_SENSE
+        return new((x,), sense, size(x))
+    end
+end
+
+head(io::IO, x::OptimizationSenseAtom) = print(io, x.sense)
+
+Base.sign(x::OptimizationSenseAtom) = sign(only(x.children))
+
+monotonicity(x::OptimizationSenseAtom) = (Nondecreasing(),)
+
+function curvature(x::OptimizationSenseAtom)
+    if x.sense == MOI.MIN_SENSE
+        return ConvexVexity()
+    else
+        @assert x.sense == MOI.MAX_SENSE
+        return ConcaveVexity()
+    end
+end
+
+evaluate(x::OptimizationSenseAtom) = evaluate(only(x.children))
+
+function new_conic_form!(context::Context, x::OptimizationSenseAtom)
+    return new_conic_form!(context, only(x.children))
+end
--- a/test/test_atoms.jl
+++ b/test/test_atoms.jl
@@ -1940,6 +1940,41 @@ function test_RationalNormAtom_less_than_1()
     return
 end
 
+function test_OptimizationSenseAtom()
+    t = Variable(Convex.Positive())
+    atom = Convex.OptimizationSenseAtom(t, MOI.MAX_SENSE)
+    @test sprint(Convex.head, atom) == "MAX_SENSE"
+    @test sign(atom) == Convex.Positive()
+    @test vexity(atom) == Convex.ConcaveVexity()
+    atom = Convex.OptimizationSenseAtom(t, MOI.MIN_SENSE)
+    @test sprint(Convex.head, atom) == "MIN_SENSE"
+    @test vexity(atom) == Convex.ConvexVexity()
+    t.value = 2
+    @test evaluate(atom) == 2
+    return
+end
+
+function test_OptimizationSenseAtom_lamba_min()
+    function lamb_min(A::Convex.AbstractExpr)
+       t = Variable()
+       n = size(A, 1)
+       @assert n == size(A, 2)
+       add_constraint!(t, A - t * LinearAlgebra.I(n) ⪰ 0)
+       return Convex.OptimizationSenseAtom(t, MOI.MAX_SENSE)
+    end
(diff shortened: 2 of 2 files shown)


odow:
I don't know what's up with nightly

blegat:
Looks good, maybe add some docstring for what this is used for ? IIUC, this is to get an error if we don't use it the right way, e.g.,
```julia
t = Variable()
add_constraint!(t, t >= x)
add_constraint!(t, t >= -x)
maximize(t)
```
Here, because we maximize `t`, our formulation does not yield the absolute value anymore
Now if you do
```julia
maximize(OptimizationSenseAtom(t, MOI.MIN_SENSE))
```
then you get an error saying it's not DCP because `t` is convex.

ericphanson:
Looking at this more, I think in #504 I did a bunch of the parts to make this work with `Problems`, which gives us the nicer syntax of `minimize`/`maximize`, and potentially more natural places to put constraints:
```julia
using Convex, LinearAlgebra, Clarabel
using Convex: AbstractExpr

# monkeypatch the existing `getproperty`
function Base.getproperty(p::Problem, s::Symbol)
    if s === :optval
        if getfield(p, :status) == Convex.MOI.OPTIMIZE_NOT_CALLED
            return nothing
        else
            return Convex.objective_value(p)
        end
    elseif s === :size
        return p.objective.size
    end
    return getfield(p, s)
end

function lamb_min(A::AbstractExpr)
    t = Variable()
    n = size(A, 1)
    n == size(A,2) || throw(ArgumentError())
    p = maximize(t, A - t*Matrix(1.0I, n, n) ⪰ 0)
    return p
end

p = maximize( lamb_min(A) + 1, [ A >= 0, A[1,1] == 2.0] )

solve!(p, Clarabel.Optimizer)
```

```julia
julia> print(p.model)
Maximize ScalarAffineFunction{Float64}:
 1.0 + 1.0 v[5]

Subject to:

VectorAffineFunction{Float64}-in-Zeros
 ┌               ┐
 │-2.0 + 1.0 v[1]│
 └               ┘ ∈ Zeros(1)

VectorAffineFunction{Float64}-in-Nonnegatives
 ┌              ┐
 │0.0 + 1.0 v[1]│
 │0.0 + 1.0 v[2]│
 │0.0 + 1.0 v[3]│
 │0.0 + 1.0 v[4]│
 └              ┘ ∈ Nonnegatives(4)

VectorAffineFunction{Float64}-in-PositiveSemidefiniteConeSquare
 ┌                                    ┐
 │0.0 + 1.0 v[1] - 1.0 v[5]           │
 │0.0 + 1.0 v[3] + 1.0 v[2] - 1.0 v[3]│
 │0.0 + 1.0 v[3]                      │
 │0.0 + 1.0 v[4] - 1.0 v[5]           │
 └                                    ┘ ∈ PositiveSemidefiniteConeSquare(2)
```

However, it's not fully correct yet, since if I flip `maximize` to `minimize` in the definition of `lamb_min` I don't get an error.

IMO if we get this working, it will be easier for users than introducing a new atom.

odow:
Okay let me take a look

odow:
Closing for now. I'll take another shot at this.

PR #645, opened 2024-05-08
[breaking] Cleanup the `Context` object: remove `id_hash`, simplify dictionaries

### LINKED ISSUE #488: Different result each time problem is solved

ericphanson:
We are using `objectid` as our hash, since there are no collisions thanks to the Julia runtime. We are using this as keys in dictionaries inside the `Context` object. However, an `IdDict` is exactly a dict that uses `objectid` for lookup. This also makes making new AbstractVariables a bit easier, although they still need to be mutable. (Before they had to be mutable and you had to set the `id_hash` yourself).

I don't expect this to necessarily have any performance implications, but `IdDict`s should be quite fast (since there's no hashing involved), and I think this should reduce the likelihood of bugs. E.g. I noticed in `ComplexConstant` we were generating the `id_hash` randomly (for some reason), which does have the chance of collision. We are still fairly type unstable, as before.

## Breakingness

I tagged this breaking since the instructions on how to create a custom AbstractVariable have changed. However, I don't think it is very breaking in practice; if you followed the old advice and had a mutable object with an `id_hash` field, it should still work fine, we just no longer inspect the `id_hash` field. However if you had an immutable variable type, and you distinguished instances via the `id_hash` value, now you will get collisions. I think there are essentially 0 users of the AbstractVariable interface in the wild though, so IMO it's worth cleaning up now.

## Aside on ordering

This PR has the potential to regress on https://github.com/jump-dev/Convex.jl/issues/488, since this drops the ordering from some dictionaries, and we don't have a test for #488. However, I've looked through the code, and I don't think we actually rely on dictionary iteration ordering **anywhere** with these dictionaries. We only iterate to retrieve results from MOI, not to add variables/constraints (since those get added during `conic_form!` whose order is determined by the problem formulation).

The PR that fixed #488 also changed a dictionary inside `Hcat`, which we definitely *did* use iteration order to choose the order of objects that got added to the model:

https://github.com/jump-dev/Convex.jl/blob/4f52b6fd4c446cda55e77201037ae22043da2c67/src/atoms/affine/stack.jl#L87

I suspect this was the real fix. We no longer build the hcat atom in such a complicated way, and this dictionary has been deleted as far as I can tell.

I did run the problem from #488 a couple times (recreating it from scratch) and got the same results (even the same values of the variable `X1` and the objective), as well.

### DIFF
--- a/docs/src/examples/mixed_integer/aux_files/antidiag.jl
+++ b/docs/src/examples/mixed_integer/aux_files/antidiag.jl
@@ -14,9 +14,8 @@ export antidiag
 
 ### Diagonal
 ### Represents the kth diagonal of an mxn matrix as a (min(m, n) - k) x 1 vector
-struct AntidiagAtom <: Convex.AbstractExpr
+mutable struct AntidiagAtom <: Convex.AbstractExpr
     head::Symbol
-    id_hash::UInt64
     children::Tuple{Convex.AbstractExpr}
     size::Tuple{Int,Int}
     k::Int
@@ -29,7 +28,6 @@ struct AntidiagAtom <: Convex.AbstractExpr
         children = (x,)
         return new(
             :antidiag,
-            hash((children, k)),
             children,
             (minimum(x.size) - k, 1),
             k,
--- a/docs/src/examples/mixed_integer/aux_files/antidiag.jl
+++ b/docs/src/examples/mixed_integer/aux_files/antidiag.jl
@@ -26,12 +26,7 @@ mutable struct AntidiagAtom <: Convex.AbstractExpr
             error("Bounds error in calling diag")
         end
         children = (x,)
-        return new(
-            :antidiag,
-            children,
-            (minimum(x.size) - k, 1),
-            k,
-        )
+        return new(:antidiag, children, (minimum(x.size) - k, 1), k)
     end
 end
 
--- a/docs/src/manual/advanced.md
+++ b/docs/src/manual/advanced.md
@@ -131,16 +131,15 @@ this. To do so, we define
 
 ```@example 1
 using Convex
+
+# Must be mutable! Otherwise variables with the same size/value would be treated as the same object.
 mutable struct ProbabilityVector <: Convex.AbstractVariable
     head::Symbol
-    id_hash::UInt64
-    size::Tuple{Int, Int}
+    size::Tuple{Int,Int}
     value::Union{Convex.Value,Nothing}
     vexity::Convex.Vexity
     function ProbabilityVector(d)
-        this = new(:ProbabilityVector, 0, (d,1), nothing, Convex.AffineVexity())
-        this.id_hash = objectid(this)
-        this
+        return new(:ProbabilityVector, (d, 1), nothing, Convex.AffineVexity())
     end
 end
 
@@ -165,8 +164,8 @@ solve!(prob, SCS.Optimizer)
 evaluate(p) # [1.0, 0.0, 0.0]
 ```
 
-Subtypes of `AbstractVariable` must have the fields `head`, `id_hash`, and
-`size`, and `id_hash` must be populated as shown in the example. Then they must also
+Subtypes of `AbstractVariable` must have the fields `head` and
+`size`. Then they must also
 
 * either have a field `value`, or implement [`Convex._value`](@ref) and
   [`Convex.set_value!`](@ref)
--- a/src/Context.jl
+++ b/src/Context.jl
@@ -8,15 +8,13 @@ mutable struct Context{T,M}
     model::M
 
     # Used for populating variable values after solving
-    var_id_to_moi_indices::OrderedCollections.OrderedDict{
-        UInt64,
+    var_to_moi_indices::IdDict{
+        Any,
         Union{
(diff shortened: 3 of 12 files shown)


odow (approved):
Seems okay. I don't know if I truly understand the implications, but it looks sensible.

odow:
I took a second look at the ordering: I concur that we do not rely on the ordering for these dicts.

PR #646, opened 2024-05-09
Add support for using Problem as an atom

odow:
Take II at #310 

See also #636

### DIFF
--- a/src/problems.jl
+++ b/src/problems.jl
@@ -95,7 +95,7 @@ end
 
 function vexity(p::Problem)
     if p.head == :satisfy
-        return vexity(p.objective)
+        return ConstVexity()
     elseif p.head == :minimize
         return vexity(p.objective) + ConvexVexity()
     elseif p.head == :maximize
--- a/src/problems.jl
+++ b/src/problems.jl
@@ -34,6 +34,9 @@ function Base.getproperty(p::Problem, s::Symbol)
         else
             return objective_value(p)
         end
+    elseif s === :size
+        # Used when Problem is interpreted as an atom
+        return p.objective.size
     end
     return getfield(p, s)
 end
@@ -90,20 +93,15 @@ function objective_vexity(p::Problem)
     return obj_vex
 end
 
-vexity(p::Problem) = objective_vexity(p)
-
-# is this right?
-# function problem_vexity(sense, obj_vexity, constr_vexity)
-#     if constr_vexity == ConstVexity()
-#         return obj_vexity
-#     end
-
-#     if constr_vexity == AffineVexity() && obj_vexity == AffineVexity()
-#         return sense === :maximize ? ConcaveVexity() : ConvexVexity()
-#     end
-
-#     return obj_vexity + constr_vexity
-# end
+function vexity(p::Problem)
+    if p.head == :satisfy
+        return vexity(p.objective)
+    elseif p.head == :minimize
+        return vexity(p.objective) + ConvexVexity()
+    elseif p.head == :maximize
+        return vexity(p.objective) + ConcaveVexity()
+    end
+end
 
 function monotonicity(p::Problem)
     if p.head === :satisfy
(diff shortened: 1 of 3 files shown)


src/problems.jl
```diff
@@ -90,20 +93,15 @@ function objective_vexity(p::Problem)
...
-#         return obj_vexity
-#     end
-
-#     if constr_vexity == AffineVexity() && obj_vexity == AffineVexity()
-#         return sense === :maximize ? ConcaveVexity() : ConvexVexity()
-#     end
-
-#     return obj_vexity + constr_vexity
-# end
+function vexity(p::Problem)
```
odow, line 96: So `vexity(::Problem)` is really just "what is the verity of the objective function"

There is also `objective_vexity` which returns `Convex` if you are maximizing a concave function, and there is `problem_vexity` which also incorporates the constraints.

`solve!` uses `problem_vexity` to check for DCP-ness.

src/problems.jl
```diff
@@ -90,20 +93,15 @@ function objective_vexity(p::Problem)
...
+    if p.head == :satisfy
+        return vexity(p.objective)
```
odow, line 98:
```suggestion
        return ConstVexity()
```

src/problems.jl
ericphanson: thank you for sorting this out!

ericphanson (commented):
the code changes look good.

I think we should also test non-scalar "objectives", since those should be allowed in this setting (but multiobjective in general is not supported). Similar to the last example of https://cvxr.com/cvx/doc/advanced.html#new-functions-via-partially-specified-problems. Taking our `huber` atom and writing it as a partially specified problem:

```julia
function huber(c, M=1.0)
    s = Variable(size(c))
    n = Variable(size(c))
    return minimize(square(s) + 2 * M * abs(n), c == s + n)
end
```

odow:
Okay, so the benefit of writing out `huber` like that is simplicity: it becomes a reformulation instead of an atom.

The downside is that multiple calls to `huber(x)` won't be cached and re-used in the same way they would it if was an atom? And there will also be more nodes in the expression graph of the problem?

That's an interesting trade-off.

ericphanson:
so I think actually they will be cached, since the `Problem` is an `AbstractExpr`, and we cache the `conic_form` for all `AbstractExpr`'s. (We cache if you have the same object in multiple places, like `x= huber(...)` and you use `x` in multiple places, but not if two instances are just `isequal` but distinct objects, since I think that wouldn't be fully correct in some cases, and also we would need to hash which can be slow).

So I think the main limitation is that there isn't a way (currently) to promise additional DCP properties that can't be derived (like maybe it is positive or increasing or something like that but it couldn't be derived automatically), which we can "inject" with atoms by defining `sign` etc.  There also isn't the opportunity to choose how to lower to the MOI level (with a custom `new_conic_form!`) which lets us do fast scalar stuff and add constraints that don't have Convex.jl-level syntax etc, though hopefully that isn't needed frequently.

odow:
Added a test for caching and the array return.

ericphanson (approved):

test/test_atoms.jl
```diff
@@ -2265,6 +2265,67 @@ function test_transpose()
...
+    x = Variable(2)
+    @test size(op_huber(x)) == (2, 1)
+    context_op_huber =
+        Convex.Context(minimize(sum(op_huber(x))), MOI.Utilities.Model{Float64})
+    context_builtin =
+        Convex.Context(minimize(sum(huber(x))), MOI.Utilities.Model{Float64})
+    # This test checks that the two printed models are identical. This relies on
+    # the conic_form! of HuberAtom being identical to the implementation in
+    # op_huber. Feel free to change this test if HuberAtom changes.
+    @test sprint(print, context.model) == sprint(print, context_builtin.model)
```
ericphanson:
```suggestion
    @test sprint(print, context_op_huber.model) == sprint(print, context_builtin.model)
```

PR #647, opened 2024-05-09
Fix ambiguities found by Aqua

### LINKED ISSUE #639: Fix Aqua ambiguities
This would catch stuff like #637

odow:
Closes #639


src/atoms/VcatAtom.jl
```diff
@@ -127,3 +127,7 @@ function Base.getindex(x::VcatAtom, inds::AbstractVector{<:Real})
     remaining = VcatAtom(keep_children...)
     return IndexAtom(remaining, inds)
 end
+
+function Base.getindex(x::VcatAtom, inds::AbstractVector{Bool})
+    return getindex(x, first.(filter!(last, collect(enumerate(inds)))))
```
ericphanson, line 132: I think this is just `findall` btw

PR #648, opened 2024-05-09
Refactor GeometricMeanHypoCone into GenericConstraint

blegat:


src/constraints/GeometricMeanEpiCone.jl
```diff
@@ -37,13 +37,18 @@ struct GeometricMeanEpiConeSquare <: MOI.AbstractVectorSet
     end
 end
 
-MOI.dimension(set::GeometricMeanEpiConeSquare) = 3 * set.side_dimension^2
+const _GeometricMeanConeSquare = Union{GeometricMeanEpiConeSquare,GeometricMeanHypoConeSquare}
```
odow, line 40: Can we _not_ do this? Let's keep the implementations separate.

blegat: So we even have to copy-paste `_get_matrices` ? Once these will be MOI sets with bridges, these will get so short that they can be moved to `GenericConstraint.jl` in which case we will be able to do the `Union` no ?

odow: > Once these will be MOI sets with bridges, these will get so short that they can be moved to GenericConstraint.jl in which case we will be able to do the Union no 

How they get implemented in MOI is a separate question. I'd just prefer not to complicate this

PR #650, opened 2024-05-10
Add summary printing to show(::IO, ::Problem)

### LINKED ISSUE #640: Improve `Problem` printing to indicate size/complexity
IMO it would be useful if the printing for a `Problem` showed something like the total size of variables/constraints/etc. I suppose we don't really know that fully until the MOI level, but we could provide some kind of Convex.jl-level estimate by recursing the tree and counting things. In particular, I think these would be useful:

- number of atoms: this can show if some scalar indexing produced 10k atoms or something
- number of AbstractVariable objects: likewise, the number of variables should be relatively small, even if the size is large
- number of constraint objects: same
- sum of `length` over all variables (with 2x for complex ones): this is the total variable "volume"
- sum of `size` over all constraint sets
- sum of `nnz` for sparse constants found in the problem
- sum of `length` for dense constants found in the problem. Too big here can indicate incorrect densification.
- `Base.summarysize` for the whole problem: Flux does this, and it seems like a useful bit of info.

These would serve two purposes:
- provide info to the user before solve-time about how big their problem is
- provide information to notice/diagnose performance bugs

Ideally we could provide a relatively compact printing with this information.

odow:
Alternative for #643

Closes #640 

## Examples

```Julia
julia> A = [1 2im 3 4; 4im 3im 2 1; 4 5 6 7]
3×4 Matrix{Complex{Int64}}:
 1+0im  0+2im  3+0im  4+0im
 0+4im  0+3im  2+0im  1+0im
 4+0im  5+0im  6+0im  7+0im

julia> y = ComplexVariable(3, 4)
Variable
size: (3, 4)
sign: complex
vexity: affine
id: 120…842

julia> p = minimize(nuclearnorm(y), y == A)
Problem statistics
  number of variables    : 1 (24 scalar elements)
  number of constraints  : 1 (24 scalar elements)
  number of coefficients : 24
  number of atoms        : 2

Solution summary
  termination status : OPTIMIZE_NOT_CALLED
  primal status      : NO_SOLUTION
  dual status        : NO_SOLUTION

Expression graph
  minimize
   └─ nuclearnorm (convex; positive)
      └─ 3×4 complex variable (id: 120…842)
  subject to
   └─ == constraint (affine)
      └─ + (affine; complex)
         ├─ 3×4 complex variable (id: 120…842)
         └─ 3×4 complex constant
```

```Julia
julia> x = Variable(1000)
Variable
size: (1000, 1)
sign: real
vexity: affine
id: 153…767

julia> for i in 1:1000
       add_constraint!(x, x[i] >= 0)
       end

julia> p = minimize(sum(x))
Problem statistics
  number of variables    : 1 (1_000 scalar elements)
  number of constraints  : 1_000 (1_000 scalar elements)
  number of coefficients : 1000
  number of atoms        : 2001

Solution summary
  termination status : OPTIMIZE_NOT_CALLED
  primal status      : NO_SOLUTION
  dual status        : NO_SOLUTION

Expression graph
  minimize
   └─ sum (affine; real)
      └─ 1000-element real variable (id: 245…329)
  subject to
   ├─ ≥ constraint (affine)
   │  └─ + (affine; real)
   │     ├─ index (affine; real)
   │     │  └─ …
   │     └─ [0;;]
```


test/test_utilities.jl
```diff
@@ -171,6 +171,13 @@ function test_show()
     @test curvature(p) == Convex.ConvexVexity()
 
     @test sprint(show, p) == """
+    summary
+    ├─ # variables    : 1 (1 scalar elements)
+    ├─ # constraints  : 2 (2 scalar elements)
+    ├─ # coefficients : 2
+    ├─ # atoms        : 3
+    └─ size           : 624 bytes
```
odow, line 179: I guess my question is: what information is useful here?

It's handy to know if:
 * There are lots of variables, both matrix and scalar valued
 * There are lots of constraints, both matrix and scalar valued
 * If there are lots of data matrices (`# coefficients` is the length of `Matrix` including zeros plus non-zeros in a sparse matrix)
 * There are lots of atoms (big expressions tree)
 * The problem uses a lot of memory

Things like what cones it lowers to etc are a bit more abstract. We should probably add default printing to MOI.

ericphanson:
hm, interesting. Some thoughts

* I agree with merging number of dense elements and sparse to just number of coefficients
* I agree with deferring MOI stuff to MOI. Could be nice if we could just `show` the model post-formulation or something.
* this printing is less compact, but has more alignment, so maybe it is easier to read
* this `Counts` implementation looks less "elegant" to me (more direct/manual, less composition of abstractions), but maybe that's for the best
* using `#` for "number" is confusing to me, since it gets highlighted like a comment (for awhile I was wondering "why isn't the `size` one commented?"
    * also flux uses `#` to inject counts as comments (see below), so I probably had that expectation
* the re-use of the tree-printing characters to show the summary is kinda interesting - makes me think maybe we should merge the counts with the tree, like Flux does (though they have flatter structures usually):
```julia
Chain(
  Dropout(0.0, active=false),
  Conv((3, 3), 1 => 32, relu),          # 320 parameters
  BatchNorm(32, relu, active=false),    # 64 parameters, plus 64
  MaxPool((2, 2)),
  Dropout(0.0, active=false),
  Conv((3, 3), 32 => 16, relu),         # 4_624 parameters
  Dropout(0.0, active=false),
  MaxPool((2, 2)),
  Dropout(0.0, active=false),
  Conv((3, 3), 16 => 10, relu),         # 1_450 parameters
  Dropout(0.0, active=false),
  var"#22#23"(),
  Dropout(0.0, active=false),
  Dense(90 => 10),                      # 910 parameters
  Dense(10 => 10),                      # 110 parameters
  Dense(10 => 10),                      # 110 parameters
  NNlib.softmax,
)         # Total: 12 trainable arrays, 7_478 parameters,
          # plus 4 non-trainable, 174 parameters, summarysize 31.431 KiB.
```

ericphanson:
I think 
```julia
summary
├─ # variables    : 1 (1_000 scalar elements)
├─ # constraints  : 1_000 (1_000 scalar elements)
├─ # coefficients : 1000
├─ # atoms        : 2001
└─ size           : 265.836 KiB

minimize
└─ sum (affine; real)
   └─ 1000-element real variable (id: 153…767)
subject to
├─ ≥ constraint (affine)
│  └─ + (affine; real)
```

would make more sense to me if you could do `problem.summary` or something to get a `ProblemSummary` object, and that object had those numbers and nice printing. Because to me it looks like we are saying: "this problem has a `summary` with these entries, and `minimize` with these children".

(In terms of implementation, I don't think a field totally makes sense, bc I don't think we want to actually cache the summary since then we have to worry about invalidation, but it could be `get_summary(problem)` or something).

odow:
What if it said "summary statistics" or "problem statistics"?

ericphanson:
what about something more distinct from the tree printing?

```julia
julia> p = minimize(nuclearnorm(y), y == A)
┌──────────────┬────────────────────────┐
│    variables │ 1 (24 scalar elements) │
│  constraints │ 1 (24 scalar elements) │
│ coefficients │ 24                     │
│        atoms │ 2                      │
│         size │ 704 bytes              │
└──────────────┴────────────────────────┘
minimize
└─ nuclearnorm (convex; positive)
   └─ 3×4 complex variable (id: 446…935)
subject to
└─ == constraint (affine)
   └─ + (affine; complex)
      ├─ 3×4 complex variable (id: 446…935)
      └─ 3×4 complex constant

```

(this I mocked up with a dataframe and `PrettyTables.pretty_table(stdout, df;  alignment=[:r, :l], show_subheader=false, show_header=false, tf=tf_unicode)`).

odow:
What about:
```julia
julia> p
Problem statistics
  number of variables    : 1 (4 scalar elements)
  number of constraints  : 5 (14 scalar elements)
  number of coefficients : 36
  number of atoms        : 19
  memory allocated       : 2.727 KiB

Solution summary
  Termination status : OPTIMAL
  Primal status      : FEASIBLE_POINT
  Dual status        : FEASIBLE_POINT
  objective value    : 9.9998

Expression graph
  minimize
   └─ sum (affine; real)
      └─ reshape (affine; real)
         └─ * (affine; real)
            ├─ …
            └─ …
  subject to
   ├─ ≤ constraint (affine)
   │  └─ + (affine; real)
   │     ├─ * (affine; real)
   │     │  ├─ …
   │     │  └─ …
   │     └─ 4×1 Matrix{Int64}
   ├─ ≥ constraint (affine)
   │  └─ + (affine; real)
   │     ├─ 4-element real variable (id: 104…638)
   │     └─ Convex.NegateAtom (constant; negative)
   │        └─ …
   ├─ ≤ constraint (affine)
   │  └─ + (affine; real)
   │     ├─ 4-element real variable (id: 104…638)
   │     └─ Convex.NegateAtom (constant; negative)
   │        └─ …
   ├─ ≤ constraint (affine)
   │  └─ + (affine; real)
   │     ├─ index (affine; real)
   │     │  └─ …
   │     └─ [-5;;]
   └─ ≤ constraint (affine)
      └─ + (affine; real)
         ├─ index (affine; real)
         │  └─ …
         ├─ index (affine; real)
         │  └─ …
         ├─ Convex.NegateAtom (affine; real)
         │  └─ …
         └─ [-10;;]
```

ericphanson:
I like it! 

One tiny nitpick is that memory allocated is potentially confusing, since more memory was likely allocated on the way, but this what we’re still using.

otherwise lgtm

odow:
> memory allocated is potentially confusing

:+1:

What about explicitly quantifying the size of the expression graph and the size of the MOI backend?

odow:
One issue with `optimization model : XXX bytes` is that it doesn't count the memory allocated in C by the solver

odow:
Maybe we should just remove the memory stuff. The rest is useful enough, and specifying the memory is just going to mislead people.

odow:
Okay, this now yields:
```Julia
julia> A = [1 2im 3 4; 4im 3im 2 1; 4 5 6 7]
3×4 Matrix{Complex{Int64}}:
 1+0im  0+2im  3+0im  4+0im
 0+4im  0+3im  2+0im  1+0im
 4+0im  5+0im  6+0im  7+0im

julia> y = ComplexVariable(3, 4)
Variable
size: (3, 4)
sign: complex
vexity: affine
id: 120…842

julia> p = minimize(nuclearnorm(y), y == A)
Problem statistics
  number of variables    : 1 (24 scalar elements)
  number of constraints  : 1 (24 scalar elements)
  number of coefficients : 24
  number of atoms        : 2

Solution summary
  termination status : OPTIMIZE_NOT_CALLED
  primal status      : NO_SOLUTION
  dual status        : NO_SOLUTION

Expression graph
  minimize
   └─ nuclearnorm (convex; positive)
      └─ 3×4 complex variable (id: 120…842)
  subject to
   └─ == constraint (affine)
      └─ + (affine; complex)
         ├─ 3×4 complex variable (id: 120…842)
         └─ 3×4 complex constant
```

src/utilities/tree_interface.jl
```diff
@@ -38,3 +38,97 @@ function AbstractTrees.printnode(io::IO, node::Constant)
...
+    counts.n_constraints += 1
+    counts.n_scalar_constraints +=
+        (iscomplex(node) ? 2 : 1) * MOI.dimension(node.set)
+    _add_to_problem_count(counts, node.child)
+    return
+end
+
+function _add_to_problem_count(counts::Counts, node::Constraint)
+    counts.n_constraints += 1
+    # TODO(odow): we don't now the general case
```
ericphanson, line 101:
```suggestion
    # TODO(odow): we don't know the general case
```

src/utilities/tree_interface.jl
```diff
@@ -38,3 +38,97 @@ function AbstractTrees.printnode(io::IO, node::Constant)
...
+
+function _add_to_problem_count(counts::Counts, node::AbstractVariable)
+    if node in counts.seen
+        return
+    end
+    counts.n_variables += 1
+    counts.n_scalar_variables += (iscomplex(node) ? 2 : 1) * length(node)
+    push!(counts.seen, node)
+    for c in get_constraints(node)
+        _add_to_problem_count(counts, c)
```
ericphanson, line 82: I wonder if one could have a constraint referenced in multiple places?
```suggestion
       if !(c in counts.seen)
            _add_to_problem_count(counts, c)
        end
```

src/utilities/tree_interface.jl
```diff
@@ -38,3 +38,97 @@ function AbstractTrees.printnode(io::IO, node::Constant)
...
+
+function _add_to_problem_count(counts::Counts, node::Constraint)
+    counts.n_constraints += 1
+    # TODO(odow): we don't now the general case
+    return
+end
+
+function _add_to_problem_count(counts::Counts, node::Constant)
+    _add_to_problem_count(counts, node.value)
+    push!(counts.seen, node)
```
ericphanson, line 107: this seems strange; I think either the caller should be pushing to seen, like in `_add_to_problem_count(counts::Counts, node::AbstractExpr)` and thus this isn't needed, or each child should do it, in which case we are missing some pushes, like in `_add_to_problem_count(counts::Counts, node::GenericConstraint)`

src/utilities/show.jl
```diff
@@ -234,17 +234,41 @@ function TreePrint.print_tree(io::IO, p::Problem, args...; kwargs...)
...
-    else
-        print(io, "\ntermination status: $(p.status)")
-        print(io, "\nprimal status: $(primal_status(p))")
-        print(io, "\ndual status: $(dual_status(p))")
-    end
-    if p.status == "solved"
-        print(io, " with optimal value of $(round(p.optval, digits=4))")
+    # Print problem statistics
+    counts = Counts(p)
+    bytes = Base.summarysize(p) - Base.summarysize(p.model)
```
ericphanson, line 250:
```suggestion
```

src/utilities/tree_interface.jl
ericphanson: Ah true. I’m not sure if they can or not

odow: If I have `satisfy([c, c])`, then MOI will count them twice, so we probably should too?

Let's leave as-is for now. We can always change if someone complains.

ericphanson (approved):

PR #655, opened 2024-05-13
audit usages of `evaluate`

ericphanson:
I tried to ensure that #654 fixed the last issue here by checking each usage of `evaluate` in `src` (outside of the problem depot). I added comments to a few places, and updated the dev docs.

### DIFF
--- a/docs/src/developer/contributing.md
+++ b/docs/src/developer/contributing.md
@@ -40,6 +40,12 @@ Convex.jl. Let's say you're adding the new function $f$.
  -   Ensure the atom is a mutable struct, so that `objectid` can be called.
  -   Add as a comment a description of what the atom does and its
      parameters.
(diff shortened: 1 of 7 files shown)


docs/src/developer/contributing.md
```diff
@@ -40,6 +40,12 @@ Convex.jl. Let's say you're adding the new function $f$.
  -   Ensure the atom is a mutable struct, so that `objectid` can be called.
  -   Add as a comment a description of what the atom does and its
      parameters.
+ -   Ensure `evaluate` is only called during the definition of `evaluate` itself, from `conic_form!`, or
+     on constants or matrices. Specifically, `evaluate` must not be called on a potentially `fix!`'d variable
+     when the expression tree is being built (e.g. when constructing an atom or reformulating),
```
ericphanson, line 45:
```suggestion
     when the expression tree is being built (for example, when constructing an atom or reformulating),
```

PR #657, opened 2024-05-14
update to Documenter v1

ericphanson:


docs/src/examples/supplemental_material/Convex.jl_intro_ISMP2015.jl
```diff
@@ -196,22 +197,14 @@ sum(x) + y[2]
...
-#-
-
-## not dcp compliant
+# - not dcp compliant
 log(x) + square(x)
 
-#-
-
-## $f$ is convex increasing and $g$ is convex
+# - Composition $f\circ g$ where $f$ is convex increasing and $g$ is convex
```
ericphanson, line 203: add leading characters to workaround https://github.com/JuliaDocs/Documenter.jl/issues/1483

docs/src/examples/general_examples/basic_usage.jl
```diff
@@ -6,7 +6,7 @@ using SCS
 
 # ### Linear program
 #
-# $$
+# ```math
```
ericphanson, line 9: I couldn't get `$$` working (even playing with the mathengine stuff), so switched all display math to this, following JuMP

docs/src/examples/supplemental_material/Convex.jl_intro_ISMP2015.jl
```diff
@@ -62,24 +62,25 @@ using Interact, Plots
     solve!(problem, SCS.Optimizer; silent_solver = true)
     histogram(evaluate(x), xlims = (0, 3.5), label = "x")
 end
+nothing # hide
```
ericphanson, line 65: this doesn't produce anything that shows up in the rendered docs (and maybe never did, I think it was a notebook before), and also causes Documenter to warn about a big block that it can't render as svg. So I add `nothing` so it doesn't try to render anything.

docs/make.jl
```diff
@@ -84,20 +87,20 @@ open(joinpath(@__DIR__, "src", "changelog.md"), "r") do in_io
...
 
 # ==============================================================================
-#  Build annd release
+#  Build and release
 # ==============================================================================
 
 Documenter.makedocs(
     sitename = "Convex.jl",
-    repo = "https://github.com/jump-dev/Convex.jl/blob/{commit}{path}#L{line}",
+    repo = Documenter.Remotes.GitHub("jump-dev", "Convex.jl"),
```
odow, line 95:
```suggestion
```

I don't think we need this

docs/src/examples/general_examples/basic_usage.jl
odow: :+1:

odow: This is much clearer anyway

It might have to be `# $$\begin \n... \n# \end{array}$$`

PR #659, opened 2024-05-14
deprecate `+` on constraints

ericphanson:
IMO it is kind of goofy, and doesn't match how the rest of Julia works which can make it confusing. There is also not much of a mathematical or translation purpose; I've never seen constraints written that way in a paper.

Also, I've updated the `Problem` constructors to be less picky about the types, which makes manual construction of the constraints vector (like `constraints = []; push!(...)` less annoying). I suspect the `p = minimize(...); p.constraints += ...` thing was partly to make the constructors work.

Lastly, I fixed a small bug: the field `constraints` was typed as an `Array`, which meant you could construct a `Matrix{Constraint}` as the `constraints` field in `Problem`, which would then break the new printing code. Here I've typed it as a `Vector`.

### DIFF
--- a/docs/src/examples/optimization_with_complex_variables/povm_simulation.jl
+++ b/docs/src/examples/optimization_with_complex_variables/povm_simulation.jl
@@ -45,7 +45,7 @@ function get_visibility(K)
     constraints = Constraint[isposdef(P[i][j]) for i in 1:6 for j in 1:2]
     push!(constraints, sum(q) == 1)
     push!(constraints, t <= 1)
-    push!(constraints, [P[i][1] + P[i][2] == q[i] * I(2) for i in 1:6])
+    append!(constraints, [P[i][1] + P[i][2] == q[i] * I(2) for i in 1:6])
     push!(
         constraints,
         t * K[1] + (1 - t) * noise[1] == P[1][1] + P[2][1] + P[3][1],
--- a/docs/src/examples/optimization_with_complex_variables/povm_simulation.jl
+++ b/docs/src/examples/optimization_with_complex_variables/povm_simulation.jl
@@ -43,13 +43,25 @@ function get_visibility(K)
     q = Variable(6, Positive())
     t = Variable(1, Positive())
     constraints = [isposdef(P[i][j]) for i in 1:6 for j in 1:2]
-    constraints += sum(q) == 1
-    constraints += t <= 1
-    constraints += [P[i][1] + P[i][2] == q[i] * I(2) for i in 1:6]
-    constraints += t * K[1] + (1 - t) * noise[1] == P[1][1] + P[2][1] + P[3][1]
-    constraints += t * K[2] + (1 - t) * noise[2] == P[1][2] + P[4][1] + P[5][1]
-    constraints += t * K[3] + (1 - t) * noise[3] == P[2][2] + P[4][2] + P[6][1]
-    constraints += t * K[4] + (1 - t) * noise[4] == P[3][2] + P[5][2] + P[6][2]
+    push!(constraints, sum(q) == 1)
+    push!(constraints, t <= 1)
+    push!(constraints, [P[i][1] + P[i][2] == q[i] * I(2) for i in 1:6])
+    push!(
+        constraints,
+        t * K[1] + (1 - t) * noise[1] == P[1][1] + P[2][1] + P[3][1],
+    )
+    push!(
+        constraints,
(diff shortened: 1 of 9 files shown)


test/test_utilities.jl
```diff
@@ -1118,6 +1118,19 @@ function test_dcp_rules()
...
+    x = Variable(1, Positive())
+    # Test untyped constraints
+    p = maximize(x, Any[x<=1])
+    @test p isa Problem
+    p = maximize(x, Any[x <= 1;; x <= 1])
+    @test p isa Problem
+    p = minimize(x, Any[x<=1])
+    @test p isa Problem
+    p = minimize(x, Any[x <= 1;; x <= 1])
+    @test p isa Problem
```
odow, line 1131:
```suggestion
    @test p isa Problem
    return
```

src/deprecations.jl
```diff
@@ -123,3 +123,33 @@ end
...
+
+function Base.:+(x::Constraint, y::Array{<:Constraint})
+    warn_deprecated_constraint_concatenation()
+    return vcat(x, y)
+end
+
+function Base.:+(x::Array{<:Constraint}, y::Constraint)
+    warn_deprecated_constraint_concatenation()
+    return vcat(x, y)
+end
```
odow, line 155: We should add tests for these

src/deprecations.jl
```diff
@@ -123,3 +123,33 @@ end
...
     return ComplexVariable((1, 1), set, sets...)
 end
+
+# `+` on constraints
+function warn_deprecated_constraint_concatenation()
+    @warn(
+        "Concatenating collections of constraints together with `+` or `+=` to produce a new list of constraints is deprecated. Instead, use `vcat` to concatenate collections of constraints.",
+        maxlog = 1
+    )
+end
```
odow, line 133:
```suggestion
end

```

src/deprecations.jl
```diff
@@ -123,3 +123,33 @@ end
 function ComplexVariable(set::Symbol, sets::Symbol...)
     return ComplexVariable((1, 1), set, sets...)
 end
+
+# `+` on constraints
+function warn_deprecated_constraint_concatenation()
+    @warn(
+        "Concatenating collections of constraints together with `+` or `+=` to produce a new list of constraints is deprecated. Instead, use `vcat` to concatenate collections of constraints.",
+        maxlog = 1
+    )
```
odow, line 132:
```suggestion
    )
    return
```

src/deprecations.jl
```diff
@@ -123,3 +123,33 @@ end
...
+function Base.:+(x::Array{<:Constraint}, y::Array{<:Constraint})
+    warn_deprecated_constraint_concatenation()
+    return vcat(x, y)
+end
+
+function Base.:+(x::Constraint, y::Constraint)
+    @warn(
+        "Adding constraints together (with `+` or `+=`) to produce a list of constraints is deprecated. Instead, construct a list of constraints via `[constraint1, constraint2]`",
+        maxlog = 1
+    )
```
odow, line 143: Does this need a separate warning? You can still use `vcat(x, y)`

src/deprecations.jl
ericphanson: added

docs/src/examples/optimization_with_complex_variables/povm_simulation.jl
```diff
@@ -42,14 +42,26 @@ function get_visibility(K)
...
-    constraints += t <= 1
-    constraints += [P[i][1] + P[i][2] == q[i] * I(2) for i in 1:6]
-    constraints += t * K[1] + (1 - t) * noise[1] == P[1][1] + P[2][1] + P[3][1]
-    constraints += t * K[2] + (1 - t) * noise[2] == P[1][2] + P[4][1] + P[5][1]
-    constraints += t * K[3] + (1 - t) * noise[3] == P[2][2] + P[4][2] + P[6][1]
-    constraints += t * K[4] + (1 - t) * noise[4] == P[3][2] + P[5][2] + P[6][2]
+    constraints = Constraint[isposdef(P[i][j]) for i in 1:6 for j in 1:2]
+    push!(constraints, sum(q) == 1)
+    push!(constraints, t <= 1)
+    push!(constraints, [P[i][1] + P[i][2] == q[i] * I(2) for i in 1:6])
```
ericphanson, line 48:
```suggestion
    append!(constraints, [P[i][1] + P[i][2] == q[i] * I(2) for i in 1:6])
```

odow (approved):

PR #665, opened 2024-05-15
Rename GenericConstraint into Constraint and remove abstract type

blegat:
This is the last breaking change I'd like to make until v0.16


ericphanson:
I think it is unlikely; I just checked the 12 registered packages that use Convex.jl (via https://juliahub.com/ui/Packages/General/Convex), and none of them seem to. Also, I don't think it was ever documented how to construct a new `Constraint` type.

odow:
Okay, then I'm :+1: for this

PR #669, opened 2024-05-16
Support vector sets in MOI wrapper

blegat:
Second option of the two options listed in https://github.com/jump-dev/Convex.jl/pull/668

Closes https://github.com/jump-dev/Convex.jl/pull/668

### DIFF
--- a/src/Constraint.jl
+++ b/src/Constraint.jl
@@ -3,17 +3,17 @@
 # Use of this source code is governed by a BSD-style license that can be found
 # in the LICENSE file or at https://opensource.org/license/bsd-2-clause
 
-mutable struct Constraint{S<:MOI.AbstractSet}
+mutable struct Constraint{S<:MOI.AbstractVectorSet}
     child::AbstractExpr
     set::S
     dual::Union{Value,Nothing}
 
-    function Constraint(child::AbstractExpr, set::MOI.AbstractSet)
+    function Constraint(child::AbstractExpr, set::MOI.AbstractVectorSet)
         return new{typeof(set)}(child, set, nothing)
     end
 end
 
-function Constraint{S}(child::AbstractExpr) where {S<:MOI.AbstractSet}
+function Constraint{S}(child::AbstractExpr) where {S<:MOI.AbstractVectorSet}
     return Constraint(child, set_with_size(S, size(child)))
 end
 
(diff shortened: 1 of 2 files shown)


src/MOI_wrapper.jl
```diff
@@ -117,8 +117,8 @@ end
 
 function MOI.supports_constraint(
     ::Optimizer{T},
-    ::Type{MOI.ScalarNonlinearFunction},
-    ::Type{<:Union{MOI.EqualTo{T},MOI.GreaterThan{T},MOI.LessThan{T}}},
+    ::Type{MOI.VectorNonlinearFunction},
+    ::Type{<:MOI.AbstractVectorSet},
```
blegat, line 121: We should probably have a list of sets for which we define `vexity` instead because otherwise, if there is a set for which `vexity` is not defined but there is a bridge to convert it into sets for which `vexity` is defined, it won't be applied because Convex lied saying that `vexity` is defined

blegat: I added a comment

src/Constraint.jl
```diff
@@ -3,17 +3,17 @@
 # Use of this source code is governed by a BSD-style license that can be found
 # in the LICENSE file or at https://opensource.org/license/bsd-2-clause
 
-mutable struct Constraint{S<:MOI.AbstractSet}
+mutable struct Constraint{S<:MOI.AbstractVectorSet}
```
odow, line 6: Why is this needed?

blegat: It's not needed, I just thought we are only supporting vector sets at the moment, we can revert that part

blegat: For example, isn't `populate_dual!` always returning a vector ?

odow: Ah true. We can always extend this non-breaking in future.

odow: I guess there'd be an error like this at the moment:
```julia
julia> reshape(2.0, (1,1))
ERROR: MethodError: no method matching reshape(::Float64, ::Tuple{Int64, Int64})
```

So we'd need a scalar `populate_dual!` method, but that is do-able (once there is demand).

src/MOI_wrapper.jl
odow: We'll say `true`, and then end up calling a missing vexity and get ```"`Convex.vexity(vex, ::$(typeof(set)))`: is not yet implemented. Please open an issue at https://github.com/jump-dev/Convex.jl"```, so that seems like an okay error. It tells them to open an issue, at least.

odow (approved):
LGTM. I tweaked the comment

PR #670, opened 2024-05-17
Deprecate `silent_solver` to `silent`

ericphanson:
Since now it is being used for Convex's verbosity also (printing the formulation time). It is also shorter which I like. I also set it to turn off the warning for non-optimal solutions for consistency and to reduce clutter in test logs.

### DIFF
--- a/src/solution.jl
+++ b/src/solution.jl
@@ -98,7 +98,6 @@ function solve!(
     elseif silent === nothing
         silent = false
     end
-
     if problem_vexity(p) in (ConcaveVexity(), NotDcp())
         throw(DCPViolationError())
     end
--- a/src/solution.jl
+++ b/src/solution.jl
@@ -59,7 +59,7 @@ end
     solve!(
         problem::Problem,
         optimizer_factory;
-        silent_solver = false,
+        silent = false,
         warmstart::Bool = true,
     )
 
@@ -69,7 +69,7 @@ duals (accessed by `cons.dual`), where applicable. Returns the input `problem`.
 
 Optional keyword arguments:
 
- * `silent_solver`: whether the solver should be silent (and not emit output or
+ * `silent`: whether the solver should be silent (and not emit output or
    logs) during the solution process.
  * `warmstart` (default: `false`): whether the solver should start the
    optimization from a previous optimal value (according to the current primal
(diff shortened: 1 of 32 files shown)


ericphanson:
IMO it would be good for this to be part of v0.16.0 to reduce churn. But it isn’t breaking so it could go in after.

src/solution.jl
```diff
@@ -69,27 +69,43 @@ duals (accessed by `cons.dual`), where applicable. Returns the input `problem`.
...
+   and warn if the problem was not solved optimally.
  * `warmstart` (default: `false`): whether the solver should start the
    optimization from a previous optimal value (according to the current primal
    value of the variables in the problem, which can be set by [`set_value!`](@ref).
 """
 function solve!(
     p::Problem,
     optimizer_factory;
-    silent_solver = false,
+    silent = nothing, # to become `false` once `silent_solver` deprecation removed
```
odow, line 82:
```suggestion
    silent::Union{Nothing,Bool} = nothing, # to become `false` once `silent_solver` deprecation removed
```

test/test_utilities.jl
```diff
@@ -1022,6 +1022,23 @@ function test_deprecation_adding_constraints()
...
+    @test_logs (:warn, str) solve!(p, SCS.Optimizer; silent_solver = true)
+
+    # Can't set both
+    @test_throws ArgumentError solve!(
+        p,
+        SCS.Optimizer;
+        silent_solver = true,
+        silent = false,
+    )
+
```
odow, line 1038:
```suggestion
```

test/test_utilities.jl
```diff
@@ -1022,6 +1022,23 @@ function test_deprecation_adding_constraints()
     return
 end
 
+function test_deprecation_silent_solver()
+    x = Variable(2)
+    p = minimize(sum(x))
+    str = "The keyword argument `silent_solver` in `Convex.solve!` has been deprecated in favor of `silent`."
+    @test_logs (:warn, str) solve!(p, SCS.Optimizer; silent_solver = true)
+
```
odow, line 1030:
```suggestion
```

src/solution.jl
```diff
@@ -69,27 +69,45 @@ duals (accessed by `cons.dual`), where applicable. Returns the input `problem`.
...
+    elseif silent_solver !== nothing
+        @warn(
+            "The keyword argument `silent_solver` in `Convex.solve!` has been deprecated in favor of `silent`.",
+            maxlog = 1,
+        )
+        silent = silent_solver
+    elseif silent === nothing
+        silent = false
+    end
+
```
odow, line 101:
```suggestion
```

PR #671, opened 2024-05-18
Add quantum conditional entropy example

ericphanson:
this is the second example from https://github.com/jump-dev/Convex.jl/pull/418#issuecomment-832190560

I thought it could be nice to have some examples using some of the CVXQUAD functionality. I put in a lot of quantum information theory preliminaries, but maybe we could trim some back if it's too much.

This example originally had 23s of formulation time and 4s solve time. I profiled and found `permutedims_matrix` was very slow and allocating (and it gets used a lot, since it is used for `transpose` and thus for `vcat`). I was able to optimize it by constructing the sparse matrix directly (in COO format). Now we have 3s formulation time and the same 4s solve time, which seems much more acceptable (maybe a little slow still, but I didn't see any other low hanging fruit).

BTW, here the reformulations and bridges really blow up the problem size:
```julia
julia> problem
Problem statistics
  problem is DCP         : true
  number of variables    : 1 (32 scalar elements)
  number of constraints  : 3 (67 scalar elements)
  number of coefficients : 31
  number of atoms        : 29
  ...

julia> get_info(problem.model)
(n_variables = 1639, n_constraints = 16, n_elements = 18259, n_constants = 43103)
```

using
```julia
julia> function get_info(model)
           n_constraints = 0
           n_elements = 0
           n_variables = MOI.get(model, MOI.NumberOfVariables())
           n_constants = 0
           for (F, S) in MOI.get(model, MOI.ListOfConstraintTypesPresent())
               n_constraints += MOI.get(problem.model, MOI.NumberOfConstraints{F,S}())

               for inds in MOI.get(model, MOI.ListOfConstraintIndices{F,S}())
                   set = MOI.get(model, MOI.ConstraintSet(), inds)
                   n_elements += MOI.dimension(set)
                   f = MOI.get(model, MOI.ConstraintFunction(), inds)
                   n_constants += count_size(f)
               end
           end
           return (; n_variables, n_constraints, n_elements, n_constants)
       end
get_info (generic function with 1 method)

julia> function count_size(f::MOI.VectorAffineFunction)
           return length(f.terms) + length(f.constants)
       end
count_size (generic function with 1 method)
```

I think Hypatia can be used to avoid some of the reformulations; I guess to hook everything up, we would need to move the sets to MOI and update Hypatia to support them directly (?).

### DIFF
--- a/docs/src/examples/optimization_with_complex_variables/quantum_conditional_entropy.jl
+++ b/docs/src/examples/optimization_with_complex_variables/quantum_conditional_entropy.jl
@@ -76,19 +76,22 @@ add_constraint!(ρ_AB, tr(ρ_AB) == 1)
 
 add_constraint!(ρ_AB, 0.5 * nuclearnorm(ρ_AB - σ_AB) ≤ ϵ)
 problem = maximize(quantum_conditional_entropy(ρ_AB, d_A, d_B))
-solve!(problem, SCS.Optimizer; silent_solver=true)
+solve!(problem, SCS.Optimizer; silent_solver = false)
 
 # We can then check the observed difference in relative entropies:
 
-difference = evaluate(quantum_conditional_entropy(ρ_AB, d_A, d_B) - quantum_conditional_entropy(σ_AB, d_A, d_B))
+difference = evaluate(
+    quantum_conditional_entropy(ρ_AB, d_A, d_B) -
+    quantum_conditional_entropy(σ_AB, d_A, d_B),
+)
 
 # We can compare to the bound:
-h(x) = -x*log(x)  - (1-x)*log(1-x)
-bound = 2 * ϵ *  log(d_A) + (1 + ϵ) * h(ϵ/(1+ϵ))
+h(x) = -x * log(x) - (1 - x) * log(1 - x)
+bound = 2 * ϵ * log(d_A) + (1 + ϵ) * h(ϵ / (1 + ϵ))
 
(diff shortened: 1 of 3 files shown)


ericphanson:
https://jump.dev/Convex.jl/previews/PR671/examples/optimization_with_complex_variables/quantum_conditional_entropy/

docs/src/examples/optimization_with_complex_variables/quantum_conditional_entropy.jl
```diff
@@ -0,0 +1,103 @@
...
+# We will choose $d_A=d_B=2$, and $\sigma$ as the maximally entangled state:
+#
+# ```math
+# \sigma = \frac{1}{2}\begin{pmatrix}1 & 0 & 0 & 1\\ 0 & 0 & 0 & 0 \\ 0 & 0 & 0 & 0 \\ 1 & 0 & 0 & 1\\\end{pmatrix}
+# ```math
+#
+# First, we can formulate the conditional entropy in terms of the relative entropy using the relationship
+#
+# ```math
+# S(A|B)_\rho = - D(\rho_^{AB} \| I_A \otimes \rho^B)
```
ericphanson, line 46:
```suggestion
# S(A|B)_\rho = - D(\rho^{AB} \| I_A \otimes \rho^B)
```

odow:
> I guess to hook everything up, we would need to move the sets to MOI and update Hypatia to support them directly

Yes, I think @blegat has a plan for this.

PR #675, opened 2024-05-19
Add "performance tips" page to docs

ericphanson:
I figure this could be helpful, since Convex is fairly different from both JuMP and regular Julia in that it has trouble with scalar indexing, and since conic solvers in general can be a bit unintuitive to folks.


docs/src/manual/performance_tips.md
```diff
@@ -0,0 +1,47 @@
...
+* If formulating the problem is much slower than solving it, there may be a performance issue in Convex.jl. See [Faster problem formulation: Avoid scalar indexing](@ref) below for one tip to speed things up.
+* If solving the problem is the slowest part, you may still be able to improve performance. See [Speeding up solve times: dualization](@ref) and [Speeding up solve times: conic-friendly formulations](@ref) below. Also consider trying other solvers. For large problems, first-order solvers like SCS and COSMO may be more performant (albeit with higher convergence tolerances) than second-order solvers like Clarabel and Hypatia.
+
+If you're struggling to improve performance, feel free to ask for help in the [community forum](https://jump.dev/forum). Before asking a question, make sure to read the post [make it easier to help you](https://discourse.julialang.org/t/psa-make-it-easier-to-help-you/14757), which contains a number of tips on how to ask a good question.
```
ericphanson, line 15: this is duplicated from the main page, but I think it could be handy to have it called out here also

docs/src/manual/performance_tips.md
```diff
@@ -0,0 +1,47 @@
...
+
+For example, it is typically easier for conic solvers to minimize $\|A*x\|$ than $\|A*x\|^2$, since the former is reformulated as a second order cone constraint, and the latter as the same second order cone constraint, plus a rotated second order cone constraint (to model the final squaring). This may be unintuitive, since for non-conic solvers, often $\|A*x\|^2$ is easier to solve (as it avoids a square-root in the typical formulation).
+
+Convex.jl provides an abstraction by automatically performing these reformulations, but as often is the case, performance leaks through. To have the best performance, you may will likely need to learn more details about the various cones supported by your solver, by [MathOptInterface](https://jump.dev/MathOptInterface.jl/stable/reference/standard_form/#Vector-sets), and how Convex.jl reformulates your problem.
```
ericphanson, line 47:
```suggestion
Convex.jl provides an abstraction by automatically performing these reformulations, but as often is the case, performance leaks through. To have the best performance, you will likely need to learn more details about the various cones supported by your solver, by [MathOptInterface](https://jump.dev/MathOptInterface.jl/stable/reference/standard_form/#Vector-sets), and how Convex.jl reformulates your problem.
```

docs/src/manual/performance_tips.md
```diff
@@ -0,0 +1,47 @@
+# Performance tips
+
+There are 3 phases of building and solving a Convex.jl problem:
+
+1. Building the expression tree. This is everything that happens before `solve!` is called, e.g. creating variables, problems, constraints, and other expressions.
```
odow, line 5:
```suggestion
1. Building the expression tree. This is everything that happens before `solve!` is called, for example, creating variables, problems, constraints, and other expressions.
```

docs/src/manual/performance_tips.md
```diff
@@ -0,0 +1,47 @@
...
+2. Formulating the problem in MathOptInterface. This happens automatically when `solve!` is called, and (as long as `solve!` is called with `silent=false`) emits a log message when completed, containing the amount of time it took and the total amount of memory allocations.
```
odow, line 6:
```suggestion
2. Formulating the problem in MathOptInterface. This happens automatically when `solve!` is called, and (as long as `solve!` is not called with `silent=true`) emits a log message when completed, containing the amount of time it took and the total amount of memory allocations.
```

docs/src/manual/performance_tips.md
```diff
@@ -0,0 +1,47 @@
...
+
+If you are experiencing performance issues, it is important to isolate which of these three steps is taking the most time. Typically, step (1) should be the fastest, then step (2), then step (3).
+
+* If building the expression tree (step 1 above) is the slowest part, there is likely a performance bug in Convex.jl. Please [file an issue](https://github.com/jump-dev/Convex.jl/issues/new) with a reproducible minimal example; it may be able to be quickly fixed.
+* If formulating the problem is much slower than solving it, there may be a performance issue in Convex.jl. See [Faster problem formulation: Avoid scalar indexing](@ref) below for one tip to speed things up.
+* If solving the problem is the slowest part, you may still be able to improve performance. See [Speeding up solve times: dualization](@ref) and [Speeding up solve times: conic-friendly formulations](@ref) below. Also consider trying other solvers. For large problems, first-order solvers like SCS and COSMO may be more performant (albeit with higher convergence tolerances) than second-order solvers like Clarabel and Hypatia.
```
odow, line 13:
```suggestion
* If solving the problem is the slowest part, you may still be able to improve performance. See [Speeding up solve times: dualization](@ref) and [Speeding up solve times: conic-friendly formulations](@ref) below. Also consider trying other solvers. For large problems, first-order solvers like SCS and COSMO may be faster (albeit with higher convergence tolerances) than second-order solvers like Clarabel and Hypatia.
```

docs/src/manual/performance_tips.md
```diff
@@ -0,0 +1,47 @@
...
+
+If you're struggling to improve performance, feel free to ask for help in the [community forum](https://jump.dev/forum). Before asking a question, make sure to read the post [make it easier to help you](https://discourse.julialang.org/t/psa-make-it-easier-to-help-you/14757), which contains a number of tips on how to ask a good question.
+
+## Faster problem formulation: Avoid scalar indexing
+
+When manipulating Convex.jl expressions, try to use a "vectorized" style, and avoid loops and scalar indexing. For example,
```
odow, line 19:
```suggestion
When manipulating Convex.jl expressions, try to use a "vectorized" style, and avoid loops and scalar indexing. For example:
```

docs/src/manual/performance_tips.md
```diff
@@ -0,0 +1,47 @@
...
+
+```julia
+function my_dot(x, y) # avoid this kind of code for Convex.jl expressions!
```
odow, line 22:
```suggestion
function bad_my_dot(x, y) # avoid this kind of code for Convex.jl expressions!
```

docs/src/manual/performance_tips.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Speeding up solve times: dualization
+
+Depending on the solver and the problem, it can sometimes speed things up quite a lot to pass the solver the dual problem to solve rather than the primal problem. This can be accomplished easily with [Dualization.jl](https://github.com/jump-dev/Dualization.jl), simply by passing `dual_optimizer(optimizer)` instead of `optimizer` to `solve!`. See [Dualization](@ref) for an example.
+
+## Speeding up solve times: conic-friendly formulations
+
+Convex.jl is built around conic programming, and works with conic solvers. This is a different optimization methodology than you may be used to, and works by formulating problems in terms of affine objective functions and affine-function-in-cone constraints for a variety of convex cones. Convex.jl reformulates all problems to this form. Specifically, the e.g. objective function you give Convex will not actually be executed line-by-line; instead, it will be reformulated to a possibly very-different looking form for the conic solver to handle. This has important consequences for performance.
```
odow, line 43:
```suggestion
Convex.jl is built around conic programming, and it works with conic solvers. This may be a different optimization methodology than you may be used to, and it works by formulating problems in terms of affine objective functions and affine-function-in-cone constraints for a variety of convex cones. Convex.jl reformulates all problems to this form. For example, the objective function you give Convex will not actually be executed line-by-line; instead, it will be reformulated to a possibly very-different looking form for the conic solver to handle. This has important consequences for performance.
```

docs/src/manual/performance_tips.md
```diff
@@ -0,0 +1,47 @@
+# Performance tips
+
+There are 3 phases of building and solving a Convex.jl problem:
```
odow, line 3:
```suggestion
There are three phases of building and solving a Convex.jl problem:
```

docs/src/manual/performance_tips.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Speeding up solve times: conic-friendly formulations
+
+Convex.jl is built around conic programming, and it works with conic solvers. This may be a different optimization methodology than you may be used to, and it works by formulating problems in terms of affine objective functions and affine-function-in-cone constraints for a variety of convex cones. Convex.jl reformulates all problems to this form. For example, the objective function you give Convex will not actually be executed line-by-line; instead, it will be reformulated to a possibly very-different looking form for the conic solver to handle. This has important consequences for performance.
+
+For example, it is typically easier for conic solvers to minimize $\|A*x\|$ than $\|A*x\|^2$, since the former is reformulated as a second order cone constraint, and the latter as the same second order cone constraint, plus a rotated second order cone constraint (to model the final squaring). This may be unintuitive, since for non-conic solvers, often $\|A*x\|^2$ is easier to solve (as it avoids a square-root in the typical formulation).
```
odow, line 45: This is not really true, since the latter can be a rotated second order cone constraint.

ericphanson: Right now we make two constraints like l wrote. Maybe there’s a better example though?

ericphanson: should I just delete this section? I feel like there's a pedagogical reason to discuss this stuff but I'm not 100% sure how actionable it is for "performance tips", since Convex basically makes you formulate conic-friendly things and I haven't seen this come up that often in practice. I guess the actionable thing is: "the problem you write down is not the problem being solved, so you may need to dig in and break the abstraction if you want to top performance"

ericphanson: (filed https://github.com/jump-dev/Convex.jl/issues/677 btw)

ericphanson: I rewrote this section to be about "look at the final problem formulation". I believe the only stable API we have for this is `write_to_file`, so I used that.

BTW, it would probably be more standard to put the filename first (like in `Base.write`), but probably it's not worth changing from the MOI convention here.

docs/src/manual/performance_tips.md
```diff
@@ -0,0 +1,47 @@
...
+
+Convex.jl is built around conic programming, and it works with conic solvers. This may be a different optimization methodology than you may be used to, and it works by formulating problems in terms of affine objective functions and affine-function-in-cone constraints for a variety of convex cones. Convex.jl reformulates all problems to this form. For example, the objective function you give Convex will not actually be executed line-by-line; instead, it will be reformulated to a possibly very-different looking form for the conic solver to handle.
+
+Convex.jl provides an abstraction by automatically performing these reformulations, but as often is the case, performance leaks through. To have the best performance, you might need to learn more details about the various cones supported by your solver, by [MathOptInterface](https://jump.dev/MathOptInterface.jl/stable/reference/standard_form/#Vector-sets), and how Convex.jl reformulates your problem. There could be alternate mathematically-equivalent formulations that are more performant, and in some cases it could be that Convex's automatic reformulations can be improved.
```
ericphanson, line 45:
```suggestion
Convex.jl provides an abstraction by automatically performing these reformulations, but as often is the case, performance leaks through. To have the best performance, you might need to learn more details about the various cones supported by your solver, by [MathOptInterface](https://jump.dev/MathOptInterface.jl/stable/reference/standard_form/#Vector-sets), and how Convex.jl reformulates your problem. There could be mathematically equivalent formulations that are faster, and in some cases it could be that Convex's automatic reformulations can be improved.
```

PR #678, opened 2024-05-20
Fix formulation of Convex.sumsquares

### LINKED ISSUE #677: `sumsquares` makes both a SOC and RSOC constraint
I believe this could be a single RSOC constraint instead, but we would need to add an atom for it. I'm not sure how much this matters in practice.

```julia
julia> using Convex

julia> using Convex: MOI

julia> problem = minimize(sumsquares(Variable(2)));

julia> context = Convex.Context(problem, MOI.Utilities.Model{Float64});

julia> print(context.model)
Minimize ScalarAffineFunction{Float64}:
 0.0 + 1.0 v[1]

Subject to:

VectorAffineFunction{Float64}-in-SecondOrderCone
 ┌              ┐
 │0.0 + 1.0 v[2]│
 │0.0 + 1.0 v[3]│
 │0.0 + 1.0 v[4]│
 └              ┘ ∈ SecondOrderCone(3)

VectorAffineFunction{Float64}-in-RotatedSecondOrderCone
 ┌              ┐
 │0.0 + 1.0 v[1]│
 │0.5           │
 │0.0 + 1.0 v[2]│
 └              ┘ ∈ RotatedSecondOrderCone(3)
```

(This is since `sumsquares` is implemented by squaring `norm`, so we first create the SOC for the norm, then the RSOC for the result of norm.)

odow:
Closes #677

### DIFF
--- a/docs/src/manual/complex-domain_optimization.md
+++ b/docs/src/manual/complex-domain_optimization.md
@@ -56,7 +56,6 @@ operate on complex variables as well. Notable exceptions include:
 
 > -   `inverse`
 > -   `square`
-> -   `quadoverlin`
 > -   `sqrt`
 > -   `geomean`
 > -   `huber`
--- a/src/atoms/QolElemAtom.jl
+++ b/src/atoms/QolElemAtom.jl
@@ -57,8 +57,6 @@ function square(x::AbstractExpr)
     return QolElemAtom(x, constant(ones(x.size)))
 end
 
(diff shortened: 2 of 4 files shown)


ericphanson:
https://github.com/jump-dev/Convex.jl/pull/679 is an alternative to 639cc285bae3da3830b0438c9ac23512cec87721 fyi

PR #680, opened 2024-05-20
remove unused incorrect method

ericphanson:
noticed this when I was looking into #679


odow (approved):

ericphanson:
this doesn't need a news entry FYI bc it was introduced in #504 (https://github.com/jump-dev/Convex.jl/pull/504/files#diff-8e3527da4f9bdc7b91a3918096a547c5594656c783b388d80b99e477b8bb3023R26)

PR #684, opened 2024-05-21
`lieb_ando`: use partially specified problem to get correct `vexity`

### LINKED ISSUE #683: `lieb_ando` reformulation doesn't propogate DCP information correctly
```julia
julia> using Convex

julia> A = Variable(2,2)
Variable
size: (2, 2)
sign: real
vexity: affine
id: 406…508

julia> B = Variable(2,2)
Variable
size: (2, 2)
sign: real
vexity: affine
id: 169…045

julia> K = rand(2,2)
2×2 Matrix{Float64}:
 0.388952  0.227043
 0.64608   0.220724

julia> lieb_ando(A, B, K, 1//1)
real (affine; real)
└─ sum (affine; real)
   └─ diag (affine; real)
      └─ * (affine; real)
         ├─ …
         └─ …

julia> lieb_ando(A, B, K, -1//2)
real (affine; real)
└─ sum (affine; real)
   └─ diag (affine; real)
      └─ * (affine; real)
         ├─ …
         └─ …

help?> lieb_ando
search: lieb_ando

  Returns LinearAlgebra.tr(K' * A^{1-t} * K * B^t) where A and B are positive
  semidefinite matrices and K is an arbitrary matrix (possibly rectangular).

  Disciplined convex programming information: lieb_ando(A,B,K,t) is concave in
  (A,B) for t in [0,1], and convex in (A,B) for t in [-1,0] or [1,2]. K is a
  fixed matrix.
```

Seems like these should be concave/convex, not affine. The issue is similar to https://github.com/jump-dev/Convex.jl/issues/682#issuecomment-2122446577: extended formulations typically must be atoms to propagate DCP information correctly. Here, we add constraints to a new variable we create [here](https://github.com/jump-dev/Convex.jl/blob/2978c32c0fcc3abc1cf10704980367cc23f8e1df/src/reformulations/lieb_ando.jl#L91-L98) and [here](https://github.com/jump-dev/Convex.jl/blob/2978c32c0fcc3abc1cf10704980367cc23f8e1df/src/reformulations/lieb_ando.jl#L101-L107).

I didn't see any other extended formulations in `src/reformulations` luckily; the other reformulations there simply compose operations without adding new variables or constraints.

We should also update the dev docs to mention this. I for one didn't fully realize this issue until today.

ericphanson:
This is in fact the approach used in CVXQUAD (https://github.com/hfawzi/cvxquad/blob/master/lieb_ando.m), but we didn't have those yet at the time.

This way we don't have to create an atom! Pretty convenient.

Also: I noticed in many of the CVXQUAD-derived atoms, we expect some variables to be PSD, but don't check or enforce it. Could be something we might want to do.

closes #683

### DIFF
--- a/docs/src/developer/contributing.md
+++ b/docs/src/developer/contributing.md
@@ -46,10 +46,10 @@ Convex.jl. Let's say you're adding the new function $f$.
      since then any changes to the variable's value (or it being `free!`'d) will not be recognized.
      See [#653](https://github.com/jump-dev/Convex.jl/issues/653) and [#585](https://github.com/jump-dev/Convex.jl/issues/585)
      for previous bugs caused by misuse here.
- -  Do not add constraints to variables directly during a reformulation.
-    Instead, create an atom to control the vexity, or return a partially
-    specified problem: `minimize(var, constraints...)`, to ensure the `vexity`
-    of the result is correct.
+ -   Do not add constraints to variables directly during a reformulation.
+     Instead, create an atom to control the vexity, or return a partially
+     specified problem: `minimize(var, constraints...)`, to ensure the `vexity`
+     of the result is correct.
  -   The most mathematically interesting part is the `new_conic_form!`
      function. Following the example in the nuclear norm atom, you'll
      see that you can just construct the problem whose optimal value is
--- a/docs/src/developer/contributing.md
+++ b/docs/src/developer/contributing.md
@@ -46,7 +46,10 @@ Convex.jl. Let's say you're adding the new function $f$.
      since then any changes to the variable's value (or it being `free!`'d) will not be recognized.
      See [#653](https://github.com/jump-dev/Convex.jl/issues/653) and [#585](https://github.com/jump-dev/Convex.jl/issues/585)
      for previous bugs caused by misuse here.
- -  Do not add constraints to variables directly during a reformulation. Instead, create an atom to control the vexity, or return a partially-specified problem: `minimize(var, constraints...)`, to ensure the `vexity` of the result is correct.
+ -  Do not add constraints to variables directly during a reformulation.
+    Instead, create an atom to control the vexity, or return a partially
+    specified problem: `minimize(var, constraints...)`, to ensure the `vexity`
+    of the result is correct.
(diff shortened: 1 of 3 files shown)


ericphanson:
unexpected/unrelated docs failure:
```julia
ERROR: LoadError: LoadError: ArgumentError: Package TimeZones does not have RecipesBase in its dependencies:
- If you have TimeZones checked out for development and have
  added RecipesBase as a dependency but haven't updated your primary
  environment's manifest file, try `Pkg.resolve()`.
- Otherwise you may need to report an issue with TimeZones
```

odow:
I just restarted it. The TimeZones release is only 2 hours old: https://github.com/JuliaTime/TimeZones.jl/releases/tag/v1.16.0, so I wonder if it's an issue with the cache

odow (approved):

odow:
Ha. I just arrived at https://github.com/JuliaTime/TimeZones.jl/pull/459#issuecomment-2123418212 to find you there too.

As a work-around, we can update the docs to build on Julia 1.10

ericphanson:
yeah, that works. I put up a patch also: https://github.com/JuliaTime/TimeZones.jl/pull/461

docs/src/developer/contributing.md
```diff
@@ -46,6 +46,7 @@ Convex.jl. Let's say you're adding the new function $f$.
      since then any changes to the variable's value (or it being `free!`'d) will not be recognized.
      See [#653](https://github.com/jump-dev/Convex.jl/issues/653) and [#585](https://github.com/jump-dev/Convex.jl/issues/585)
      for previous bugs caused by misuse here.
+ -  Do not add constraints to variables directly during a reformulation. Instead, create an atom to control the vexity, or return a partially-specified problem: `minimize(var, constraints...)`, to ensure the `vexity` of the result is correct.
```
odow, line 49:
```suggestion
 -  Do not add constraints to variables directly during a reformulation.
    Instead, create an atom to control the vexity, or return a partially
    specified problem: `minimize(var, constraints...)`, to ensure the `vexity`
    of the result is correct.
```

PR #685, opened 2024-05-21
change `LogSumExp` atom to be per-column, for efficient `logisticloss`

### LINKED ISSUE #682: Scalar indexing in `logisticloss`
https://github.com/jump-dev/Convex.jl/blob/2978c32c0fcc3abc1cf10704980367cc23f8e1df/src/atoms/LogSumExpAtom.jl#L48

I guess I missed this one in #614

ericphanson:
closes #682 

the actual atom was never documented, only `logsumexp`, so this should be non-breaking

### DIFF
--- a/src/atoms/LogSumExpAtom.jl
+++ b/src/atoms/LogSumExpAtom.jl
@@ -3,40 +3,50 @@
 # Use of this source code is governed by a BSD-style license that can be found
 # in the LICENSE file or at https://opensource.org/license/bsd-2-clause
 
-mutable struct LogSumExpAtom <: AbstractExpr
+# Returns:
+# [logsumexp(v) for v in eachcol(x)]
+# where `logsumexp(v) = log(sum(exp, v))` except hopefully more numerically stable
+mutable struct ColwiseLogSumExpAtom <: AbstractExpr
     children::Tuple{AbstractExpr}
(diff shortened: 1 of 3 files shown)


odow:
Since this is a non-breaking performance issue, I think I'd prefer to come up with a standard way to represent `dims` arguments. So `dims = 0` is element-wise, `dims = 1` makes a row vector, and `dims = 2` makes a col vector.

We have quite a few hodgepodge elementwise/row/column conventions. It'd be useful to standardize.

ericphanson:
I wonder if there's a lil abstraction we use, like
```julia
eachcol(x) = EachColAtom(x)
eachrow(x) = EachRowAtom(x)
```
and then these have slow IndexAtom based fallbacks, but we can sneak in optimizations like `logsumexp.(::EachColAtom)` lowering to something nice. (Either literally using broadcasting and hooking into the machinery there, or using `map` and regular dispatch etc).

odow:
Yeah, I don't have a solution off the top of my head, but something like that. 

This is my main bugbear:
https://github.com/jump-dev/Convex.jl/blob/bce84a9620b093fc95a4c0daf9262d77f0f73aab/src/atoms/EntropyAtom.jl#L34-L36

ericphanson:
closing in favor of #692

PR #688, opened 2024-05-23
Use two constraints in example problem

ericphanson:
this was pointed out as unnecessarily complicated by @matbesancon

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ x = Variable(n)
 
 # The problem is to minimize ||Ax - b||^2 subject to x >= 0
 # This can be done by: minimize(objective, constraints)
-problem = minimize(sumsquares(A * x - b), [x >= 0])
+problem = minimize(sumsquares(A * x - b), x >= 0)
 
 # Solve the problem by calling solve!
 solve!(problem, SCS.Optimizer)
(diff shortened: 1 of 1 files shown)


blegat (approved):

README.md
```diff
@@ -49,7 +49,7 @@ x = Variable(n)
 
 # The problem is to minimize ||Ax - b||^2 subject to x >= 0
 # This can be done by: minimize(objective, constraints)
-problem = minimize(sumsquares(A * x - b), [x >= 0])
+problem = minimize(sumsquares(A * x - b), x >= 0)
```
odow, line 52: I actually prefer the `[]`. It makes it more obvious that you can pass a vector of constraints.

ericphanson: I wonder if this one is good for the simplest possible example, and then we could do one more line that has two constraints that uses the `[]`? e.g.

```julia
# Note collections of constraints can also be passed as a vector:
problem = minimize(sumsquares(A * x - b), [x >= 0, x <= 1])
```
just after the last line in this example.

blegat: Yes, you probably should have both in the README, otherwise you will have many questions about how to create the vectorized version

odow: I don't think we should encourage the varargs version. Let's just have the single `[x >= 0, x <= 1]` version. It follows that you can have different number of constraints in the vector.

ericphanson: ok, done

PR #689, opened 2024-05-24
Update CI badge

ericphanson:
I believe the current badge shows failing if the latest CI run was on a PR that was failing (maybe? I believe I saw this a few days ago at least). This one I got from github by going to https://github.com/jump-dev/Convex.jl/actions/workflows/ci.yml, clicking the 3 dots, and "create status badge": 
Screenshot 2024-05-24 at 14 35 46

This changes the name to `CI`, but I kinda think that is better than "build status".

xref https://github.com/actions/starter-workflows/issues/1525#issuecomment-1763431305


odow:
JuMP uses `#branch=master`, but this doesn't work for Convex because we don't run CI on each commit to master...

PR #692, opened 2024-05-27
Add dims argument to LogSumExpAtom

### LINKED ISSUE #682: Scalar indexing in `logisticloss`

odow:
Closes #682 
Closes #685

### DIFF
--- a/src/atoms/LogSumExpAtom.jl
+++ b/src/atoms/LogSumExpAtom.jl
@@ -4,7 +4,7 @@
 # in the LICENSE file or at https://opensource.org/license/bsd-2-clause
 
 """
-    LogSumExpAtom(x::AbstractExpr, dims::Int = 0)
+    LogSumExpAtom(x::AbstractExpr, dims::Union{Colon,Int} = :)
 
 Represents the expression `log.(sum(exp.(x); dims))`.
 """
--- a/src/atoms/LogSumExpAtom.jl
+++ b/src/atoms/LogSumExpAtom.jl
@@ -3,17 +3,26 @@
 # Use of this source code is governed by a BSD-style license that can be found
 # in the LICENSE file or at https://opensource.org/license/bsd-2-clause
 
+"""
+    LogSumExpAtom(x::AbstractExpr, dims::Int = 0)
+
+Represents the expression `log.(sum(exp.(x); dims))`.
+"""
 mutable struct LogSumExpAtom <: AbstractExpr
     children::Tuple{AbstractExpr}
     size::Tuple{Int,Int}
+    dims::Union{Colon,Int}
 
-    function LogSumExpAtom(x::AbstractExpr)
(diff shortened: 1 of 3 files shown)


odow:
So perhaps the simplest thing is just if we add a `; dims` argument to relevant of atoms.

ericphanson (approved):
Lgtm, nice and simple. I think it would be worth carrying over the tests in src/problem_depot/problems/exp.jl from the other PR

src/atoms/LogSumExpAtom.jl
```diff
@@ -28,22 +37,29 @@ curvature(::LogSumExpAtom) = ConvexVexity()
 function evaluate(x::LogSumExpAtom)
     _x = evaluate(x.children[1])
     max_x = maximum(_x)
-    return max_x + log(sum(exp.(_x .- max_x)))
+    return max_x .+ log.(sum(exp.(_x .- max_x); x.dims))
```
ericphanson, line 40:
```suggestion
    return output(max_x .+ log.(sum(exp.(_x .- max_x); x.dims)))
```
We may want this so we don’t return vectors for scalars

odow: I think this just works. I'd already added a test for the scalar case:
```julia
@test evaluate(logsumexp(x; dims = :)) ≈ 6.456193316018123
```

src/problem_depot/problems/exp.jl
```diff
@@ -121,14 +121,55 @@ end
...
     p = minimize(logsumexp(y), y >= 1; numeric_type = T)
-
     if test
         @test problem_vexity(p) == ConvexVexity()
     end
     handle_problem!(p)
     if test
         @test p.optval ≈ log(exp(1) * 5) atol = atol rtol = rtol
     end
+
```
odow, line 131:
```suggestion

```

src/atoms/LogSumExpAtom.jl
```diff
@@ -3,17 +3,26 @@
 # Use of this source code is governed by a BSD-style license that can be found
 # in the LICENSE file or at https://opensource.org/license/bsd-2-clause
 
+"""
+    LogSumExpAtom(x::AbstractExpr, dims::Int = 0)
```
ericphanson, line 7: should this be `:`? also do we want to canonicalize 0 to `:` or something?

odow: Oops yeah. I originally did the `0` think that canonicalized to `:`, then I just made it `Union{Int,Colon}`. Did't update the docstring

odow, line 7:
```suggestion
    LogSumExpAtom(x::AbstractExpr, dims::Union{Colon,Int} = :)
```

ericphanson (approved):

PR #695, opened 2024-06-17
Support AbstractArray in relative_entropy and log_perspective

### LINKED ISSUE #694: `relative_entropy` doesn't support constants which aren't wrapped in `constant`
```julia
julia> using Convex

julia> N = 10
10

julia> b = rand(N)
10-element Vector{Float64}:
 0.13262633847871297
 0.9254408726815795
 0.13936575019827857
 0.3329759851512133
 0.5964209280418487
 0.8590087709390417
 0.7599860068305675
 0.841494184551078
 0.1349705410637151
 0.017608754157530382

julia> x = Variable(10)
Variable
size: (10, 1)
sign: real
vexity: affine
id: 493…512

julia> relative_entropy(b, x)
ERROR: MethodError: no method matching relative_entropy(::Vector{Float64}, ::Variable)

Closest candidates are:
  relative_entropy(::Convex.AbstractExpr, ::Convex.AbstractExpr)
   @ Convex ~/.julia/packages/Convex/y7lu0/src/atoms/RelativeEntropyAtom.jl:53

Stacktrace:
 [1] top-level scope
   @ REPL[72]:1

julia> relative_entropy(constant(b), x)
relative_entropy (convex; real)
├─ 10×1 Matrix{Float64}
└─ 10-element real variable (id: 493…512)
```

We should add some methods to add `constant` automatically.

odow:
Closes #694

### DIFF
--- a/src/supported_operations.jl
+++ b/src/supported_operations.jl
@@ -1056,7 +1056,12 @@ julia> size(atom)
 (1, 1)
 ```
 """
-log_perspective(x::AbstractExpr, y::AbstractExpr) = -relative_entropy(y, x)
+function log_perspective(
+    x::Union{AbstractExpr,AbstractArray},
+    y::Union{AbstractExpr,AbstractArray},
+)
+    return -relative_entropy(y, x)
+end
 
 """
     LinearAlgebra.logdet(X::Convex.AbstractExpr)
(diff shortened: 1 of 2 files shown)


src/supported_operations.jl
```diff
@@ -1857,6 +1861,16 @@ julia> size(atom)
...
 
+function relative_entropy(x::AbstractExpr, y::AbstractArray)
+    return RelativeEntropyAtom(x, constant(y))
+end
+
+function relative_entropy(x::AbstractArray, y::AbstractExpr)
+    return RelativeEntropyAtom(constant(x), y)
+end
+
+relative_entropy(x::AbstractArray, y::AbstractArray) = sum(y .* log(x ./ y))
```
odow, line 1872:
```suggestion
relative_entropy(x::AbstractArray, y::AbstractArray) = sum(y .* log.(x ./ y))
```

PR #699, opened 2024-06-20
Fix distance_to_set for sets without a definition in MOI

### LINKED ISSUE #698: MethodError in distance_to_set
```Julia
julia> using Convex, HiGHS

julia> x = Variable(2, 2)
Variable
size: (2, 2)
sign: real
vexity: affine
id: 909…059

julia> fix!(x, [1 0; 0 1])
Variable
size: (2, 2)
sign: real
vexity: constant
id: 909…059
value: [1 0; 0 1]

julia> problem = satisfy([sum(x; dims = 1) <= 1])
Problem statistics
  problem is DCP         : true
  number of variables    : 1 (4 scalar elements)
  number of constraints  : 1 (2 scalar elements)
  number of coefficients : 5
  number of atoms        : 4

Solution summary
  termination status : OPTIMIZE_NOT_CALLED
  primal status      : NO_SOLUTION
  dual status        : NO_SOLUTION

Expression graph
  satisfy
   └─ nothing
  subject to
   └─ ≤ constraint (constant)
      └─ + (constant; real)
         ├─ * (constant; real)
         │  ├─ …
         │  └─ …
         └─ Convex.NegateAtom (constant; negative)
            └─ …


julia> solve!(problem, HiGHS.Optimizer)
ERROR: distance_to_set using the distance metric MathOptInterface.Utilities.ProjectionUpperBoundDistance() for set type MathOptInterface.Nonpositives has not been implemented yet.
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] distance_to_set(d::MathOptInterface.Utilities.ProjectionUpperBoundDistance, ::Matrix{Float64}, set::MathOptInterface.Nonpositives)
    @ MathOptInterface.Utilities ~/.julia/packages/MathOptInterface/gBojA/src/Utilities/distance_to_set.jl:75
```

odow:
Closes #698

### DIFF
--- a/src/Constraint.jl
+++ b/src/Constraint.jl
@@ -46,6 +46,9 @@ is_feasible(x, set, tol) = missing
 function is_feasible(
     x::Vector,
     set::Union{
+        MOI.Nonnegatives,
+        MOI.Nonpositives,
+        MOI.Zeros,
         MOI.SecondOrderCone,
         MOI.RotatedSecondOrderCone,
         MOI.ExponentialCone,
--- a/src/Constraint.jl
+++ b/src/Constraint.jl
@@ -44,7 +44,7 @@ AbstractTrees.children(c::Constraint) = (c.child,)
 # defined.
 function is_feasible(x, set, tol)
     try
-      return MOI.Utilities.distance_to_set(x, set) <= tol
+        return MOI.Utilities.distance_to_set(x, set) <= tol
     catch
         return true  # default to `true`
     end
--- a/src/Constraint.jl
+++ b/src/Constraint.jl
@@ -44,7 +44,7 @@ AbstractTrees.children(c::Constraint) = (c.child,)
 is_feasible(x, set, tol) = missing
 
 function is_feasible(
-    x,
+    x::Vector,
     set::Union{
         MOI.SecondOrderCone,
         MOI.RotatedSecondOrderCone,
--- a/src/Constraint.jl
+++ b/src/Constraint.jl
@@ -40,14 +40,26 @@ end
 
 AbstractTrees.children(c::Constraint) = (c.child,)
 
-# A fallback. Define a new method if `MOI.Utilities.distance_to_set` is not
-# defined.
-function is_feasible(x, set, tol)
-    try
-        return MOI.Utilities.distance_to_set(x, set) <= tol
-    catch
-        return true  # default to `true`
-    end
+# A default fallback which means that we are unsure.
+is_feasible(x, set, tol) = missing
+
+function is_feasible(
+    x,
+    set::Union{
+        MOI.SecondOrderCone,
+        MOI.RotatedSecondOrderCone,
+        MOI.ExponentialCone,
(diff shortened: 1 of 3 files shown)


src/Constraint.jl
```diff
@@ -40,10 +40,18 @@ end
...
 AbstractTrees.children(c::Constraint) = (c.child,)
 
-# A fallback. Define a new method if `MOI.Utilities.distance_to_set`
-# is not defined.
+# A fallback. Define a new method if `MOI.Utilities.distance_to_set` is not
+# defined.
 function is_feasible(x, set, tol)
-    return MOI.Utilities.distance_to_set(x, set) <= tol
+    try
+      return MOI.Utilities.distance_to_set(x, set) <= tol
```
odow, line 47:
```suggestion
        return MOI.Utilities.distance_to_set(x, set) <= tol
```

src/Constraint.jl
```diff
@@ -40,10 +40,18 @@ end
...
+    catch
```
ericphanson, line 48: I’ve found blanket catches can make debugging hard, but I think we only need to catch MethodErrors?
```suggestion
    catch e
    e Isa MethodError() || rethrow()
```

odow: The other option is just to opt-in to some sets

src/Constraint.jl
```diff
@@ -40,10 +40,18 @@ end
...
+        return true  # default to `true`
```
ericphanson, line 49: Is this correct? Do we skip adding the constraint altogether or do we pass it to the solver who will mark it infeasible?

odow: Ooooo. This actually needs three-way logic

test/test_constraints.jl
```diff
@@ -397,6 +397,32 @@ function test_RelativeEntropyEpiConeSquare()
...
 end
 
+function test_distance_to_set_matrix()
+    x = Variable(2, 2)
+    y = Variable()
+    fix!(x, [1 0; 0 1])
+    # Constraint has a fixed `Matrix` value.
+    model = minimize(y, [sum(x; dims = 1) <= 1, y >= 1])
+    solve!(model, SCS.Optimizer; silent = true)
+    @test ≈(model.optval, 1.0; atol = 1e-6)
```
odow, line 407:
```suggestion
    @test ≈(model.optval, 1.0; atol = 1e-3)
```

src/Constraint.jl
```diff
@@ -40,12 +40,32 @@ end
...
 AbstractTrees.children(c::Constraint) = (c.child,)
 
-# A fallback. Define a new method if `MOI.Utilities.distance_to_set`
-# is not defined.
-function is_feasible(x, set, tol)
+# A default fallback which means that we are unsure.
+is_feasible(x, set, tol) = missing
+
+function is_feasible(
+    x,
```
odow, line 47:
```suggestion
    x::Vector,
```

PR #710, opened 2024-11-17
Fix indexing AbstractExor with BitVector and BitMatrix

odow:
Replacement for https://github.com/jump-dev/Convex.jl/pull/708

cc @RoyiAvital 

We'll leave the `Bool` methods as-is, but this now works as expected like the rest of Convex atoms.

### DIFF
--- a/src/atoms/VcatAtom.jl
+++ b/src/atoms/VcatAtom.jl
@@ -119,7 +119,7 @@ function Base.getindex(x::VcatAtom, inds::AbstractVector{<:Real})
     return IndexAtom(remaining, inds)
 end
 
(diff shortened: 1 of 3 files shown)


src/atoms/VcatAtom.jl
```diff
@@ -119,6 +119,10 @@ function Base.getindex(x::VcatAtom, inds::AbstractVector{<:Real})
     return IndexAtom(remaining, inds)
 end
 
+function getindex(x::VcatAtom, inds::BitVector)
```
odow, line 122:
```suggestion
function Base.getindex(x::VcatAtom, inds::BitVector)
```

PR #717, opened 2024-12-31
Add a Warning to Avoid Julia's Broadcatisng in Expressions

RoyiAvital:
Adding a warning to avoid using `.` in expressions.

Closes https://github.com/jump-dev/Convex.jl/issues/716.

### DIFF
--- a/docs/src/manual/types.md
+++ b/docs/src/manual/types.md
@@ -86,7 +86,9 @@ evaluate(expr)
 ```
 
 !!! warning
-    Julia's broadcasting, using the `.` operator should be avoided. The expression `-log(1.0 .+ A * x)` will fail while `-log(1.0 + A * x)` will be correctly encoded. 
+    Avoid using Julia's broadcasting (the `.` operator). As an example, the
+    expression `-log(1.0 .+ A * x)` will fail while `-log(1.0 + A * x)` will
+    succeed.
 
 ## Constraints
 
--- a/docs/src/manual/types.md
+++ b/docs/src/manual/types.md
@@ -84,6 +84,7 @@ solve!(problem, SCS.Optimizer)
 # Once the problem is solved, we can call evaluate() on expr:
 evaluate(expr)
 ```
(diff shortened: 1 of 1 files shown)


docs/src/manual/types.md
```diff
@@ -84,6 +84,7 @@ solve!(problem, SCS.Optimizer)
 # Once the problem is solved, we can call evaluate() on expr:
 evaluate(expr)
 ```
+!!! warning Julia's broadcasting, using the `.` operator should be avoided. The expression `-log(1.0 .+ A * x)` will fail while `-log(1.0 + A * x)` will be correctly encoded.
```
ericphanson, line 87: I think we need to format it like this:
```suggestion

!!! warning
    Julia's broadcasting, using the `.` operator should be avoided. The expression `-log(1.0 .+ A * x)` will fail while `-log(1.0 + A * x)` will be correctly encoded. 
```
xref https://documenter.juliadocs.org/stable/showcase/#Admonitions

RoyiAvital: I actually was wondering about this. Decided just to follow the current warning in https://github.com/jump-dev/Convex.jl/blob/master/docs/src/manual/advanced.md.
Yet now I see the rendered MarkDown failed me :-).

ericphanson (approved):

docs/src/manual/types.md
```diff
@@ -85,6 +85,9 @@ solve!(problem, SCS.Optimizer)
 evaluate(expr)
 ```
 
+!!! warning
+    Julia's broadcasting, using the `.` operator should be avoided. The expression `-log(1.0 .+ A * x)` will fail while `-log(1.0 + A * x)` will be correctly encoded.
```
odow, line 89:
```suggestion
    Avoid using Julia's broadcasting (the `.` operator). As an example, the
    expression `-log(1.0 .+ A * x)` will fail while `-log(1.0 + A * x)` will
    succeed.
```

Issue #479, opened 2022-02-23
Convex does not support broadcasting


odow:
Convex.jl lacks support for a lot of broadcasting operations. 

This causes a bunch of problems like this:
```Julia
julia> using Convex

julia> x = Variable(2)
Variable
size: (2, 1)
sign: real
vexity: affine
id: 181…059

julia> A = rand(2, 2)
2×2 Matrix{Float64}:
 0.290045  0.201488
 0.890725  0.771457

julia> A * x .+ 1
ERROR: MethodError: no method matching iterate(::Convex.MultiplyAtom)
Closest candidates are:
  iterate(::Union{LinRange, StepRangeLen}) at range.jl:664
  iterate(::Union{LinRange, StepRangeLen}, ::Int64) at range.jl:664
  iterate(::T) where T<:Union{Base.KeySet{var"#s79", var"#s78"} where {var"#s79", var"#s78"<:Dict}, Base.ValueIterator{var"#s77"} where var"#s77"<:Dict} at dict.jl:693
  ...
Stacktrace:
 [1] copyto!(dest::Vector{Any}, src::Convex.MultiplyAtom)
   @ Base ./abstractarray.jl:843
 [2] _collect(cont::UnitRange{Int64}, itr::Convex.MultiplyAtom, #unused#::Base.HasEltype, isz::Base.HasLength)
   @ Base ./array.jl:608
 [3] collect(itr::Convex.MultiplyAtom)
   @ Base ./array.jl:602
 [4] broadcastable(x::Convex.MultiplyAtom)
   @ Base.Broadcast ./broadcast.jl:682
 [5] broadcasted(::Function, ::Convex.MultiplyAtom, ::Int64)
   @ Base.Broadcast ./broadcast.jl:1312
 [6] top-level scope
   @ REPL[4]:1
```

The purpose of this issue is to collate the currently open issues into a single thread, and to make it more discoverable for users coming to report bugs.

Replaces:
* https://github.com/jump-dev/Convex.jl/issues/410
* https://github.com/jump-dev/Convex.jl/issues/355
* https://github.com/jump-dev/Convex.jl/issues/218

RoyiAvital:
In https://github.com/jump-dev/Convex.jl/pull/717 a documentation warning was added.  
It should be removed once this issue is solved.

RoyiAvital:
I had another case with type conversion.
The following will fail:

```julia
[...]

PR #728, opened 2026-02-22
Add broadcasting to Convex.jl

elsoroka:
Claude and I addressed issue #479 in the CVXPY Workshop hackathon by adding support and tests for broadcasting operations.

We also added a CLAUDE.md file for other developers using Claude.


odow:
I'm not necessarily against something like this, but I need to think about it before merging.

They may be a simpler option that works for all operators, like defining a different broadcast style.

